PR #623, opened 2018-06-12
chore: change default pg username to ark

faustbrian:


faustbrian:
No, this is handled via using an `.env` file as you can see based on the `process.env` variables.

vRobM:
I understand that. Why can't those variables represent the current user, which solves a lot of problems with running multiple instances on the same machine (still need to manage ports) as well as bridgechains (deployer v1 which now manages ports).

PR #667, opened 2018-06-16
Error logging updates per function

vRobM:
## Proposed changes

Adding detail to generic error messages

## Types of changes

### DIFF
--- a/packages/core-p2p/lib/monitor.js
+++ b/packages/core-p2p/lib/monitor.js
@@ -64,7 +64,7 @@ module.exports = class Monitor {
         return this.updateNetworkStatus()
       }
     } catch (error) {
-      logger.error(error.stack)
+      logger.error('Network Status: ${error.stack}')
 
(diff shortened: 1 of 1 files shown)


packages/core-p2p/lib/monitor.js
```diff
@@ -281,7 +281,7 @@ module.exports = class Monitor {
 
       return blocks
     } catch (error) {
-      logger.error(error.message)
+      logger.error(Block download: ${error.message}')
```
ItsANameToo: You're missing a single quote

packages/core-p2p/lib/monitor.js
```diff
@@ -64,7 +64,7 @@ module.exports = class Monitor {
         return this.updateNetworkStatus()
       }
     } catch (error) {
-      logger.error(error.stack)
+      logger.error('Network Status: ${error.stack}')
```
faustbrian: This should only log the message, we don't need the whole stack.

PR #674, opened 2018-06-17
Made account api responses consistent with v1

ItsANameToo:
## Proposed changes

There is some inconsistency between the api responses in the old and new api (e.g. strings instead of number or vice versa). I'll take a look at the other endpoints tomorrow.

Additionally, the `accounts` endpoint returns undefined for `vote`, see for example [this one](http://95.179.145.93:4003/api/accounts/delegates?address=DARiJqhogp2Lu6bxufUFQQMuMyZbxjCydN), and as a result it also returns `NaN` for approval (as they both use the `balance` property). I think it should be solved if we use `account.balance` instead of `delegate.balance` (and pass the `account` object to calculate the approval), but currently I'm unable to check this. So if anyone could chime in, I can add that change to this PR too.

## Types of changes

## Checklist


JeremiGendron:
#711 suffers from this issue, will update remaining delegate-calculator files across the repo to reflect these changes (convert to number with +, add 0 check and verify relevant tests)

PR #683, opened 2018-06-20
[WIP] Peer common blocks

### LINKED ISSUE #677: node not always selecting the main chain
The way it selects node to rebuild is too random. which can be really complex to find consensus when there are two advanced chains like on devnet.

Solution, perform a more detailed map of the nodes it can reach.

EDIT:
This issue is created by the nothing at stake (inexpensive process to create a block on both chains)

Implementation details:
-  implement configuration to let user select the chain. This is by identifying  block (id, height) you want to follow
- implement a signalisation method added to the existing (nethash, port, version) to make simple for nodes to identify peers on the same chain. The idea is to hash all selected branches together and send the first 4 bytes of the hash to the signalisation. (nethash, branch, port, version) would become the new signalisation data. ignoring this branch (like for v1) makes the selection of the branch more cumbersome.

alexbarnsley:
## Proposed changes

Connects to highest height peer (@fix please confirm this is done correctly), then determines all peers that have common blocks. This is matching v1 functionality which seems to be missing. Resolves #677 

- Still need to fix tests as it fails to connect to peers
- move common blocks check to cleanPeers
- figure out what to do if there are no peers after cleaned
- update peer list every x rounds to make sure following the right fork
- use common blocks from highest peer?

## Types of changes

## Checklist

## Further comments

@fix my only concern here is using the highest height peer. I think this could be used to sway a network. Could you confirm for me please?

### DIFF
--- a/packages/core-p2p/lib/monitor.js
+++ b/packages/core-p2p/lib/monitor.js
@@ -244,16 +244,8 @@ module.exports = class Monitor {
         logger.error('Could not determine highest peer')
         return this.discoverPeers()
       }
+      const blockIds = await this.__getRecentBlockIds()
       const list = await highestPeer.getPeers()
-
-      const blocks = await container.resolvePlugin('database').query
-        .select('id')
-        .from('blocks')
-        .orderBy({ timestamp: 'DESC' })
-        .limit(10)
-        .all()
-      const blockIds = blocks.map(block => block.id)
-
       for (const peer of list) {
         if (peer.status !== 'OK' || this.peers[peer.ip] || isLocalhost(peer.ip)) {
           continue
@@ -389,5 +381,20 @@ module.exports = class Monitor {
     }
 
     return false
+  }
+
+  /**
+   * Get last 10 block IDs from database.
+   * @return {[]String}
+   */
+  async __getRecentBlockIds () {
+    const blocks = await container.resolvePlugin('database').query
+      .select('id')
+      .from('blocks')
+      .orderBy({ timestamp: 'DESC' })
+      .limit(10)
+      .all()
+
+    return blocks.map(block => block.id)
   }
 }
--- a/packages/core-p2p/lib/monitor.js
+++ b/packages/core-p2p/lib/monitor.js
@@ -9,6 +9,7 @@ const emitter = container.resolvePlugin('event-emitter')
 
 const Peer = require('./peer')
 const isLocalhost = require('./utils/is-localhost')
+const { first, orderBy } = require('lodash')
 const moment = require('moment')
 const delay = require('delay')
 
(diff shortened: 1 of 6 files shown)


fix:
yes i have removed this getCommonBlock issue because i was not convinced, in v2.

now it can be needed in some corner cases like the one we experience in devnet, there is one thing that makes implementation rather subtle:
in v1 this API call is very costy and can take down the network because peers are too much pinging it (when network halts, there is a natural forks, they are all pinging this API).

Somehow we should keep the last say 200 block headers in memory and the api should request on this memory.

side note, we might as well keep last few blocks tx in memory as well from another api call /peer/transactionsFromIds

packages/core-p2p/lib/monitor.js
```diff
@@ -352,4 +398,19 @@ module.exports = class Monitor {
 
     return false
   }
+
+  /**
+   * Get last 10 block IDs from database.
+   * @return {[]String}
+   */
+  async __getRecentBlockIds () {
```
fix: best is too keep last blocks headers in memory instead

packages/core-p2p/lib/monitor.js
```diff
@@ -221,13 +244,25 @@ module.exports = class Monitor {
    */
   async discoverPeers () {
     try {
-      const list = await this.getRandomPeer().getPeers()
+      const highestPeer = await this.getHighestPeer()
```
fix: this is useless, can be tricked to attack the node

packages/core-p2p/lib/monitor.js
```diff
@@ -192,6 +198,23 @@ module.exports = class Monitor {
     return randomPeer
   }
 
+  /**
+   * Get the highest peer.
+   * @return {Peer}
+   */
+  async getHighestPeer () {
```
fix: we cannot make guarantee about the height of the peer, because height was updated the last time there was a connection to it.

a better estimation would be height + (currentTime + lastTimeConnection)/blocktime

fix: and i would shuffle the filtered list before, also removing the very highest one

faustbrian:
@alexbarnsley are you still working on this?

alexbarnsley:
@faustbrian i paused it to look at go-client. i was also having issues with loading blocks into memory of the block manager

alexbarnsley:
@kristjank do you want to take a look at this to see if it's of any use? If not it's probably worth closing as a lot has changed since i was working on this

kristjank:
@alexbarnsley gonna reuse the manager. We can close this issue, but keep the branch?

alexbarnsley:
no worries, i'll close the PR and keep the branch open, use what you need 👌

PR #689, opened 2018-06-21
added pretty-ms to tick-sync-tracker.js

JeremiGendron:
Issue #688 - looking for feedback as my testing environment is not proper atm. The package is included properly and should work, but again my environment is limited  I would like someone to validate it gets printed to the logger properly.

## Proposed changes

As @faustbrian suggested, pretty-ms package does provide an easy to use function for printing human-readable time. I invoke its function and indicate it to follow the format: Xd Yh Zm As and pass the refactored variable tracker.msLeft in the logger string to be displayed while syncing.

## Types of changes

What types of changes does your code introduce?
_Put an `x` in the boxes that apply_

- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Refactoring (improve a current implementation without adding a new feature or fixing a bug)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Build (changes that affect the build system)
- [ ] Docs (documentation only changes)
- [ ] Test (adding missing tests or fixing existing tests)
- [ ] Other... Please describe:

## Checklist

_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code._

- [x] I have read the [CONTRIBUTING](https://docs.ark.io/docs/contributing) documentation
- [x] Lint and unit tests pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)

looking for responses

Issue #688, opened 2018-06-21
Integrate pretty-ms for sync tracker


faustbrian:
Integrating https://github.com/sindresorhus/pretty-ms would give people a pretty printed estimate of how long the download will take at the current speed.

Could look like `(${downloadedBlocks} of ${networkHeight} blocks) ${prettyMs(secondsLeft)}` in https://github.com/ArkEcosystem/core/blob/master/packages/core-blockchain/lib/utils/tick-sync-tracker.js#L30. **Just need to make sure that the time handed in is actually presented in milliseconds for an accurate display!**

JeremiGendron:
@faustbrian all you

PR #691, opened 2018-06-21
pretty-ms revisit (local environment difficulties)

JeremiGendron:
cloned sync-estimate and modified a little bit

Issue #688 - looking for feedback as my testing environment is not proper atm. The package is included properly and should work, but again my environment is limited I would like someone to validate it gets printed to the logger properly.

closes #689 in favor of self

## Further comments

1. height of last downloaded block gets queried every time the sync gets called 
2. tracker.downloadedBlocks gets set only once, when tracker is initiated and reflects the block where the sync begins as a Point of Reference for calculations 
3. percent is really the height * 100 / networkHeight and is independent of how many blocks were downloaded before the sync started at the current iteration
4. blockPerMs takes the elapsed time and divides it by the height minus block start from tracker.downloadedBlocks static constant 
5. timeleft is remaining blocks by blockPerMs

these changes, specifically regarding tracker.downloadedBlocks being constant, help maintain a steady sync timeleft estimate.

### DIFF
--- a/packages/core-blockchain/lib/utils/tick-sync-tracker.js
+++ b/packages/core-blockchain/lib/utils/tick-sync-tracker.js
@@ -6,8 +6,8 @@ const database = container.resolvePlugin('database')
 let tracker = null
 
 module.exports = async block => {
+  const { height } = await database.blocks.getLastHeight()
   if (!tracker) {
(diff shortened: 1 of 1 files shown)


faustbrian:
Just take your time and get your development environment set up properly before you get started so you can test changes.

PR #694, opened 2018-06-21
Find and connect to peers using the API client

j-a-m-l:
This feature will be used on the new desktop wallet, but it could be used in more projects.

Currently, to use the API client, is necessary to discover the peers first, but to do that, it's necessary knowing 1 at least.

These changes provide a way to connect to an almost random peer of the network, without prior knowledge of the peers.

I've added a list of peers (`mainnet` and `devnet`) to not add `core` as dependency. Maybe this could be changed.

Apart from that, if this approach is accepted, we could implement it on other clients.


j-a-m-l:
About the list of peers, maybe we could move all the config from `core/lib/config` to a new package `config`, `configs`, `networks`, or something like that.

packages/client/lib/peers.js
```diff
@@ -0,0 +1,122 @@
+/*
+ * List from https://github.com/ArkEcosystem/peers
+ */
+module.exports = {
+  devnet: [
```
faustbrian: We should avoid hardcoding peers as they will be modified by tharude or adrian here https://github.com/ArkEcosystem/peers and then you have to copy over the lists again from there.

fix: i think we should have both: in case of unavailability of github the whole network could be rekt

packages/client/lib/index.js
```diff
@@ -1,15 +1,71 @@
...
+const sortPeers = require('./utils/sort-peers')
 
 module.exports = class ApiClient {
+  /**
+   * Finds all the available peers, sorted by block heigh and delay
+   *
+   * @param {String} network - Network name
+   * @param {Number} version - API version
+   */
+  static async findPeers (network, version) {
```
faustbrian: This really shouldn't be the job of an API client, the developer that implements the API client should implement a way of finding and filtering peers he would like to use.

All `*-client` packages should only manage the interaction with the API **(send request, return response)** and not implement anything that is implementation specific to other applications.

This looks more like a utility for wallets then something that should be part of any `client` packages.

fix: i disagree, this is something a bit more subtle:
- either you have a full node you ping. This full node is making some work to keep in sync, connecting to random nodes, as random as possible to prevent from attacks. you trust YOUR node
- as a lite client, you are trusting someone else node. To prevents attacks, this is up to the client to be able to play the random access game and find correct peers by analysing the network

j-a-m-l: I understand your idea, but I find more practical having something like this.

I mean, with the typical API client you don't need to know which host is online. You just make requests. When you're just "playing" with the client, it should be straightforward, it shouldn't be very involved.

I started to implement this for the new wallet, but I realized that we've this in different places (the current `desktop-wallet` and the mobile-wallet`, which reuses part of `ark-ts`, which uses parts of `ark-js`) and that, for more use cases, would be useful:

For instance, if a user wants to get some of her transactions to do her taxes, we shouldn't force her to find out how to get which hosts are available for `mainnet` (maybe, she would search that on Google, on GitHub, on the wallet, maybe she would ask on Slack, etc.).

Maybe 1 month later, she try her script/app and it doesn't work because the node is down. So she has to search for a new peer again.

packages/client/lib/peers.js
j-a-m-l: Maybe we could try to use those harcoded peers first, and, in case all of them failed, use GH as a fallback.

j-a-m-l: Other option, maybe better than requesting to GH, is creating a new NPM package (`@arkecosystem/peers`) on the [peers repository](https://github.com/ArkEcosystem/peers). That way, we'd be sure that the peers are updated.

packages/client/lib/index.js
roks0n: I'm with @faustbrian on this one, clients shouldn't be that smart. They should be light and "dumb" and only offer a way to communicate with an API.

Everything else (eg. fetching peers, getting the peer with least latency etc.) should be up to the app to decide what it wants to do and how. If you want to have more complex logic create a separate module for it or if its a single function, put it into utils for example

zillionn: I believe @fix and @j-a-m-l are on your side.

packages/client/lib/peers.js
zillionn: Yes, why don't you guys use the peers repo (https://github.com/ArkEcosystem/peers) instead of hard coding them? This way you'll have to update the peers only on one place.

packages/client/lib/index.js
j-a-m-l: To me, the main problem is that, if we don't add this, @roks0n, we have a half-baked solution that implies that every developer that wants to use this client, should choose between using hardcoded peers or implement something like this. And that is not user friendly at all.

j-a-m-l:
I was waiting to having a resolution about https://github.com/ArkEcosystem/peers/pull/2, but I'm going to continue with this PR (including the `peer.js` data) and adapt it later when we've decided.

j-a-m-l:
Since we don't have a firm decision about how to proceed with https://github.com/ArkEcosystem/peers/pull/2, this PR could be considered as finished now.

Other clients should implement this new feature, but we could wait until having the peer/network configuration proposal properly defined and implemented to do it.

packages/client/__tests__/fixtures/peers.js
```diff
@@ -0,0 +1,47 @@
...
+  ip: '4.4.4.4',
+  port: 4002,
+  version: '1.1.1',
+  errors: 0,
+  os: 'linux',
+  height: 3663605,
+  status: 'OK',
+  delay: 17
+}
+const peers = [peer1, peer2, peer3, peer4]
```
alexbarnsley: Might it be worth just moving this to `exports.peers = [peer1, peer2, peer3, peer4]`? Just saves storing it when it's only used once

packages/client/__tests__/client.test.js
```diff
@@ -37,5 +57,118 @@ describe('API - Client', () => {
...
+      expect(foundPeers).toEqual([
+        peers[1],
+        peers[3],
+        peers[0],
+        peers[2]
+      ])
+    })
+
+    it('should ignore local peers', async () => {
+      ;['127.0.0.1', '::1'].forEach(async ip => {
```
alexbarnsley: semi colon at the start of the line

alexbarnsley: may also be worth adding `::ffff:127.0.0.1` which is in the whitelists

packages/client/lib/index.js
```diff
@@ -1,15 +1,72 @@
...
+    }
+
+    const networkPeers = initialPeers[network]
+    let peers = null
+
+    // Shuffle the peers to avoid connecting always to the first ones
+    shuffle(networkPeers)
+
+    // Connect to each peer to get an updated list of peers until a success response
+    for (let i = 0; i < networkPeers.length; i++) {
```
alexbarnsley: not important here but you can also do `for (const peer of networkPeers) {

packages/client/lib/index.js
```diff
@@ -1,15 +1,72 @@
...
+      const peer = networkPeers[i]
+
+      const client = new ApiClient(peer.ip, version)
+      const response = await client.resource('peers').all()
+      const { data } = response.data
+
+      if (data.success && data.peers) {
+        peers = data.peers
```
alexbarnsley: This may not be the best way to do it. In theory (albeit difficult to do), I could gain access to one of the nodes in `peers.js` and return a different set of peers here. As I say, it'd be very difficult to do but not impossible. It can always be revised later though if needed, as it wouldn't always cause issues anyway due to the shuffling

packages/client/lib/index.js
```diff
@@ -1,15 +1,72 @@
...
+        break
+      }
+    }
+
+    // Return at least the initial (hardcoded) peers
+    if (!peers) {
+      return networkPeers
+    }
+
+    const selfIps = ['127.0.0.1', '::1']
```
alexbarnsley: as mentioned earlier, maybe add `::ffff:127.0.0.1` just in case

packages/client/lib/peers.js
alexbarnsley: Any more thoughts on this Juan?

packages/client/lib/utils/sort-peers.js
```diff
@@ -0,0 +1,10 @@
+const orderBy = require('lodash.orderby')
+
+/**
+ * Sorts the peers, in place, by block height and delay
+ */
+const sortPeers = peers => {
```
alexbarnsley: As the const is only used once, could we directly assign it to `module.exports`?

packages/client/__tests__/client.test.js
j-a-m-l: I'm not sure about it. It doesn't look like a valid IP address

packages/client/lib/index.js
j-a-m-l: Yes, I expect that in the next weeks we start to sign the seed peers.

PR #696, opened 2018-06-21
Add instructions to remove unnecessary texts on the PR template

j-a-m-l:
Some contributors keep the entire text of the PR template instead of removing the unnecessary parts. These changes try to reduce the size of those paragraphs by asking to remove the parts that aren't used.

This could be changed in other respositories too (and https://github.com/ArkEcosystem/boilerplate/blob/master/.github/PULL_REQUEST_TEMPLATE.md).

What other instructions could we modify to avoid it?

- [ ] Decide about the text
- [ ] Copy it to boilerplate
- [ ] Copy it to other projects


faustbrian:
Could this be send to https://github.com/ArkEcosystem/boilerplate so we can copy it over from there to other repositories?

j-a-m-l:
Thanks @luciorubeens for the tip. I've updated it.

What do you think, guys? When it's ready, I'll replace those from other projects too.

kristjank:
@j-a-m-l I think its ok. We can improve and adjust later on

PR #713, opened 2018-06-26
imp: change delegate "votes" property to "vote" as in v1

ItsANameToo:
## Proposed changes

Changes the "votes" property that is currently returned by the delegates API to "vote" (and also from Number to String) to be in line with the v1 api

## Types of changes

## Checklist

### DIFF
--- a/packages/core-api/lib/versions/1/transformers/delegate.js
+++ b/packages/core-api/lib/versions/1/transformers/delegate.js
@@ -12,7 +12,7 @@ module.exports = (delegate) => {
     username: delegate.username,
     address: delegate.address,
     publicKey: delegate.publicKey,
-    votes: delegate.votebalance,
+    vote: delegate.votebalance + '',
     producedblocks: delegate.producedBlocks,
     missedblocks: delegate.missedBlocks,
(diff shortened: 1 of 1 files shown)


packages/core-api/lib/versions/1/transformers/delegate.js
```diff
@@ -12,7 +12,7 @@ module.exports = (delegate) => {
     username: delegate.username,
     address: delegate.address,
     publicKey: delegate.publicKey,
-    votes: delegate.votebalance,
+    vote: delegate.votebalance + '',
```
zillionn: Something to keep in mind when converting to string. If for some reason the `votebalance` is `1E-8`, then `votebalance + ''` will still return `number`.

ItsANameToo: Can you elaborate on this? `typeof(0.00000001 + '')` returns `String` for me, same goes for `typeof(1E-8 + '')` or am I misunderstanding you?

zillionn: Nevermind, ignore my comment.

PR #714, opened 2018-06-26
Protect against SQL injections

### LINKED ISSUE #560: The Sequelize query builder could accept SQL injections
Since it's using raw queries, but they aren't escaped properly: https://github.com/fix/ark-core/blob/575ae88505a1a92a648825f9069c2ff381e79280/packages/core-database-sequelize/lib/query-builder.js

It could be fixed using replacements, probably: http://docs.sequelizejs.com/manual/tutorial/raw-queries.html

supaiku0:
## Proposed changes

Proposed solution to fix #560.

The PR basically consists of the following:
1. Turn all column names into snake_case, via migration scripts (Is this the right way?)
and add field mappings to the models
2. Adjust raw sql to use new column names
3. Finally make use Sequelize replacements
4. Fix unit tests / regressions

The column renaming is done to follow the PostgreSQL standard naming conventions and to ultimately get rid of quoted column names in raw queries which don't work nicely together with Sequelize's escaping mechanism.

## Types of changes

## Further comments
There's most likely a lot of random breakage, since I didn't test it yet beyond `[ERROR]   : Block download: Peer 167.114.29.51 is unresponsive`, but up to that point it's working. :stuck_out_tongue: 

-  I'm not sure if it makes sense to use migration scripts since it's a big change, why not just start over completely fresh?
- Unit tests need to be fixed
- Code inconsistently uses blockId / blockid, don't know the implications yet, should it be refactored to blockId?
- I guess the SQLite webhooks db needs to be adapted as well?
- Instead of [replacements, bind parameters](http://docs.sequelizejs.com/manual/tutorial/raw-queries.html) could also be used
- Please leave your feedback

### DIFF
--- a/packages/core-database-sequelize/lib/query-builder/sql-builder.js
+++ b/packages/core-database-sequelize/lib/query-builder/sql-builder.js
@@ -67,13 +67,6 @@ class SqlBuilder {
       }
 
       if (['IN', 'NOT IN'].includes(item.operator)) {
-        if (Array.isArray(item.value)) {
-          item.value = item.value.map(value => {
-            this.__replacements.push(value)
-            return '?'
-          }).join(',')
-        }
-
         this.__replacements.push(item.value)
         return `${item.column} ${item.operator} (?)`
       }
--- a/packages/core-database-sequelize/lib/query-builder/sql-builder.js
+++ b/packages/core-database-sequelize/lib/query-builder/sql-builder.js
@@ -35,12 +35,7 @@ class SqlBuilder {
    * @return {String}
    */
   __buildSelect (clauses) {
-    const columns = clauses.select.columns
-      .map(column => column)
-
-    const aggregates = clauses.select.aggregates
-      .map(column => column)
-
+    const { columns, aggregates } = clauses.select
     return `SELECT ${columns.concat(aggregates).join(',')} `
   }
 
--- a/packages/core-database-sequelize/lib/query-builder/sql-builder.js
+++ b/packages/core-database-sequelize/lib/query-builder/sql-builder.js
@@ -1,5 +1,4 @@
 const { camelCase, upperFirst } = require('lodash')
-const escape = require('./utils/escape')
 
 class SqlBuilder {
   /**
@@ -13,7 +12,9 @@ class SqlBuilder {
       'groupBy', 'orderBy', 'limit', 'offset'
     ]
 
-    return clauseOrder
+    this.__replacements = []
+
+    const sql = clauseOrder
       .map(clause => {
         if (!clauses[clause]) {
           return false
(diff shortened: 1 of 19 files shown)


faustbrian:
@supaiku0 There seems to be a conflict. I will try to find time to test it this weekend. @alexbarnsley could you also take a look if you find a minute?

supaiku0:
@faustbrian Thanks, I rebased onto the master branch. Should work again.

supaiku0:
I'm working on the unit tests of `core-database-sequelize` which exposed some quirks. Please wait until I updated the branch before you give it a try.

supaiku0:
The majority of the unit tests pass again (only 2 fails instead of 3 as on master). I hope that means there aren't any major regressions left.

Just hit me on Slack or gitter if realtime collaboration is desired.

packages/core-database-sequelize/lib/query-builder/sql-builder.js
```diff
@@ -67,13 +67,6 @@ class SqlBuilder {
       }
 
       if (['IN', 'NOT IN'].includes(item.operator)) {
-        if (Array.isArray(item.value)) {
```
faustbrian: Does Sequelize automatically join the `item.value` array with `,`?

faustbrian: Alright, will have time to test it tomorrow.

supaiku0:
@fix @kristjank I'll rebase the changes onto master later today, can someone of you do a review then?

supaiku0:
@fix @kristjank Ping :rocket:

kristjank:
@supaiku0 very nice PR 🎢 🎉  
- did you sync from 0, was all 🚦 green :) 
- will spin up a node on this PR also... 
more feedback incoming...

packages/core-database-sequelize/lib/query-builder/sql-builder.js
```diff
@@ -33,7 +36,7 @@ class SqlBuilder {
    */
   __buildSelect (clauses) {
     const columns = clauses.select.columns
-      .map(column => escape(column))
+      .map(column => column)
```
j-a-m-l: This line doesn't do anything

supaiku0:
@j-a-m-l No, I don't mind. Will do it tomorrow.

packages/core-database-sequelize/lib/query-builder/sql-builder.js
supaiku0: The line after this one neither. :yum:

supaiku0:
@j-a-m-l I added some basic tests. Feel free to provide more examples. 

I wonder why the drop table one doesn't "work" on master, the generated SQL statement definitely drops the table when I execute it manually on a test SQLite db. Any idea? :thinking:

alexbarnsley:
@supaiku0, does it fail when running on postgres? We noticed that sqlite and postgres sometimes behaved differently

supaiku0:
Rebased.

No point in putting any more effort into the unit tests since they don't even run against Postgres. But I talked with Brian, if I find the time I will look into dropping SQLite for the tests. Afterwards we should revise the unit tests.

PR #732, opened 2018-07-05
Add a docker-compose configuration for development

j-a-m-l:
The aim of this PR is provide an easy way to run `core` without installing and configuring PostgreSQL and Redis.

- [X] `docker-compose.yml` file
- [ ] Update docs: https://docs.ark.io/v1.0/docs/docker (after merging the PR)

### DIFF
--- a/docker/development/docker-compose.yml
+++ b/docker/development/docker-compose.yml
@@ -1,0 +1,29 @@
+#
+# For running some services on development without tainting your system
+#
+version: '2'
+services:
+
+  redis:
+    image: "redis:alpine"
+    container_name: ark-development-redis
+    ports:
+      - '127.0.0.1:6379:6379'
+    volumes:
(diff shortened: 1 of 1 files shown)


j-a-m-l:
It could be done with `docker-compose -d redis postgres` too, but this way, we isolate these services (and future ones) from the defaults used on `devnet` and `mainnet`.

faustbrian:
What exactly is the point of this as we already have docker files?

j-a-m-l:
We've them, but they include `ark-core` too. It's easy to accidentally download, install and run the `ark-core` container. It'd be simpler having a `docker-compose` configuration without it. 

This file could support later other alternative services or dbs, for developing or testing purposes, without changing the other configurations, which provide the standard way.

Apart from that, it seems that they're going to be replaced by @adrian69 with production-ready Docker images.

PR #742, opened 2018-07-09
Fix issue with unconfirmedBalance being undefined

ItsANameToo:
## Proposed changes

When querying the `accounts/getAllAccounts` endpoint, the unconfirmedBalance would be `undefined` (and some other account endpoints that made use of the same transformer had this too) as it was not available in the `wallet` model. Simple fix: use `balance` also for `unconfirmedBalance`, since they are already interchangeably used for other endpoints such as `accounts/get`:

```
return utils.respondWith({
      balance: account ? `${account.balance}` : '0',
      unconfirmedBalance: account ? `${account.balance}` : '0'
    })
```

## Types of changes

## Checklist

### DIFF
--- a/packages/core-api/lib/versions/1/transformers/account.js
+++ b/packages/core-api/lib/versions/1/transformers/account.js
@@ -15,7 +15,7 @@ module.exports = (model) => {
     votes: model.votes,
     username: model.username,
     balance: `${model.balance}`,
-    unconfirmedBalance: `${model.unconfirmedBalance}`,
+    unconfirmedBalance: `${model.balance}`,
     multisignatures: [],
     u_multisignatures: [],
     unconfirmedSignature: hasSecondSignature ? 1 : 0,


kristjank:
@ItsANameToo try to get account also from transactionPool and display balance from there. If there are no pending transaction it should be the same, if there are pending transaction it should be smaller for that pending amount...

you can try to get pool account via this call:
`blockchain.transactionPool.walletManger.getWalletByAddress(ADDRESS).balance .. this returns wallet object... 

let me know how it goes

ItsANameToo:
@kristjank thanks, I'll take a look!

ItsANameToo:
Just a heads up: I'll look into this again on friday. Sorry for the delay!

faustbrian:
This actually needs to stay the way it is for the v1 API as ark-node always provides the same value for for `balance` and `unconfirmedBalance`. If this is changed it is a breaking change without backwards compatibility.

I also remember it being mentioned a long time ago that `unconfirmedBalance` is a useless relic from LISK so there is no point in supporting it in the future.

PR #746, opened 2018-07-10
Fix remaining core-database-sequelize tests

supaiku0:
~~The test `should find all transactions that holds all the conditions (AND)`  didn't pass, because the genesis transaction is of type 0.~~

The test `should update the delegate` didn't pass, because calling `updateDelegateStats`  when `blocksInCurrentRound` is not set has no effect. Calling `applyRound` ensures that `blocksInCurrentRound` is set and the last failing test passes.

### DIFF
--- a/packages/core-database-sequelize/__tests__/connection.test.js
+++ b/packages/core-database-sequelize/__tests__/connection.test.js
@@ -159,7 +159,9 @@ describe('Sequelize Connection', () => {
       const wallet = connection.walletManager.getWalletByPublicKey('03e59140fde881ac437ec3dc3e372bf25f7c19f0b471a5b35cc30f783e8a7b811b')
       expect(wallet.missedBlocks).toBe(0)
 
-      await connection.updateDelegateStats(genesisBlock, delegates)
+      const { height } = genesisBlock.data
+      await connection.applyRound(height)
+      await connection.updateDelegateStats(height, delegates)
 
       expect(wallet.missedBlocks).toBe(1)
     })


supaiku0:
Btw. before you start wondering I forgot to mention that I ran the tests on my sql branch. I don't think all the tests are fixed on master. :thinking:

supaiku0:
The tx you provided has a completely different recipient in my genesisBlock:

```javascript
{
    'type': 3,
    'amount': 0,
    'fee': 0,
    'recipientId': 'AcmXmomxpP8NahbbFivq32QmLuKFkTkqRg',
    'senderPublicKey': '0311077c86a98b67850e7ed2c81775d094cf81c6991082ddc33fc7be5347dc765d',
    'timestamp': 0,
    'asset': {
      'votes': [
        '+0311077c86a98b67850e7ed2c81775d094cf81c6991082ddc33fc7be5347dc765d'
      ]
    },
    'signature': '3045022100b1615d16763c46d42ca2aae967f04c1c07c119b5af7a378c262ba85515a8d35002202cf7df91676cd137943720e93f06c11907412a6bdc5ef2157cf536a203cf83a3',
    'id': '76fb5a1de90f245b1eeb79cb11c7bea7c8b738add0fb8cd95191186a944b0229',
    'senderId': 'AcmXmomxpP8NahbbFivq32QmLuKFkTkqRg'
  }
```

But indeed there is a tx which should have been returned (maybe you meant this one?):

```javascript
{
    'type': 3,
    'amount': 0,
    'fee': 0,
    'recipientId': 'AHXtmB84sTZ9Zd35h9Y1vfFvPE2Xzqj8ri',
    'senderPublicKey': '02ad799c6bd670746892bd4331e1aebada26a2cc3ccaf0fde1e94942b20066b05a',
    'timestamp': 0,
    'asset': {
      'votes': [
        '+02ad799c6bd670746892bd4331e1aebada26a2cc3ccaf0fde1e94942b20066b05a'
      ]
    },
    'signature': '3045022100e3c7b5d6a72acde4d22e8c1c6cd864c549deba89683f4b84320407d6c380827c02202da57df0ab7cd381b776bdf85802aed371e7cea7269a84f911b1d8e9956badee',
    'id': '8da75c8100e6248ab37cc92f72ed9facec3067f4f82f03db8bb8063791463fb3',
    'senderId': 'AHXtmB84sTZ9Zd35h9Y1vfFvPE2Xzqj8ri'
  }
```

I'll look at it again, thanks!

j-a-m-l:
Sorry, I copied the wrong ID.

Yes, the that other transaction should be returned by the search.

supaiku0:
@j-a-m-l 
I investigated a bit and found the problem. If you inspect the genesis block you'll notice that the recipients of all VOTE / Type 3 transactions (the transactions in the block around 100+) are from the devnet.

To be precise, this line is the problem:
https://github.com/ArkEcosystem/core/blob/master/packages/crypto/lib/models/transaction.js#L351

It deserializes the serialized mainnet address into a devnet address meaning
instead of `AHXtmB84sTZ9Zd35h9Y1vfFvPE2Xzqj8ri` we get `D6u7ewD5qioHHf1gs5sFKYARnkq93tPbDZ` .

j-a-m-l:
@supaiku0 Good job!

supaiku0:
@j-a-m-l Reverted the changes to the test. I guess this one goes a bit out of scope.

j-a-m-l:
Ok, @supaiku0, I'll review it myself

PR #770, opened 2018-07-17
fix deregister test  in plugin.test

### LINKED ISSUE #769: Fix 'deregister plugins' test
**Describe the bug**
The de-register plugins tests fails because it's written erroneously. Awilix doesn't appear to have a concept of "removing" an instance once you register it(w/o tearing down the entire thing). However, since we're interested in making sure the deregister function of a plugin is called(or not called if not set), we can sufficiently test this via jest mock fns.

**To Reproduce**
modify plugin.test.js to run "tearDown"(it's currently ignored)

**Expected behavior**
Test should pass

paroxysm:
## Proposed changes
Fixing a test by slightly changing it's behavior

## Types of changes

## Checklist

## Fixes #769


packages/core-container/__tests__/registrars/plugin.test.js
```diff
@@ -73,7 +73,7 @@ describe('Plugin Registrar', () => {
   })
 
   describe('tearDown', () => {
-    xit('should deregister all the plugins in inverse order', async () => {
+    it('should deregister plugins supporting deregister', async () => {
```
j-a-m-l: Everything looks OK, but I'll split this test into 2 different tests:
 1. "in inverse order": it's important that this is done in a predictable way
 2 "plugins supporting deregister"

packages/core-container/__tests__/__stubs__/plugin-a.js
```diff
@@ -9,7 +9,9 @@ exports.plugin = {
   register: (container, options) => {
     return {
       container,
-      options
+      options,
+      deregister: deregisterMockFn
```
j-a-m-l: Why is necessary adding the `deregisterMockFn` in 2 different places?

packages/core-container/__tests__/registrars/plugin.test.js
paroxysm: it's a little bit more difficult the inverse order. These "deregister" callbacks for each plugin are not linekd to each other, so from the outside perspective, they're all unregistered at once. You'd have to have the plugins know of a shared collection(an array or so) such that when 'deregister' is called, they push their plugin name into the collection.

Then you'd have to have a way to propagate that collection to the test so that it can create assertions on that.

It seems like a convoluted method just to test this, seemingly, atomic operation.

packages/core-container/__tests__/__stubs__/plugin-a.js
paroxysm: Because of how the plugins are registered in the code. The obj that ends up in the Awilix container, is not the same one that the plugins export. The 'deregister' property needs to be set on the exported object. However, you cannot access this object from the Awilix container.

This is the approach I chose such that I can still acquire a reference to the mock function(from Awilix), while still ensuring that it's called by the PluginRegistrar plumbing

kristjank:
@j-a-m-l  @paroxysm what is the status here? just checking... tnx

packages/core-container/__tests__/registrars/plugin.test.js
j-a-m-l: It's simpler than that: all the registered plugins that are "deregisterable" belong to the `this.deregister` array.
Each time a new plugin is registered, they're pushed to that Array, so they keep the order.

The current test doesn't check that, but it could be renamed as the 2 option, as you had done, while adding a new one to deal with the first case.

Apart from that, a supplementary test would be testing that registering a new plugin, adds it to the `deregister` array.

packages/core-container/__tests__/__stubs__/plugin-a.js
j-a-m-l: The `deregister` method should be on the root level only. Other packages, such as `core-database-sequelize` and `core-p2p`, are using it that way.

Maybe the problem is other thing, such as https://github.com/ArkEcosystem/core/blob/2c02ef64ac5d2f90fc47858ba4e5a71144167be1/packages/core-container/lib/registrars/plugin.js#L63.

paroxysm: So @j-a-m-l , given the way the plugins are registered right now(file based approach) this is one approach. 
A second approach is to have the test manually 'require()' all the plugins that are supposed to have deregister to get a reference to the mockFn; that will remove the need for 2 references. I'll make those changes

paroxysm:
I've put back the test for reverse order. I've also re-written the test such that we don't need to set 'deregister' in two places @j-a-m-l

paroxysm:
Cool, I'll take a look.

boldninja:
Hi @paroxysm can you please provide me your ARK address or contact me on ARK's Slack (username : boldninja)?

paroxysm:
@boldninja  I'll send u a PM on Slack when I get home!

paroxysm:
@boldninja  Just PM'd u on Slack

paroxysm:
u678&^

PR #771, opened 2018-07-19
Bug fix : findAllLegacy senderId parameter was not converted correctly to sender_public_key database field

air1one:
## Proposed changes

Bug fix : findAllLegacy senderId parameter was not converted correctly to sender_public_key database field, causing "crash" (API call resulted in "internal server error"). 

## Types of changes

- [x] Bugfix (non-breaking change which fixes an issue)

## Checklist

- [x] I have read the [CONTRIBUTING](https://docs.ark.io/docs/contributing) documentation
- [x] Lint and unit tests pass locally with my changes
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)

PR #776, opened 2018-07-23
 Bug fix : findAll / findAllLegacy search transactions with senderId invalid or unknown should return no transaction

air1one:
## Proposed changes
Bug fix : findAll / findAllLegacy search transactions with senderId invalid or unknown should return no transaction.

findAll / findAllLegacy were ignoring search parameter senderId when it was invalid or unknown (no corresponding public key found). They would just make the query without any sender_public_key where clause, then returning unwanted results.

Related to #771 : please first merge (if OK) #771 before going through this one. @j-a-m-l might review as he did for #771.

## Types of changes

## Checklist


packages/core-database-sequelize/lib/repositories/transactions.js
```diff
@@ -28,9 +28,7 @@ module.exports = class TransactionsRepository extends Repository {
     if (params.senderId) {
       const senderPublicKey = this.__publicKeyfromSenderId(params.senderId)
 
-      if (senderPublicKey) {
-        params.senderPublicKey = senderPublicKey
-      }
+      params.senderPublicKey = senderPublicKey
```
j-a-m-l: In the legacy method, the conditions are disjunctions (OR) instead of all necessary (AND), so we could return an empty result here, without performing any request.

packages/core-database-sequelize/lib/repositories/transactions.js
```diff
@@ -75,9 +73,7 @@ module.exports = class TransactionsRepository extends Repository {
...
```
j-a-m-l: Since the condition has been removed, then you could assign the value directly to `params.senderPublicKey`

packages/core-database-sequelize/__tests__/repositories/transactions.test.js
```diff
@@ -121,6 +121,15 @@ describe('Transaction Repository', () => {
...
     })
 
+    it('should find no transaction when passed parameter senderId is invalid or unknown', async () => {
+      await connection.saveBlock(genesisBlock)
+      const invalidSenderId = 'thisIsNotAValidSenderId'
+
+      const transactionsSenderId = await repository.findAll({ senderId : invalidSenderId })
+
+      expect(transactionsSenderId.count).toBe(0);
+    })
```
j-a-m-l: Check the number of rows too

air1one:
OK @j-a-m-l I applied the changes you suggested, hope it's all good now 👌

PR #777, opened 2018-07-23
Fix core container deregister test

j-a-m-l:
This PR continues the work started on https://github.com/ArkEcosystem/core/pull/770


packages/core-container/lib/registrars/plugin.js
```diff
@@ -57,10 +57,12 @@ module.exports = class PluginRegistrars {
    * @return {void}
    */
   async tearDown () {
-    const plugins = this.deregister.reverse()
+    const modules = this.deregister.reverse()
 
-    for (let i = 0; i < plugins.length; i++) {
-      await plugins[i].plugin.deregister(this.container, plugins[i].options)
+    for (let i = 0; i < modules.length; i++) {
+      if (modules[i].plugin.deregister) {
```
paroxysm: This 'if' block here is un-necessary? Modules only end up in the 'this.deregister' array if they have a 'deregister' property set

paroxysm:
Ah I see. I wanted to avoid changing the structure that's sent to the PluginRegistrar to avoid side-effects but for testing purposes this makes sense.

packages/core-container/lib/registrars/plugin.js
j-a-m-l: Good catch!

PR #780, opened 2018-07-24
Fix : acceptNewPeer condition to accept peers others than 'myself'

air1one:
## Proposed changes

acceptNewPeer was accepting only its own IP, this fixes it by accepting peers others than 'myself'.

## Types of changes

## Checklist

- [x] I have added tests that prove my fix is effective or that my feature works : **tested on community-run v2 devnet**

## Further comments

This was tested on community-run v2 devnet. I'm not sure how to add this kind of test to be run on testnet (need 2 peers + ark_env test is changing behaviour).


kristjank:
Good catch  @air1one 

and @supaiku0  :) always good to remove. @air1one please update the PR and remove the section a few lines below, that does the same as isMySelf and test again.

thanks

air1one:
Thanks @supaiku0 yes it always feels good to remove some code 😄 
Done and tested again, looks good !

kristjank:
all green, let's meeeerge 👍

PR #783, opened 2018-07-24
Fix some problems of packages with tests that run indefinitevely

j-a-m-l:
## Proposed changes

The packages that mount a blockchain during tests keep running without executing all their tests or even after having finished.

This PR fixed that, although it doesn't enable all those packages, because some of them have failing tests.

## Further comments

The changes are:
  - Close the Redis connection (of the database cache) when Sequelize is disconnected
  - Add a new state (`stopped`) to the blockchain state machine to be able to stop the blockchain when tests are finished
  - Change Travis configuration and test script to increase the speed and detect problems

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,8 @@ before_script:
   - sudo redis-server /etc/redis/redis.conf --port 6379
 
 script:
-  - yarn test:coverage --ci
+  # See https://jestjs.io/docs/en/troubleshooting.html#tests-are-extremely-slow-on-docker-and-or-continuous-integration-ci-server
+  - yarn test:coverage --ci --maxWorkers 4
   - codecov
 
 jobs:


faustbrian:
@j-a-m-l any change you finish this until monday/tuesday?

j-a-m-l:
Yes, this PR would be ready, although I don't know if it would address and solve all the problems that are affecting the tests of these packages.

j-a-m-l:
I've realized that the problems of the tests are harder than I though. It may imply rewriting a lot of code of several packages and mocking more parts. For instance:

 - At least 1 package tries to connect to external sources to accomplish its task (core-p2p) when it's not necessary. Since it's not able to get the fresh peers, it tries indefinitely.

I'll continue my work on another PR to avoid making this one bigger and bigger.

PR #784, opened 2018-07-24
Fix: Delay creation of Genesis Block in unit tests to after setup

supaiku0:
My original plan was to simply fix the tests  `should search transactions that includes all of them (AND)` and `should find all transactions that holds all the conditions (AND)` of  `core-database-sequelize`, but it turned out to be a fundamental issue.

### Why were the tests failing?
They both use the genesis transaction's recipientId to find a type-3 transaction. When the genesis block is created and serialized, the transactions are serialized with a **devnet** network version. Now when a transaction is deserialized a wrong recipientId is calculated and since all addresses in the genesis block start with `A`, it fails to find anything.

### Why devnet?
It is the default config. The creation of the genesis block happens **before** any test has the chance to run `app.setUp()`. This means all tests right now (except those using a devnet config) are using a genesis block with broken recipientIds, which kinda works except in those two cases.

### The fix
Delay the creation of the genesis block until `app.setUp()`has been called which loads the expected network config.

Due to the changes some tests suddenly failed, but they only required minor changes  to work with the now correct genesis block.  I also checked all other unit tests using a genesis block for regressions.

Now that all tests of `core-database-sequelize` pass, I re-enabled them in the jest config.

### DIFF
--- a/packages/core-database-sequelize/__tests__/repositories/transactions.test.js
+++ b/packages/core-database-sequelize/__tests__/repositories/transactions.test.js
@@ -9,9 +9,9 @@ const SPV = require('../../lib/spv')
 
 const app = require('../__support__/setup')
 const createConnection = require('../__support__/utils/create-connection')
-const genesisBlock = require('../__fixtures__/genesisBlock')
-const genesisTransaction = genesisBlock.transactions[0]
 
+let genesisBlock
+let genesisTransaction
 let connection
 let repository
 let spv
@@ -20,8 +20,12 @@ const getWallet = address => {
   return spv.walletManager.getWalletByAddress(address)
 }
 
-beforeAll(async () => {
+beforeAll(async (done) => {
   await app.setUp()
+  genesisBlock = require('../__fixtures__/genesisBlock')
+  genesisTransaction = genesisBlock.transactions[0]
+
+  done()
 })
 
 afterAll(async () => {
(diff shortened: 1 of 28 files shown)


packages/core-database-sequelize/__tests__/repositories/transactions.test.js
```diff
@@ -20,8 +20,12 @@ const getWallet = address => {
   return spv.walletManager.getWalletByAddress(address)
 }
 
-beforeAll(async () => {
+beforeAll(async (done) => {
```
j-a-m-l: This callback could be removed

packages/core-database/__tests__/repositories/delegates.test.js
```diff
@@ -1,17 +1,17 @@
...
 
+let genesisBlock
 let repository
 let walletManager
 let calculateApproval
 let calculateProductivity
 
 beforeAll(async (done) => {
   await app.setUp()
+  genesisBlock = require('../__fixtures__/genesisBlock')
```
j-a-m-l: Could you add a brief comment here and in the other files? It's the best way to avoid that other devs that aren't aware of these subtle problems, make this same error

packages/core-database-sequelize/__tests__/repositories/transactions.test.js
supaiku0: I copied it from another test: https://github.com/ArkEcosystem/core/blob/master/packages/core-database/__tests__/interface.test.js#L8

But I can remove it.

packages/core-database/__tests__/repositories/delegates.test.js
supaiku0: Done.

supaiku0:
@j-a-m-l Ping.

Rebased, added comments and changed a few more tests which I missed before. Many of them were already failing before so I'm not entirely sure if I broke something, but the amount of  passed/failed are the same.

j-a-m-l:
@supaiku0 Tests are failing after updating the branch

supaiku0:
@j-a-m-l Green again.

j-a-m-l:
Thanks

PR #786, opened 2018-07-27
fix: errors in transaction guard

supaiku0:
I found two little bugs in the `TransactionGuard` while looking over https://github.com/ArkEcosystem/core/pull/785 but they are not directly related to that PR since they are already on master.

The first seems to be a copy paste mistake and the second a typo.


packages/core-transaction-pool/lib/guard.js
```diff
@@ -140,7 +140,7 @@ module.exports = class TransactionGuard {
         return true
       }
 
-      this.invalid.push(this.transactions)
+      this.invalid.push(this.transaction)
```
kristjank: `this.invalid.push(transaction)`

PR #792, opened 2018-08-01
Fix handling of "boolean" environment variables

vasild:
`expression || true` always evaluates to `true`.
`expression || false` always evaluates to `expression`.

So, while it makes perfect sense to have something like
`process.env.ARK_JSON_RPC_PORT || 8080`, constructs like
`process.env.ARK_TRANSACTION_POOL_ENABLED || true` mean that it is not
possible to disable (the transaction pool).

`process.env.ARK_DB_LOGGING || false` is just clutter and is the same as
`process.env.ARK_DB_LOGGING`.

After this change it is possible to disable
ARK_API_ENABLED
ARK_GRAPHQL_ENABLED
ARK_TRANSACTION_POOL_ENABLED
ARK_WEBHOOKS_ENABLED
by not setting the relevant variable in the environment (or setting it
to an empty string).

This is a behavioral change - users that did not had e.g.
ARK_TRANSACTION_POOL_ENABLED set in the environment before this change
would have it enabled, whereas after this change it will be disabled for
them and in order to enable it they must define
ARK_TRANSACTION_POOL_ENABLED in the environment (to some value).

## Types of changes

## Checklist

- [ ] I have read the [CONTRIBUTING](https://docs.ark.io/docs/contributing) documentation

### DIFF
--- a/packages/core-database-sequelize/README.md
+++ b/packages/core-database-sequelize/README.md
@@ -16,7 +16,7 @@ yarn add @arkecosystem/core-database-sequelize
 module.exports = {
   dialect: 'sqlite',
   storage: `${process.env.ARK_PATH_DATA}/database/${process.env.ARK_NETWORK_NAME}.sqlite`,
-  logging: process.env.ARK_DB_LOGGING || false
+  logging: process.env.ARK_DB_LOGGING
 }
 ```
 
--- a/packages/core-database-sequelize/lib/defaults.js
+++ b/packages/core-database-sequelize/lib/defaults.js
@@ -3,7 +3,7 @@
 module.exports = {
   dialect: 'sqlite',
   storage: process.env.ARK_DB_STORAGE || `${process.env.ARK_PATH_DATA}/database/${process.env.ARK_NETWORK_NAME}.sqlite`,
-  logging: process.env.ARK_DB_LOGGING || false,
+  logging: process.env.ARK_DB_LOGGING,
   redis: {
     host: process.env.ARK_REDIS_HOST || 'localhost',
     port: process.env.ARK_REDIS_PORT || 6379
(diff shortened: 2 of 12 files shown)


vasild:
If this looks too risky, maybe add another change on top of it that renames FOO_ENABLED to FOO_DISABLED, so that things that were (permanently) enabled before keep being enabled (assuming the user does not have FOO_ENABLED or FOO_DISABLED in the environment)? And then in order to disable, the user would have to explicitly define FOO_DISABLED to some non-empty value.

vasild:
Aha! Travis choked on this:
```
[2018-08-01 19:29:17][WARN]    : Redis transaction pool disabled - please enable if run in production
[2018-08-01 19:29:18][ERROR]   : Could not get transactions for forging from Redis:  TypeError: Cannot read property 'join' of undefined
```

faustbrian:
```
This is a behavioral change - users that did not had e.g.
ARK_TRANSACTION_POOL_ENABLED set in the environment before this change
would have it enabled, whereas after this change it will be disabled for
them and in order to enable it they must define
ARK_TRANSACTION_POOL_ENABLED in the environment (to some value).
```

Things like a transaction pool should always be enabled by default. We could modify the commander but not everyone uses the commander, especially not for a local environment.

The defaults that were in place before this change should be retained.

vasild:
@faustbrian so, go with https://github.com/ArkEcosystem/core/pull/792#issuecomment-409694044?

faustbrian:
Seems like an OK solution for now.

faustbrian:
Travis is still failing at the moment because the transaction pool is not enabled by default.

vasild:
I just extended this change to also rename ..._ENABLED to ..._DISABLED so that things remain enabled by default (assuming people do not have ..._DISABLED defined in their environment).

PR #796, opened 2018-08-02
feat: p2p request throttling

faustbrian:
## Proposed changes

Basic P2P throttling based on IP and if the address is blacklisted or not.

## Types of changes

## Checklist


packages/core-p2p/lib/server/plugins/throttle/bucket.js
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  /**
+   * Check if a rate limiter is associated with the given IP.
+   *
+   * @param  {String}  ip
+   * @return {Boolean}
+   */
+  has (ip) {
+    return this.get(ip)
```
alexbarnsley: This doesn't return a Boolean like it should?

packages/core-p2p/lib/server/plugins/throttle/bucket.js
```diff
@@ -0,0 +1,81 @@
...
+  remove (ip) {
+    delete this.limiters[ip]
+  }
+
+  /**
+   * Decrement the number of tokens by the given amount.
+   *
+   * @param  {String} ip
+   * @param  {Number} amount
+   * @return {Boolean}
```
alexbarnsley: Returns void

PR #802, opened 2018-08-02
Fix forged casting

alexbarnsley:
## Proposed changes
There are issues when applying a block to a wallet. I've done casting to number in a few places to ensure values get incremented and not concatenated.

@faustbrian can you confirm whether bignum is necessary here please.

## Types of changes

## Checklist


faustbrian:
Just for `this.balance += +block.reward + +block.totalFee` should be fine until we hit a value higher then `9007199254740991` but we will work on that BigNumber replacement soon.

PR #803, opened 2018-08-02
Peer common blocks

alexbarnsley:
## Proposed changes
Matching similar functionality to v1. It makes sure the node I'm connecting to download blocks has blocks in common with myself.

## Types of changes

## Checklist


fix:
sounds good, however i believe that getting recent block ids from database is killing performance (as for v1).

I would rather keep in memory the last chained blocks, and serve them

fix:
also hasCommonBlock returning only one id should be enough imo

alexbarnsley:
@fix you've mis-read the PR. it only gets block ids from database if they're not already stored in memory

https://github.com/ArkEcosystem/core/pull/803/files#diff-16040e58e4605390f4d958931bd9026cR592
https://github.com/ArkEcosystem/core/pull/803/files#diff-2399730a6203f0d874150192f6169caaR383

PR #809, opened 2018-08-03
feat: peer guard

faustbrian:
## Proposed changes

A Peer Guard that handles suspensions and all logic related to deciding if a peer is good or bad.

## Types of changes

## Checklist

### DIFF
--- a/packages/core-p2p/lib/guard.js
+++ b/packages/core-p2p/lib/guard.js
@@ -10,8 +10,9 @@ class Guard {
     this.suspensions = {}
   }
 
-  init (config) {
+  init (config, monitor) {
     this.config = config
+    this.monitor = monitor
 
     return this
   }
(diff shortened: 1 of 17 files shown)


packages/core-p2p/lib/guard.js
```diff
@@ -0,0 +1,113 @@
...
+
+    const until = moment().add(this.manager.config.suspendMinutes, 'minutes')
+
+    this.suspensions[peer.ip] = {
+      peer,
+      until,
+      untilHuman: until.format('h [hrs], m [min]')
+    }
+
+    delete this.peers[peer.ip]
```
supaiku0: This has no effect here. The `peers` only exist in `Monitor`.

alexbarnsley: :ok_hand: thanks

Issue #766, opened 2018-07-14
Add tests to core-graphql


j-a-m-l:
Currently this package doesn't have any test. Ideally all the code would be covered:

- [ ] Test API
- [ ] Test queries and relationships

luciorubeens:
I'll assign this to me.

faustbrian:
Any progress on this? Has already been 3 weeks.

JeremiGendron:
@faustbrian may I complete the GraphQL tests?

PR #813, opened 2018-08-04
server error log & stop GRAPHQL

JeremiGendron:
## Proposed changes

use Hapi native stop() function instead of process.exit on server.start() error in try{} at catch(error) {} block
use logger.error to display information
## Types of changes

- [ x ] Refactoring (improve a current implementation without adding a new feature or fixing a bug)

## Checklist

- [ x ] I have read the [CONTRIBUTING](https://docs.ark.io/developers/guidelines/contributing.html) documentation
- [ x ] Lint and unit tests pass locally with my changes

Uses the same function as plugin.deregister (server.stop()); therefore I did not add any tests for this implementation.

[packages/core-graphql/lib/index.js#L24
](https://github.com/JeremiGendron/core/blob/bf35087c785b3507d10f29a0b39fc231cb35b30e/packages/core-graphql/lib/index.js#L24)
the resolvedPlugin returns an instance of require(server) -> the hapi.server created in server.js that we call stop on if the start fails.

[hapi documentation](https://hapijs.com/api#-await-serverinitialize), why it's important:

> If you must try to resume after an error, call server.stop() first to reset the server state.

#766

### DIFF
--- a/packages/core-graphql/lib/server.js
+++ b/packages/core-graphql/lib/server.js
@@ -47,8 +47,7 @@ module.exports = async (config) => {
 
     return server
   } catch (error) {
-    logger.error(error.stack)
-    // TODO no exit here?
-    process.exit(1)
+    logger.error(`Error starting GraphQL API Server, stopping in 5 seconds: ${error.stack}`)
+    await server.stop()
   }
 }


faustbrian:
If a plugin fails to mount we should exit because it means something is wrong with the node configuration.

PR #817, opened 2018-08-06
refactor: Optimisations fix

fix:
WIP

### DIFF
--- a/packages/core-blockchain/lib/blockchain.js
+++ b/packages/core-blockchain/lib/blockchain.js
@@ -244,7 +244,7 @@ module.exports = class Blockchain {
 
   /**
    * Hande a block during a rebuild.
-   * NOTE: We should be sure this is fail safe (ie callback() is being called)
+   * NOTE: We should be sure this is fail safe (ie callback() is being called only ONCE)
    * @param  {Block} block
    * @param  {Function} callback
    * @return {Object}
@@ -284,7 +284,7 @@ module.exports = class Blockchain {
 
   /**
    * Process the given block.
-   * NOTE: We should be sure this is fail safe (ie callback() is being called)
+   * NOTE: We should be sure this is fail safe (ie callback() is being called only ONCE)
    * @param  {Block} block
    * @param  {Function} callback
    * @return {(Function|void)}
--- a/packages/core-blockchain/lib/blockchain.js
+++ b/packages/core-blockchain/lib/blockchain.js
@@ -244,7 +244,7 @@ module.exports = class Blockchain {
 
   /**
    * Hande a block during a rebuild.
-   * We should be sure this is fail safe (ie callback() is being called)
+   * NOTE: We should be sure this is fail safe (ie callback() is being called)
    * @param  {Block} block
    * @param  {Function} callback
    * @return {Object}
@@ -284,7 +284,7 @@ module.exports = class Blockchain {
 
   /**
    * Process the given block.
-   * We should be sure this is fail safe (ie callback() is being called)
+   * NOTE: We should be sure this is fail safe (ie callback() is being called)
    * @param  {Block} block
    * @param  {Function} callback
    * @return {(Function|void)}
--- a/packages/core-blockchain/lib/blockchain.js
+++ b/packages/core-blockchain/lib/blockchain.js
@@ -244,6 +244,7 @@ module.exports = class Blockchain {
 
   /**
    * Hande a block during a rebuild.
+   * We should be sure this is fail safe (ie callback() is being called)
    * @param  {Block} block
    * @param  {Function} callback
    * @return {Object}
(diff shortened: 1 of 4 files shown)


alexbarnsley:
Your changes now make it so that it gets recent blocks from the database every time a block is downloaded, instead of the in-memory array.

fix:
yes but there is no garantee about the state of this.recentBlockId
there is nowhere in the code when ids are removed when a block is reverted for instance

fix:
this will provoke the network to fork hard

alexbarnsley:
So why not add in a "force from database" and run that once all blocks are reverted. you complained about common blocks causing slowness in my PR, yet this will make it worse. just my $0.02.

fix:
well i prefer slow and not forking than fast and forking

fix:
also, i think these 'last blocks' should be managed centrally at the level of te state machine (together with other stateful data)

fix:
Also essentially keeping last 100s full blocks in memory will help syncing fast, most peers will only download last 100 blocks (in worst case scenario when removing blocks)

packages/core-blockchain/lib/blockchain.js
```diff
@@ -263,80 +264,81 @@ module.exports = class Blockchain {
...
     } else {
       logger.warn(`Block ${block.data.height.toLocaleString()} disregarded because verification failed :scroll:`)
-      callback()
+      return callback()
     }
   }
 
   /**
    * Process the given block.
+   * We should be sure this is fail safe (ie callback() is being called)
```
faustbrian: This is a `TODO:`.

packages/core-database-sequelize/lib/connection.js
```diff
@@ -588,18 +588,14 @@ module.exports = class SequelizeConnection extends ConnectionInterface {
    * @return {[]String}
    */
   async getRecentBlockIds () {
-    if (!this.recentBlockIds.length) {
```
faustbrian: This is a big hit on performance as already tested in the past. Constantly hitting the blocks table for a read operation is resource intense and can easily result in a deadlock.

fix:
adding an issue to track this down and clarify this

packages/core-blockchain/lib/blockchain.js
fix: no i have made it failsafe, putting it as a reminder if somebody touch it

packages/core-blockchain/lib/blockchain.js
```diff
@@ -263,80 +264,81 @@ module.exports = class Blockchain {
...
    * @return {(Function|void)}
    */
   async processBlock (block, callback) {
     if (!block.verification.verified) {
       logger.warn(`Block ${block.data.height.toLocaleString()} disregarded because verification failed :scroll:`)
-
       return callback()
     }
 
-    const state = stateMachine.state
```
faustbrian: Missing tests for the changes from here to `return this.dispatch('FORK')`.

packages/core-database-sequelize/lib/connection.js
fix: see my comment for @alexbarnsley this.recentBlockIds as no garantee for being in sync with the state of the blockchain

packages/core-blockchain/lib/blockchain.js
faustbrian: Then `NOTE:` to make it clearly visible that it should be read.

packages/core-database-sequelize/lib/connection.js
faustbrian: Then `should be managed centrally at the level of the state machine` should be included in this PR as it is part of the same behaviour that is being modified.

packages/core-blockchain/lib/blockchain.js
fix: i have just put the same logic to the calling function instead to be sure this is threadsafe.
IE moving logic from `acceptChainedBlock` to `processBlock`

packages/core-database-sequelize/lib/connection.js
fix: this more complex than that, opening an issue for that, see again te discussion

packages/core-blockchain/lib/blockchain.js
fix: OK

faustbrian:
Travis is currently failing and in the next PR please respect the guidelines for commits https://docs.ark.io/developers/guidelines/git-commit-guidelines.html#types to make it easier to compile changelogs.

Also always make sure that PRs are squashed according to guidelines https://docs.ark.io/developers/guidelines/repository-management.html#squashing-pull-requests with a meaningful commit.

fix:
https://github.com/ArkEcosystem/core/issues/818

fix:
@faustbrian yo will do

PR #829, opened 2018-08-08
fix(core-api): add cors additionalHeaders

luciorubeens:
## Proposed changes
The request in browsers is refused because these headers are not allowed.

## Types of changes

## Checklist


packages/core-api/lib/server.js
```diff
@@ -13,7 +13,9 @@ module.exports = async (config) => {
     host: config.host,
     port: config.port,
     routes: {
-      cors: true,
+      cors: {
+        additionalHeaders: ['nethash', 'port', 'version', 'api-version']
```
faustbrian: `'nethash', 'port', 'version'` are `core-p2p` exclusive headers.

luciorubeens: It's used to post the transaction: https://github.com/ArkEcosystem/desktop-wallet/blob/master/client/app/src/services/network.service.js#L203

luciorubeens: and also on all requests made by the client: https://github.com/ArkEcosystem/core/blob/master/packages/client/lib/http.js#L90

faustbrian: The public API doesn't use those headers so they shouldn't be send to it, only the P2P API needs those for identification because the P2P API is for communication between nodes which always send those headers.

All the endpoints and code you are references is based on v1 which is the P2P API, not public.

luciorubeens: got it, so should we remove these headers from all clients or add the compatibility here?

faustbrian: Outside of the JavaScript Client no other SDKs have support for the P2P API as the only reason to use the P2P API in v1 was that there was no public endpoint to post transactions which is resolved by the v2 Public API.

If you work on the v2 Wallet you should be using the `POST /api/v2/transactions` endpoint to post transactions to a relay.

luciorubeens:
@faustbrian I removed these headers. Please, take a look.

PR #831, opened 2018-08-08
Detect open handles

j-a-m-l:
## Proposed changes
The `--detectOpenHandles` option is useful for finding possible reasons of why Jest can't finished cleanly.

I've moved the commit from https://github.com/ArkEcosystem/core/pull/783, to split that PR into smaller autonomous things


j-a-m-l:
It looks like, at least in this case, after splitting #783, Jest / Travis waits a lot more to finish. Probably the best option then, is forcing the exit on Travis while those problems aren't fixed.

faustbrian:
A few conflicts that need to be resolved.

PR #833, opened 2018-08-08
Stop blockchain

j-a-m-l:
## Proposed changes

The intention of this changes is creating a way to stop the blockchain. It'd be used on tests to:
 1. Reduce the moving parts: running the blockchain on some tests produces, sometimes, side effects. This could be used to investigate them.
 2. Increase the speed. Stopping and re-starting the blockchain would be faster than creating it on every test.

## Types of changes

I've moved the commit from https://github.com/ArkEcosystem/core/pull/783, to split that PR into smaller autonomous things


packages/core-blockchain/lib/index.js
```diff
@@ -11,12 +12,15 @@ exports.plugin = {
...
+    blockchain = new Blockchain(options.networkStart)
 
     if (!process.env.ARK_SKIP_BLOCKCHAIN) {
       await blockchain.start()
     }
 
     return blockchain
+  },
+  async deregister (container, options) {
+    await blockchain.stop()
```
faustbrian: No need for that wonky variable solution, can simply do `await container.resolve('blockchain').stop()`.

j-a-m-l: Thanks @faustbrian, I've updated it with your suggestion

faustbrian:
Could you run the tests locally? Travis has a lot of timeout issues the last few days and everytime I run the blockchain tests some random one fails.

j-a-m-l:
I did it and 13 tests we're failing in both cases.

j-a-m-l:
All tests are OK

alexbarnsley:
@j-a-m-l I did remove this comment but re-adding. It's not major but I think it looks better with your STOP fix, just doesn't seem to get triggered during a sync for me

>i just re-tested this. Looks like it doesn't stop during a sync on devnet:
>
>My local forging node (in sync):
>image
>
>My devnet node (sync in progress):
>image
>
>The `The blockchain has been stopped` message doesn't show during

PR #839, opened 2018-08-09
fix: sort peers by latency, not random

alexbarnsley:
## Proposed changes
It's currently fairly useless to sort peers randomly. Due to the pagination used, peers are randomly sorted and then cropped at the page limit. This could result in getting the X amount of bad peers, and never a peer with a low latency, because every page request (e.g. page 2 of the peer list) will result in a new completely random set of peers. 

I understand this is likely to prevent some sort of attack, but I can think of 3 easy ways to get round this and no easy solution for any of them:

1. send necessary headers and access the p2p endpoint `:4002/peer/list` which has the full list
2. spam the public peer endpoint `:4003/api/v2/peers` a few hundred times until i've got a good chunk of the list
3. run my own node and have access to them all in any way i want

I don't see the benefit of randomising the peer list here. @fix	i'd like your input

## Types of changes

## Checklist

### DIFF
--- a/packages/core-api/lib/versions/1/handlers/peers.js
+++ b/packages/core-api/lib/versions/1/handlers/peers.js
@@ -22,14 +22,12 @@ exports.index = {
       return utils.respondWith('No peers found', true)
     }
 
-    let peers = allPeers.sort(() => 0.5 - Math.random())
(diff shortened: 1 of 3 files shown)


zillionn:
Yes, there is probably no point of randomizing the list, because If you can get the full list, the attacker can.

PR #848, opened 2018-08-11
[WIP] Circle ci configuration

j-a-m-l:
I'm just testing Circle This PR is part of it, but it would include the final configuration when it's ready.


j-a-m-l:
We're going to use a different Travis plan, so this isn't necessary now.

PR #816, opened 2018-08-06
Graphql/tests minimum viability

JeremiGendron:
## Proposed changes

#766 there is no testing for GraphQL package:
add single test file for graphql
add __support__ for test setup
yarn test is long but passes: expect(graphql).toBe(graphql/object)

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [ x ] New feature (non-breaking change which adds functionality)
- [ x ] Refactoring (improve a current implementation without adding a new feature or fixing a bug)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Build (changes that affect the build system)
- [ ] Docs (documentation only changes)
- [ x ] Test (adding missing tests or fixing existing tests)
- [ ] Other... Please describe:

## Checklist

- [ x ] I have read the [CONTRIBUTING](https://docs.ark.io/developers/guidelines/contributing.html) documentation
- [ x ] Lint and unit tests pass locally with my changes
- [ x ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)

## Further comments

I will further work on this branch through more branches, schedule:
- Have custom config to skip unused plugins during container setup
- Test every function's minimum viability
- Learn the API and develop more extensive tests
- Work on improving the API if needed

PR #854, opened 2018-08-13
Nice GQL tests config

JeremiGendron:
## Proposed changes

Goal was to optimize the tests to remove unrelated plugins, the required ones are Logger, Logger-Winston and I've found it easier to set a local variable with the require('../lib').register(container, options) as the graphql plugin rather than include it into the config/plugins.js copied from core/config/testnet/* (I was getting undefined issues cause I suck at async) -- now tests pass in a few seconds.

There's an issue with tests running indefinitely if you don't use forceExit which I think the core-container causes (can't see why most plugins would return the same error without --detectOpenHandles and no --forceExit) so I've changed the package.json file to use both and --runInBand (don't know what --rIB does, saw it in some other plugins, thought it was "good").

Tried to remove as much redundancy and will be working on this further.
## Types of changes

- [ x ] Bugfix (non-breaking change which fixes an issue)
- [ x ] Refactoring (improve a current implementation without adding a new feature or fixing a bug)
- [ x ] Test (adding missing tests or fixing existing tests)

## Checklist

- [ x ] I have read the [CONTRIBUTING](https://docs.ark.io/developers/guidelines/contributing.html) documentation
- [ x ] Lint and unit tests pass locally with my changes (somereason I'm getting trolled)
- [ x ] I have added tests that prove my fix is effective or that my feature works
- [ x ] I have added necessary documentation (if appropriate)

In line with #816


packages/core-graphql/package.json
```diff
@@ -9,7 +9,7 @@
   "main": "lib/index.js",
   "scripts": {
     "build:docs": "../../node_modules/.bin/jsdoc -c jsdoc.json",
-    "test": "cross-env ARK_ENV=test jest --runInBand --detectOpenHandles",
+    "test": "cross-env ARK_ENV=test jest --forceExit --runInBand --detectOpenHandles",
```
faustbrian: `detectOpenHandles` replaces `forceExit`

JeremiGendron:
Ok, sorry I'll read some jest, I'll remove detectOpenHandles too

packages/core-graphql/package.json
JeremiGendron: Thanks

faustbrian:
Also please use `[x]` instead of `[ x ]` in your PRs or the to-do lists won't work. You can see that it doesn't render checkboxes.

packages/core-graphql/package.json
```diff
@@ -9,7 +9,7 @@
   "main": "lib/index.js",
   "scripts": {
     "build:docs": "../../node_modules/.bin/jsdoc -c jsdoc.json",
-    "test": "cross-env ARK_ENV=test jest --forceExit --runInBand --detectOpenHandles",
+    "test": "cross-env ARK_ENV=test jest --runInBand --forceExit",
```
faustbrian: This is still wrong, `forceExit` is replaced by `detectOpenHandles`.

JeremiGendron: what do you suggest?

faustbrian: As I said, use `--detectOpenHandles` instead of `--forceExit`.

JeremiGendron: works locally now with proper plugin management (doesnt hang with `detectopenhandles`)

PR #858, opened 2018-08-14
feat: log usernames in forger (via P2P API)

faustbrian:
## Proposed changes

Adds the ability to log usernames in the forger as it usually doesn't have access to that kind of data. Currently the forger only logs public keys which makes testing really tedious unless you want to remember every public key in existence.

## Types of changes

## Checklist


boldninja:
puppy

PR #871, opened 2018-08-16
feat: blockchain ready plugin for P2P API

faustbrian:
## Proposed changes

This adds a plugin for the P2P API that will return a 503 Service Unavailable on certain endpoints that require the blockchain to be ready. For v1 endpoints it will return 200 Success with a body of `{ success: false }` because that is how it used to be handled in the legacy P2P API where there is no proper use of status codes.

## Types of changes

## Checklist


faustbrian:
@alexbarnsley 👁

PR #881, opened 2018-08-18
Fix/db interface presentation

JeremiGendron:
## Proposed changes

Use of less ambiguous term for current round's active delegates as computed in the database and sequelize plugins.

Thought it would be overkill to provide a patch for the sequelize plugin, as it is supposed to undergo a restructure in favor of the database plugin.

Please correct me !
## Types of changes

- [x Refactoring (improve a current implementation without adding a new feature or fixing a bug)

## Checklist


faustbrian:
Will leave it as is for now because there are many more occurrences of this then you renamed.

Issue #886, opened 2018-08-19
Forger with 8 relays


Jarunik:
**Describe the bug**
Forger with 8 relays will fail with fork 6.

**To Reproduce**
1. Setup 8 relays
2. Whitelist the forger on all 8 relays
3. Put all 8 relays as hosts on the forger
4. Wait for Forging
5. Fork 6

**Expected behavior**
Forging works.

**Screenshots**
n/a

**Server (please complete the following information):**
 - OS: Ubuntu
 - Version 18.04]

**Additional context**

```
2|ark-core-forger  | [2018-08-19 10:19:11][DEBUG]                      : Loaded jarunik (035ee26e3c3597652cca2497642ade9651569815b8ef697415cbc70e828417bd53) delegates.
2|ark-core-forger  | [2018-08-19 10:19:11][INFO]                       : ForgerManager started with 1 forgers
2|ark-core-forger  | [2018-08-19 10:22:50][INFO]                       : Next forging delegate jarunik (035ee26e3c3597652cca2497642ade9651569815b8ef697415cbc70e828417bd53) is active on this node.
2|ark-core-forger  | [2018-08-19 10:22:50][DEBUG]                      : Sending wake-up check to relay node(s) http://127.0.0.1:4002
2|ark-core-forger  | [2018-08-19 10:22:51][DEBUG]                      : Sending wake-up check to relay node(s) http://45.76.94.199:4002
2|ark-core-forger  | [2018-08-19 10:22:51][DEBUG]                      : Sending wake-up check to relay node(s) http://104.238.159.113:4002
2|ark-core-forger  | [2018-08-19 10:22:52][DEBUG]                      : Sending wake-up check to relay node(s) http://149.28.165.214:4002
2|ark-core-forger  | [2018-08-19 10:22:52][DEBUG]                      : Sending wake-up check to relay node(s) http://66.42.60.249:4002
2|ark-core-forger  | [2018-08-19 10:22:52][DEBUG]                      : Sending wake-up check to relay node(s) http://198.13.38.68:4002
2|ark-core-forger  | [2018-08-19 10:22:53][DEBUG]                      : Sending wake-up check to relay node(s) http://104.238.173.32:4002
2|ark-core-forger  | [2018-08-19 10:22:53][DEBUG]                      : Sending wake-up check to relay node(s) http://45.77.61.218:4002
[...]

PR #887, opened 2018-08-19
fix: sending too many requests to relays in forger

faustbrian:
## Proposed changes

The forger was sending wake-up signals to all relays instead of choosing one which could result in a lot of requests and missing blocks in certain cases. Resolves https://github.com/ArkEcosystem/core/issues/886

## Types of changes

## Checklist

### DIFF
--- a/packages/core-forger/lib/client.js
+++ b/packages/core-forger/lib/client.js
@@ -48,8 +48,8 @@ module.exports = class Client {
 
     try {
       await this.__get(`${this.host}/internal/syncCheck`)
-    } catch (e) {
-      logger.error(e.message)
+    } catch (error) {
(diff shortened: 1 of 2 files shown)


packages/core-forger/lib/client.js
```diff
@@ -43,15 +42,15 @@ module.exports = class Client {
...
-      }
-    })
+    await this.__chooseHost()
+
+    logger.debug(`Sending wake-up check to relay node ${this.host}`)
+
+    try {
+      await this.__get(`${this.host}/internal/syncCheck`)
+    } catch (e) {
+      logger.error(e.message)
```
alexbarnsley: Is it worth a prefix message to the error here? E.g. `Could not sync check: ${e.message}`

Issue #890, opened 2018-08-20
Forger events not emitted


faustbrian:
**Describe the bug**
Forger events currently are not emitted on the same emitter

**To Reproduce**
Setup a listener in any plugin, start the relay and forger separately. The relay will never receive the forger events.

**Expected behavior**
Forger event should be emitted and be received by 3rd party plugins and all other listeners setup.

**Proposed Solution**
Implement a P2P remote endpoint to emit events from the relay rather then the forger.

PR #891, opened 2018-08-20
feat: internal event emitting via P2P API

faustbrian:
## Proposed changes

When the forger and relay were launched separately the emitted events would never reach the relay as the relay and forger were isolated. This will emit the given event via API to make sure the relay will be notified.

Resolves https://github.com/ArkEcosystem/core/issues/890.

## Types of changes

## Checklist


packages/core-forger/lib/client.js
```diff
@@ -39,6 +39,35 @@ module.exports = class Client {
...
+   * @param  {String} event
+   * @param  {Object} body
+   * @return {Object}
+   */
+  async emitEvent (event, body) {
+    // NOTE: Events need to be emitted to the localhost. If you need to trigger
+    // actions on a remote host based on events you should be using webhooks
+    // that get triggered by the events you wish to react to.
+
+    const host = this.hosts.find(host => host.includes('127.0.0.1'))
```
alexbarnsley: it might be worth including all localhost IPs here:

alexbarnsley: https://github.com/ArkEcosystem/core/blob/develop/packages/core/lib/config/devnet/plugins.js#L98

packages/core-forger/lib/manager.js
```diff
@@ -145,7 +144,9 @@ module.exports = class ForgerManager {
    * @param {Object} round
    */
   async __forgeNewBlock (delegate, round) {
-      emitter.emit('forger.started', delegate.publicKey)
+      // NOTE: Disabled for now as this could cause a delay in forging that
```
alexbarnsley: I think there should be a TODO here, to work on not missing the blocks and re-enabling the emitter

Issue #852, opened 2018-08-12
Fork issues with common blocks


alexbarnsley:
**Describe the bug**
Currently if a forked node cannot find any peers with a common block, it crashes (see below image). This prevents a rollback at all, and means a rebuild is necessary. In order to fix this, I think some sort of "max attempts" is required, and once that is hit it needs to establish what block it needs to rollback to.

**To Reproduce**
Begin syncing a forging node with the chain. Once established a few rounds, stop the node and then run in test mode (`ARK_ENV=test ./bin/ark start ...`). This will forge on its own chain (it has to be a forging node). Make sure at least 51 blocks are forged, then stop the node and start forging on the chain again (without `ARK_ENV=test`).

**Expected behavior**
It establishes what block to rollback to and does so.

**Screenshots**
![image](https://user-images.githubusercontent.com/8069294/44006573-93eb27de-9e7e-11e8-8fe1-39b3ddcfee77.png)

**Server (please complete the following information):**
 - OS: MacOS
 - Version: High Sierra (10.13.3)

**Additional context**
In my case I am running a 5 forger, 2 node setup. I believe the issue to still apply here.

fix:
mmm we are using here a recursive function to find a random peer. if for some reasons, the node is disconnected, it loops recursively until recursive call stack maximum is reached.

I have seen this a few times. I am thinking about the following fix:
- de-recursify the getRandomPeer using a while loop
- perform a maximum number of tried before raising an error ('could not find peer')
- launch a network audit (ntp/dns connection test)

alexbarnsley:
Hmm, i may have fixed this already @fix:

https://github.com/ArkEcosystem/core/commit/963f17923248adbfa08fc7c5b553871f03a6ed84#diff-5457b1a950959ee566c3948a47b74735R224

https://github.com/ArkEcosystem/core/commit/963f17923248adbfa08fc7c5b553871f03a6ed84#diff-6c9d91045605634e8224209e0a190b82R55

excluding the audit part. It tries random peers upto 10 times, then throws an exception. Might need updating/fine-tuning so can leave this issue open if you like?

PR #900, opened 2018-08-21
refactor: p2p complexity

faustbrian:
## Proposed changes

The complexity of the P2P Interface was quite high **(and still is)** so this PR tries to reduce this a bit.

## Types of changes

## Checklist


packages/core-p2p/lib/monitor.js
```diff
@@ -15,54 +15,62 @@ const Peer = require('./peer')
...
     try {
       // TODO: for tests that involve peers we need to sync them
       if (process.env.ARK_ENV !== 'test') {
         await this.discoverPeers()
         await this.cleanPeers()
       }
 
-      if (Object.keys(this.peers).length < this.config.peers.list.length - 1 && process.env.ARK_ENV !== 'test') {
-        this.config.peers.list
+      if (Object.keys(this.peers).length < config.peers.list.length - 1 && process.env.ARK_ENV !== 'test') {
```
alexbarnsley: Shouldn't this be `this.config`? I don't see config passed into updateNetworkStatus, or declared outside of the class.

packages/core-p2p/lib/manager.js
```diff
@@ -1,198 +0,0 @@
...
-  async updateNetworkStatus () {
-    await this.monitor.updateNetworkStatus()
-  }
-
-  /**
-   * Download blocks from a random peer.
-   * @param  {Number}   fromBlockHeight
-   * @return {Object[]}
-   */
-  async downloadBlocks (fromBlockHeight) {
```
alexbarnsley: This needs moving to `monitor.js` as it catches an exception in getRandomPeers, or at least it used to. Related to this Exception spam: https://github.com/ArkEcosystem/core/issues/852

alexbarnsley:
I meant to comment - not tested yet, as I think there may be a couple of issues. Check the config => `this.config` comment i made, then I will test whether the downloadBlocks try/catch removal is still an issue.

packages/core-p2p/lib/monitor.js
faustbrian: `config` is the `container.resolvePlugin('config')`

packages/core-p2p/lib/manager.js
faustbrian: Will check that later, should have been fine as it handles the exception itself already 🤔

faustbrian: The `getRandomDownloadBlocksPeer` was throwing the exception so I moved it inside the `try catch` block in `downloadBlocks`.

packages/core-p2p/lib/monitor.js
alexbarnsley: I guess i needed sleep, I looked for it at the top and didn't see it 🤦‍♂️

packages/core-p2p/lib/manager.js
alexbarnsley: Will test now, thanks!

PR #903, opened 2018-08-21
Tests/graphql query skel

JeremiGendron:
## Proposed changes

Ensure the queries are properly accessible, write the skeleton for graphql query tests, remove history until it has more value.

Everything passes locally, next step is adding graphql-tools to the package and performing tests with mockFunctions on the schema found in lib/schema.

Continuing more work on this...

also added documentation for newer devs who might be interested in working on this.

## Types of changes

## Checklist


faustbrian:
The tests contain way to many comments, if you need that many comments the tests are not clear enough to show their intent by itself.

packages/core-graphql/__tests__/graphql.test.js
```diff
@@ -1,51 +1,146 @@
...
+        logger.debug('Proceeding with Block test queries')
+
+        testQueries.block.forEach((block) => {
+          logger.info(`Processing Block test query with id: ${block.id}`)
+        })
+      })
+    })
+
+    describe('test query for Blocks', () => {
+      it('Should have queries', () => {
```
faustbrian: `it('Should` should always be `it('should` so it reads like a sentence and is consistent with all other occurrences of `it('should` throughout tests.

faustbrian:
@JeremiGendron once you removed all comments in tests and adjusted the `it should` I will merge, otherwise the PR works.

JeremiGendron:
On it

JeremiGendron:
@faustbrian I have available progress, would like this merged

packages/core-graphql/__tests__/graphql.test.js
```diff
@@ -28,6 +27,79 @@ describe('GraphQL', () => {
...
     it('should be an object', () => {
       expect(testQueries).toBeObject()
+      logger.info('Processing test queries')
+    })
+
+    describe('test query for Block', () => {
+      it('should have queries', () => {
+        expect(testQueries.block.length).toBeTruthy()
+
+        logger.debug('Proceeding with Block test queries')
```
faustbrian: The logger is a dependency of core and should not be used inside tests. Logging inside tests should be avoid in general.

JeremiGendron: thanks, changing and pushing

JeremiGendron:
@faustbrian dont think theres a conflict with cors file which snuck into there, can fix if needed

PR #904, opened 2018-08-22
Keeping the relay that responded to syncCheck and syncCall - until forging is done.

kristjank:
## Proposed changes
Using the relay that responded to synCheck and making sure the relay is synced. Also no host choosing between getting network state and transaction from pool. Saving time and sticking with the host that responded and is synced.

## Types of changes

## Checklist


kristjank:
@faustbrian

packages/core-forger/__tests__/client.test.js
```diff
@@ -79,6 +85,10 @@ describe('Client', () => {
     })
 
     describe('when the host is available', () => {
+      beforeEach(async () => {
```
faustbrian: There is only 1 test in this `describe` block, `await client.__chooseHost()` should be called inside the `it` block.

packages/core-forger/__tests__/client.test.js
```diff
@@ -98,6 +108,10 @@ describe('Client', () => {
...
```
faustbrian: There is only 1 test in this `describe` block, `await client.__chooseHost()` should be called inside the `it` block.

packages/core-forger/__tests__/client.test.js
```diff
@@ -45,7 +45,13 @@ describe('Client', () => {
...
```
faustbrian: There is only 1 test in this `describe` block, `await client.__chooseHost()` should be called inside the `it` block.

Jarunik:
I will try set it up again on new devnet and test there. Already killed nodes on current devnet.

PR #905, opened 2018-08-22
fixes on rebuilding mainnet

fix:
Query has changed with code updates.

One potential issue: balance = NULL may have different reaction over sequelize implementation for different db


faustbrian:
**mainnet**

```
[2018-08-22 11:38:10][ERROR]   : Refused new block {"version":0,"timestamp":16368,"height":715,"previousBlockHex":"bdb1664e0ed95428","previousBlock":"13668818829488772136","numberOfTransactions":0,"totalAmount":0,"totalFee":0,"reward":0,"payloadLength":0,"payloadHash":"e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855","generatorPublicKey":"03afac7e796d781e17600945010be34cb6760fa919be67baec2cfb691cf4ce5f33","blockSignature":"304402206853523fe20927efd28a0ed9f0d5baecc40389e39acbff37a0767834cce0ac2002200d4a2a3c83f3b080ffdf55e0fc69a350f7f8e1c8faf16358a7fcc13bb61bb9c8","idHex":"9d1955cc86cb0e54","id":"11320173475295596116"}
[2018-08-22 11:38:10][DEBUG]   : Error: Delegate genesis_42 (03afac7e796d781e17600945010be34cb6760fa919be67baec2cfb691cf4ce5f33) not allowed to forge, should be genesis_40 (03bc7be5c8b1221beb63f6ed23d4b5e2748b64983087986c4953579fa7d9022a71) 👎
    at SequelizeConnection.validateDelegate (/Users/brian/Playbox/p2p/packages/core-database/lib/interface.js:361:13)
[2018-08-22 11:38:10][DEBUG]   : event 'FORK': {"syncWithNetwork":"downloadBlocks"} -> {"fork":"analysing"} -> actions: ["startForkRecovery","analyseFork"]
```

**devnet**
```
[2018-08-22 11:38:59][ERROR]   : Refused new block {"version":0,"timestamp":43197234,"height":154,"previousBlockHex":"e38a2122afd3ff4d","previousBlock":"16395953826258222925","numberOfTransactions":0,"totalAmount":0,"totalFee":0,"reward":0,"payloadLength":0,"payloadHash":"e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855","generatorPublicKey":"02fdcb7706624aba3c050c19bdd05f74a0c746614462584dfbba3705810ba2d69c","blockSignature":"3045022100cfd9d2f5c34306faf3a90d6f19da3320a9fc0ec4a2e3989c58ee792f7646cc3e02205d9ecd4ee2c62d3aa1fb57f762acbd6c3ede921034076b2938c012f367ff6c47","idHex":"83d87b65b7e77139","id":"9500479090745372985"}
[2018-08-22 11:38:59][DEBUG]   : Error: Delegate genesis_29 (02fdcb7706624aba3c050c19bdd05f74a0c746614462584dfbba3705810ba2d69c) not allowed to forge, should be genesis_18 (02ff046250cc01d4ccb74fcf6ff9b6977f5a67539cd9c461f9d41436677f4035b7) 👎
    at SequelizeConnection.validateDelegate (/Users/brian/Playbox/p2p/packages/core-database/lib/interface.js:361:13)
[2018-08-22 11:38:59][DEBUG]   : event 'FORK': {"syncWithNetwork":"downloadBlocks"} -> {"fork":"analysing"} -> actions: ["startForkRecovery","analyseFork"]
```

fix:
i think you need to wait for rebuild, or wipe out the db

faustbrian:
This is a sync from 0 in slow mode for both networks, there is no previous data.

fix:
i have no issue on mainnet.

on devnet, the code was wrong so there is no way it can sync

fix:
using pg on mainnet?

faustbrian:
Yes, no one besides you is using sqlite to test any changes as no one is going to use it.

fix:
i guess this is linked to the way balance = null is ordered then

fix:
reproduced on pg

alexbarnsley:
Personally I vote we get rid of sqlite as an option. There's too many inconsistencies we have to cater for. If someone wants sqlite they should do a new plugin to support it, that way any bugs are related solely to sqlite.

faustbrian:
Yeah, we have to handle way to many inconsistencies at this point and could make use of something like https://github.com/vitaly-t/pg-promise which would also be faster because we can have pre-made queries in files instead of having to rely on a query builder.

fix:
ok the new commit make it working on pg

As for sqlite, I would like to keep it for relay nodes. But in parallel we could develop a core-database-pg plugin and switch on it when needed

fix:
also sqlite tested

faustbrian:
`develop a core-database-pg plugin and switch on it when needed` makes it sound like postgres should be the second choice for people. We should promote postgres as the primary solution, SQLite was never designed to store the amount of data a blockchain stores long-term and it already constantly goes into a deadlock while syncing. On top of that nodes that use SQLite can go into a deadlock if to many API requests come in as opposed to Postgres who will handle that without any issues.

Postgres should be our primary solution and things like sqlite can be developed by the community if they want to use it for whatever reason.

PR #907, opened 2018-08-22
fix: post transaction endpoint to closer match v1

alexbarnsley:
## Proposed changes
Currently the response from `/peer/transactions` always returns successful even with invalid transactions. This changes that, and brings the values more in-line with the v1 endpoint.

## Types of changes

## Checklist


faustbrian:
https://github.com/ArkEcosystem/core/pull/907/files#diff-3b486d98232d78d103cca9bd1a6a338aR340

This line should do `guard.accept.concat(guard.excess)` or do `if (guard.hasAny('invalid') || guard.hasAny('excess')) {`, otherwise we are inconsistent with v1 behaviour for accepting transactions.

PR #910, opened 2018-08-22
intended way to upgrade network config

fix:
That is the way to perform softork/hardfork:
- fix a block height
- add the new or upgraded parameters to be used by the node


fix:
by the way a few config update will have some side effects not taken into account yet into the way things are calculated:
- anything related to blocktime change
- if number of active delegates change at height different than a multiple of round size.

PR #913, opened 2018-08-23
peer/blocks should at least return the last block, if the blockHeight…

### LINKED ISSUE #912: peer/blocks fails ungracefully if 'lastBlockHeight' param is not specified
**Describe the bug**
The api call /peer/blocks expects a lastBlockHeight param to be passed in. If you don't however, the API should either, return meaningful data or fail gracefully. Right now it fails with a SQL error because it tries to query with an invalid value.

**To Reproduce**
Invoke /peer/blocks with no 'lastBlockHeight' parameter

**Expected behavior**
Return at least, the last block(or an empty block list if there's none)

**Screenshots**
peer_blocks_lastblockheightfailure

**Server (please complete the following information):**
Ubuntu 18

paroxysm:
## Fixes #912 
In an effort to increase the robustness of the API, I've modified /peer/blocks to return the last block if its called w/o a 'lastBlockHeight' parameter.

The other alternative would be have it fail w/ a validation error saying 'lastBlockHeight' param is required. However, that's a high fidelity change(other endpoints are also affected if we implement a validation middleware)  w/ little value.

## Types of changes

## Checklist

### DIFF
--- a/packages/core-p2p/lib/server/versions/1/handlers.js
+++ b/packages/core-p2p/lib/server/versions/1/handlers.js
@@ -359,12 +359,12 @@ exports.getBlocks = {
    */
   async handler (request, h) {
     try {
-      let database = container.resolvePlugin('database')
+      const database = container.resolvePlugin('database')
+      const blockchain = container.resolvePlugin('blockchain')
       let reqBlockHeight = parseInt(request.query.lastBlockHeight)
       let blocks = []
       if (!request.query.lastBlockHeight || Number.isNaN(reqBlockHeight)) {
-        blocks.push(await database.getLastBlock())
-
+        blocks.push(blockchain.getLastBlock())
       } else {
         blocks = await database.getBlocks(parseInt(reqBlockHeight) + 1, 400)
       }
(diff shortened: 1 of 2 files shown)


packages/core-p2p/lib/server/versions/1/handlers.js
```diff
@@ -359,7 +359,15 @@ exports.getBlocks = {
    */
   async handler (request, h) {
     try {
-      const blocks = await container.resolvePlugin('database').getBlocks(parseInt(request.query.lastBlockHeight) + 1, 400)
+      let database = container.resolvePlugin('database')
+      let reqBlockHeight = parseInt(request.query.lastBlockHeight)
+      let blocks = []
+      if (!request.query.lastBlockHeight || Number.isNaN(reqBlockHeight)) {
+        blocks.push(await database.getLastBlock())
```
faustbrian: You should avoid calling the database as the last block there could be wrong as data is periodically stored.

`blocks.push(blockchain.getLastBlock())`

faustbrian: `My intention was to return the last block at the time of API call.` then you have to use `blocks.push(blockchain.getLastBlock())` as I said in my comment because the database will have older data. You should always refer to in-memory state for `getLast*` data.

paroxysm: Ah I see, yes sir I'd prefer that approach too.

PR #914, opened 2018-08-23
Feat: Limiting the number of transactions sent in one request

kristjank:
## Proposed changes
Limiting the size of transaction allowed in one request.

## Types of changes

## Checklist

## Further comments
there could be a better place for this thru the API, @faustbrian your feedback needed


faustbrian:
This should be done in the transaction pool configuration rather then the network configuration.

packages/core-api/lib/versions/2/handlers/transactions.js
```diff
@@ -40,12 +40,17 @@ exports.store = {
...
   async handler (request, h) {
-    if (!transactionPool) {
+    const blockchain = container.resolvePlugin('blockchain')
+    if (!transactionPool || !blockchain) {
       return {
         data: []
       }
     }
 
+    if (request.payload.transactions.length > transactionPool.options.maxTransactionsPerRequest) {
```
faustbrian: This is a v2 endpoint, we are not using those ghetto v1 solutions here.

You will have to update the joi schema to be limited to the configured value.

https://github.com/ArkEcosystem/core/blob/develop/packages/core-api/lib/versions/2/schema/transactions.js#L18

packages/core-transaction-pool-redis/lib/defaults.js
```diff
@@ -5,6 +5,7 @@ module.exports = {
   key: 'ark',
   maxTransactionsPerSender: process.env.ARK_TRANSACTION_POOL_MAX_PER_SENDER || 100,
   allowedSenders: [],
```
faustbrian: Missing the `whitelist` option.

kristjank:
@faustbrian

packages/core-api/lib/versions/2/handlers/transactions.js
```diff
@@ -40,17 +40,12 @@ exports.store = {
    * @return {Hapi.Response}
    */
   async handler (request, h) {
-    const blockchain = container.resolvePlugin('blockchain')
-    if (!transactionPool || !blockchain) {
+    if (!transactionPool) {
       return {
```
faustbrian: This should return a `Boom.serverUnavailable('Transaction Pool not ready')`.

packages/core-api/lib/versions/2/handlers/transactions.js
```diff
@@ -40,17 +40,12 @@ exports.store = {
    * @return {Hapi.Response}
    */
   async handler (request, h) {
-    const blockchain = container.resolvePlugin('blockchain')
-    if (!transactionPool || !blockchain) {
+    if (!transactionPool) {
```
faustbrian: Actually this whole block can be removed, the API will only be mounted after the pool and blockchain are ready.

packages/core-api/lib/versions/2/handlers/transactions.js
```diff
@@ -41,7 +41,7 @@ exports.store = {
    */
   async handler (request, h) {
     if (!transactionPool.options.enabled) {
-      return Boom.teapot()
+      return Boom.teapot('Transaction pool is disabled.')
```
faustbrian: I said `serverUnavailable` in my previous comment, not `teapot` as that doesn't give a clear indication of what is wrong just based on the status code.

kristjank:
done as suggested

packages/core-api/lib/versions/2/handlers/transactions.js
kristjank: done

packages/core-api/lib/versions/2/schema/transactions.js
```diff
@@ -3,6 +3,9 @@
 const Joi = require('joi')
 const pagination = require('./pagination')
 
+const container = require('@arkecosystem/core-container')
+const transactionPool = container.resolvePlugin('transactionPool')
```
faustbrian: If you don't need the full plugin but only it's options you can use `container.resolveOptions('transactionPool')`.

packages/core-p2p/lib/server/versions/1/handlers.js
```diff
@@ -297,13 +297,23 @@ exports.postTransactions = {
    * @return {Hapi.Response}
    */
   async handler (request, h) {
+    const blockchain = container.resolvePlugin('blockchain')
+
+    if (!blockchain) {
+      return h.response({ success: false, error: 'Blockchain not ready' }).code(500)
```
faustbrian: This is already handled by a plugin. https://github.com/ArkEcosystem/core/blob/develop/packages/core-p2p/lib/server/plugins/blockchain-ready.js

packages/core-p2p/lib/server/versions/1/handlers.js
```diff
@@ -297,13 +297,23 @@ exports.postTransactions = {
...
+
     if (!request.payload || !request.payload.transactions || !transactionPool) {
       return {
         success: false,
         transactionIds: []
       }
     }
 
+    if (request.payload.transactions.length > transactionPool.options.maxTransactionsPerRequest) {
+      return h.response({ success: false, error: 'Number of transactions is exceeding max payload size per single request.' }).code(500)
```
faustbrian: This should be broken down into multiple lines to increase readability.

```js
return h.response({
    success: false,
    error: 'Number of transactions is exceeding max payload size per single request.'
}).code(500)
```

packages/core/lib/config/testnet/plugins.js
```diff
@@ -25,7 +25,7 @@ module.exports = {
   },
   '@arkecosystem/core-database-sequelize': {
     dialect: 'sqlite',
-    storage: process.env.ARK_DB_STORAGE || `${process.env.ARK_PATH_DATA}/database/${process.env.ARK_NETWORK_NAME}_.sqlite`,
+    storage: process.env.ARK_DB_STORAGE || `${process.env.ARK_PATH_DATA}/database/${process.env.ARK_NETWORK_NAME}.sqlite`,
```
faustbrian: This should be reverted as it isn't a config change directly related to this PR.

PR #915, opened 2018-08-23
fix: transaction validation when validating blocks

alexbarnsley:
## Proposed changes
Currently blocks were only validating the actual block and payload. There was no validation on the actual transactions being sent. This meant it was possible to modify a forging node to accept invalid transactions, which ended up in a forged block. As a result, this could be broadcast to peers and they would also accept the block.

## Types of changes
The changes here mean that not only does it check if a transaction is valid in terms of signature, it checks the transaction can be applied to the sender, and that the schema of the transaction is valid.

I've also refactored one area (receiving new blocks from other peers) where it was validating a block 3 times in one go which was unnecessary.

**Edit: The changes i've made don't really feel "clean", in that passing the wallet manager to the Block model isn't ideal. I think this whole process needs a refactor at some point.**

## Checklist


kristjank:
@alexbarnsley I would simplify this as wallet manager does this in the blockchain.js, after the basic validation of block is done... so as you suggested: walletManager in the block model is not the best...

I would do structural testing of types and values (adding basic validator call `transaction.validate` to block.verify function)... 
when this is ok, then the next step is already done as with walletManager.applyBlock().. here: https://github.com/ArkEcosystem/core/blob/cc5c846976c3132d8c7652d965020a8f7c7c3c46/packages/core-blockchain/lib/blockchain.js#L309-L321

and here
https://github.com/ArkEcosystem/core/blob/cc5c846976c3132d8c7652d965020a8f7c7c3c46/packages/core-database/lib/interface.js#L385

alexbarnsley:
@kristjank ah I didn't see that, i thought I traced through it all. I was sure my tests I was able to send a tx for someone else. Just tried more extensively and failed so will remove the wallet manager stuff 👌

packages/crypto/lib/models/block.js
```diff
@@ -321,6 +326,12 @@ module.exports = class Block {
             result.errors.push('Encountered duplicate transaction: ' + transaction.data.id)
           }
 
+          if (!transaction.verified) {
+            result.errors.push('Invalid transaction in block: ' + transaction.data.id)
+          } else if (transactionValidator.validate(transaction.data).fails) {
```
kristjank: Id keep just his part... the rest is in blockchain.js

alexbarnsley: 👌 i wanted to check for verified first to reduce any unnecessary processing with the validator, but just had the thought while testing that 0 fee could mean the transaction is "verified", but the schema could still be invalid. will remove that check and test

alexbarnsley: That being said, if we do it later on it may have to do the whole revert transactions process. The benefit of doing it here is it doesn't start applying transactions, only to find that 1 of them is invalid. I'm going to split it into a separate `if` statement. It's only checking a boolean so speed impact will be next to nothing

faustbrian: What part to you want to remove @kristjank? There is no point to run any `if` conditions outside of the first one if the transactions is already not verified, thus saving time.

kristjank: yes I agree, I would just loop the block transactions and verify them with validator... and return error or not...

everything else related to:
- wallet manager move it out....

kristjank: aa was looking at old branch...

kristjank: the only concern I have is performance, as:

- before `new Block` changes and adding the blockObject... the `new Block` was called before entering and doing the final apply check (after basic chained checks and other stuff). Now we call it on the queue level, which could be more expensive, as verify is called thru constructor. 

and we should all test this from 0 sync

fix:
Is there a way in github to get all requests for reviews. I am always missing those

boldninja:
https://help.github.com/articles/choosing-the-types-of-notifications-you-receive/

packages/crypto/lib/models/block.js
alexbarnsley: you say expensive, but before those changes it was creating a new block object 3 times when one was received from another peer. Surely that's more expensive when it has to do validation each time? 🤔 

we'll have to wait on syncing from 0, because devnet v2 will be reset again

alexbarnsley:
@fix it is needed. Brian managed to modify his own node to forge a 0 fee tx, which then got distributed around the network. It was then accepted 🤷‍♂️ 

How do you propose fixing issues if we can't add fixes due to performance decreases? I will close if you want me to, since performance is so critical.

fix:
@boldninja i only want to get notifications for review, not everything

fix:
found: 
screen shot 2018-08-24 at 12 27 28

faustbrian:
Without those extra checks I was able to forge a block that included a 0 fee transaction and killed the whole devnet after which we had to reset to 0 because all nodes were dead and stuck at that block.

alexbarnsley:
Also i don't see how relying on applyTransaction is a benefit? If there's an invalid transaction found which could have been caught sooner, the node has to start reverting transactions and blocks. Why not just check before even starting to apply them?

kristjank:
@fix this is more of a structural check... in case a bad delegate, changes and comments out the limits and other checks... he can forge a block with 0 fee transactions.... as on the block forge we check the sums all is ok...

so other nodes accept this block - meaning it is possible to forge transaction with 0 fees....

So to make the same independent check on the receiving nodes we suggested to add validator, to check (already deserialised) transaction if all is within limits.. and if so... accept block... or not

I know the same bad delegate can change  the code in the validator, but still we have one more check there.. until the bad delegate is voted out...

fix:
the only rules that is missing in the applyTransaction is `tx.fees > 0`

fix:
you can't apply a 'local' rules about market fees on block validation otherwise the network will fork.

kristjank:
@fix need to check something else, as this rule is there in the `.canApply` the transaction.validator is already run - so this should be ok then...

kristjank:
@fix you where right...

found the buggy then... in case of coldWallet... the `.canApply` is never called...  pr incoming, gonna change that part of wallet-manager a little bit and add some other  vote/ isdelegate checks for votes not coming thru...

@fix @alexbarnsley @faustbrian

alexbarnsley:
Not sure where market fees come into this? This is transaction fees? Whatever, looks like you have this in hand @kristjank so i'm closing this.

fix:
market fees: Rules that a delegate decide on what is acceptable fees or tx to be included in his own block

as all delegates may have different strategies, the only rule or anybody to validate a block is to have all tx fees included in blocks strictly > 0

PR #928, opened 2018-08-24
Don't load .env in test environment

supaiku0:
## Proposed changes
The `~/.ark/.env` doesn't exist on Travis, so it makes sense to not even attempt to load it 
when `ARK_ENV=test` is true. 

This fixes an issue I ran into when running tests (core-forger) locally:
screenshot_20180824_175318

Since the container loaded my `.env` and thus set `ARK_P2P_PORT` to `4002`, which caused `process.env.ARK_P2P_PORT || 4000` from `core/lib/config/testnet/plugins.js` to return the wrong port.

## Checklist

### DIFF
--- a/packages/core-container/lib/environment.js
+++ b/packages/core-container/lib/environment.js
@@ -69,6 +69,12 @@ module.exports = class Environment {
    * @return {void}
    */
   __exportVariables () {
+    // Don't pollute the test environment, which is more in line with how
+    // travis runs the tests.
+    if (process.env.NODE_ENV === 'test') {
+      return
+    }
+
     const envPath = expandHomeDir(`${process.env.ARK_PATH_DATA}/.env`)
 
     if (fs.existsSync(envPath)) {


faustbrian:
If this is a travis specific issue you should use `if(process.env.ARK_CI_TEST)` because the current check will also be active for testnet.

faustbrian:
If we no longer load the .env file for testnet the local testing can become quite painful. I have an .env file that I use to store all my testnet configuration that I copy between local and server so I can always have the exact same config.

I will take a closer look when I am back tomorrow.

packages/core-container/lib/environment.js
```diff
@@ -69,6 +69,12 @@ module.exports = class Environment {
...
   __exportVariables () {
+    // Don't pollute the test environment, which is more in line with how
+    // travis runs the tests.
+    if (process.env.ARK_ENV === 'test') {
+      return
+    }
+
     const envPath = expandHomeDir(`${process.env.ARK_PATH_DATA}/.env`)
 
     if (fs.existsSync(envPath)) {
```
faustbrian: Shouldn't this already prevent the loading of any file if it doesn't exist?

PR #931, opened 2018-08-27
Use container to retrieve 'p2p' ref in api handlers

paroxysm:
## Proposed changes
Some minor changes to the p2p handlers to retrieve the p2p reference from the container. This makes it easier to write tests for them since we're simplifying the dependency.
(i.e. We'd only need to mock the container's resolvePlugin call as opposed to mimicking a request.server.app object)

I realize @faustbrian  will overhaul the implementation of p2p API at some point, however, there's some salvageable code in there so this won't entirely be deprecated

## Types of changes

## Checklist

<!--

### DIFF
--- a/packages/core-p2p/lib/monitor.js
+++ b/packages/core-p2p/lib/monitor.js
@@ -313,7 +313,7 @@ class Monitor {
 
   /**
    * Check if we have any peers.
-   * @return {bool}
+   * @return {boolean}
    */
   hasPeers () {
     return !!this.getPeers().length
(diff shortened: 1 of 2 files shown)


faustbrian:
Will have a look at this later this week. I remember there being some issues with resolving `core-p2p` inside `core-p2p` itself, which is why it initially was bound to the `request.server.app` object of `hapi.js`.

faustbrian:
I reviewed it and will close this as the `container.resolve` will be unnecessary after my next refactor as I will require the monitor from inside the plugin itself. Thanks for the PR though!

paroxysm:
no prob.

Issue #925, opened 2018-08-24
Wallet manager updates


kristjank:
**Describe the bug**
1. If a wallet is cold - the canAppy check was not performed in walletManager
Related to #915 

2. Also check for vote transaction is not working - you can send votes to non-existing delegates (to normal wallets)... which makes delegate list a bit bigger...

3. (optional) Try updating the delegate query to just take vote transaction into account if the voted wallet is a registered delegate

PR #933, opened 2018-08-27
Fix: validation of incoming payloads

kristjank:
## Proposed changes
Changed the validation of applyTransactions to take into account recent bug #925.
- Not allowing to to double votes (voting again when voted)
- Not allowing to vote for non-existing delegate (although) wallet exists
- In case of cold wallet, still check transaction conditions for vote...

## Types of changes

## Checklist


kristjank:
@faustbrian

packages/core-transaction-pool/lib/pool-wallet-manager.js
```diff
@@ -94,17 +94,18 @@ module.exports = class PoolWalletManager extends WalletManager {
     if (!recipient && recipientId) { // cold wallet
       recipient = new Wallet(recipientId)
       this.walletsByAddress[recipientId] = recipient
+    }
```
faustbrian: Did you make sure that breaking apart this `if else` block won't have any side effects?

kristjank: sending data to cold wallet - so first if is true... so if the if-else remain together 

`  } else if (!sender.canApply(data)) {` is never executed and transaction is accepted with (0 fees for example or anything else.

-- 
we should then if - transaction is not applied/accepted - remove the cold wallet from wallet manager.. @fix

fix: mmm true that. Can you document this  in the code or maybe extract these lines as a function:
https://github.com/ArkEcosystem/core/blob/fc4a29e866f03cd26ebc0a0b8ca1e0206becc4ee/packages/core-database/lib/wallet-manager.js#L212-L218

replaced by `const recipient = this.findOrCreateWallet(recipientId)`

faustbrian: @fix https://github.com/ArkEcosystem/core/pull/933/commits/3126c3713d0025873f0fef45ca7c26ddb866cbdc

Should be ready to be merged.

PR #942, opened 2018-08-29
refactor: use immutable Map instead of object as storage

faustbrian:
## Proposed changes

_This PR refactors the `core-storage` package to use an immutable map to handle the storage instead of a plain object._

Maps have the same performance, sometimes are even faster, and have built-in methods to modify their state that make dependencies like `lodash` unnecessary. Also added some helper methods to create immutable objects without straight-away adding them to the storage.

## Types of changes

## Checklist

### DIFF
--- a/packages/core-storage/lib/storage.js
+++ b/packages/core-storage/lib/storage.js
@@ -11,10 +11,18 @@ module.exports = class Storage {
   }
 
   /**
-   * Get all of the storage entries.
+   * Get the storage map.
    * @return {Object}
    */
   all () {
+    return this.storage
+  }
+
+  /**
+   * Get all of the storage entries.
+   * @return {Object}
+   */
+  entries () {
     return this.storage.entries()
   }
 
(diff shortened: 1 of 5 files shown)


packages/core-storage/lib/storage.js
```diff
@@ -1,24 +1,47 @@
...
-   * Get all of the storage data.
-   * @return {[type]}
+   * Get the storage map.
+   * @return {Object}
    */
   all () {
     return this.storage
   }
 
+  /**
```
alexbarnsley: Perhaps these might be better as get methods?:

faustbrian: It is a method to mimic the behaviour of the underlying `immutable.js` library as I want to keep the interface the same because the storage is only a wrapper.

alexbarnsley: Ah okay, if it's to match that then it makes sense 👌

PR #790, opened 2018-07-31
[WIP] Tx exceptions

alexbarnsley:
Will update this later

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Refactoring (improve a current implementation without adding a new feature or fixing a bug)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Build (changes that affect the build system)
- [ ] Docs (documentation only changes)
- [ ] Test (adding missing tests or fixing existing tests)
- [ ] Other... Please describe:

## Checklist

- [ ] I have read the [CONTRIBUTING](https://docs.ark.io/docs/contributing) documentation
- [ ] Lint and unit tests pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)

Issue #754, opened 2018-07-12
Invalid Second Signature Registration ID computed (Type 1)


faustbrian:
**Describe the bug**
Currently the IDs of second signature registration transactions are invalid after deserializing because data is added that does not exist in the transaction itself which results in an altered byte representation that produces a different hex string.

Type 1 transactions don't have a `recipientId` themselves and if they do it is null and results in the field being ignored by `getBytes` in ark-js.

The problem is that we add the `recipientId` when performing the deserialization which results in a wrong ID due to more bytes being added.

Commenting out this line https://github.com/ArkEcosystem/core/blob/master/packages/crypto/lib/models/transaction.js#L354 results in a valid ID being computed.

This is the same issue as with https://github.com/ArkEcosystem/core/issues/715 where we add data that does not exist in v1 and results in wrong IDs.

**To Reproduce**
```
const fixture = {
    data: {
        type: 1,
        amount: 0,
        fee: 500000000,
        recipientId: null,
        senderPublicKey: "034151a3ec46b5670a682b0a63394f863587d1bc97483b1b6c70eb58e7f0aed192",
        timestamp: 41271109,
        asset: {
            signature: {
                publicKey: "03699e966b2525f9088a6941d8d94f7869964a000efe65783d78ac82e1199fe609"
            }
        },
        signature: "304402202ae5adf4b33af9d3d98b1becf06590fd6b5a879fb36dc79f41c9406b09f13273022007722ff3c1672b7933fb29d12f70054430da4493d333c9ca957374ac2c258f69",
        id: "ccf91b59afece43f44d49dfe4959c85301ed951a87ef5b8bae1e0a52894d89b7"
    },
    serialized: "ff011e0145bf7502034151a3ec46b5670a682b0a63394f863587d1bc97483b1b6c70eb58e7f0aed1920065cd1d000000000003699e966b2525f9088a6941d8d94f7869964a000efe65783d78ac82e1199fe609304402202ae5adf4b33af9d3d98b1becf06590fd6b5a879fb36dc79f41c9406b09f13273022007722ff3c1672b7933fb29d12f70054430da4493d333c9ca957374ac2c258f69"
}

const serialized = Transaction.serialize(fixture.data)
console.log(serialized.toString('hex') === fixture.serialized) // true

const deserialized = Transaction.deserialize(serialized.toString('hex'))
console.log(fixture.data.id === deserialized.id) // false
```

**Expected behavior**
[...]

Issue #715, opened 2018-06-28
Invalid multi signature registration (type 4) IDs computed


faustbrian:
**Describe the bug**
Below transaction has an ID of `cbd6862966bb1b03ba742397b7e5a88d6eefb393a362ead0d605723b840db2af` but the computed ID after deserialising it is `a8eefb5c216845d2eccda5e489076bc70d2eaef8e8f127726c249c16187a25a4`.

**This issue only exists with multi signature registrations! All other transaction types have their correct IDs computed.**

**To Reproduce**
```js
const transaction = {
  version: 1,
  network: 23,
  id: "cbd6862966bb1b03ba742397b7e5a88d6eefb393a362ead0d605723b840db2af",
  blockid: "1844069042066945391",
  type: 4,
  timestamp: 10112114,
  amount: 0,
  fee: 2000000000,
  senderId: "AMw3TiLrmVmwmFVwRzn96kkUsUpFTqsAEX",
  senderPublicKey: "036928c98ee53a1f52ed01dd87db10ffe1980eb47cd7c0a7d688321f47b5d7d760",
  signature: "30440220324d89c5792e4a54ae70b4f1e27e2f87a8b7169cc6f2f7b2c83dba894960f987022053b8d0ae23ff9d1769364db7b6fd03216d93753c82a711c3558045e787bc01a5",
  signSignature: "304402201fcd54a9ac9c0269b8cec213566ddf43207798e2cf9ca1ce3c5d315d66321c6902201aa94c4ed3e5e479a12220aa886b259e488eb89b697c711f91e8c03b9620e0b1",
  signatures: [
    "304502210097f17c8eecf36f86a967cc52a83fa661e4ffc70cc4ea08df58673669406d424c0220798f5710897b75dda42f6548f841afbe4ed1fa262097112cf5a1b3f7dade60e4",
    "304402201a4a4c718bfdc699bbb891b2e89be018027d2dcd10640b5ddf07802424dab78e02204ec7c7d505d2158c3b51fdd3843d16aecd2eaaa4c6c7a555ef123c5e59fd41fb",
    "304402207e660489bced5ce80c33d45c86781b63898775ab4a231bb48780f97b40073a63022026f0cefd0d83022d822522ab4366a82e3b89085c328817919939f2efeabd913d"
  ],
  asset: {
    multisignature: {
      min: 2,
      keysgroup: [
        "+03543c6cc3545be6bac09c82721973a052c690658283472e88f24d14739f75acc8",
[...]

PR #956, opened 2018-08-31
Fix for id calculation

fix:
This is the fix for ill calculation of id (80d75c7b90288246199e4a97ba726bad6639595ef92ad7c2bd14fd31563241ab)

Tested on mainnet full rebuild


fix:
@faustbrian i dont understand what you mean.

Essentially the issue is v1 mainnet had issue on tx type 1 and 4 where recipientId (which is useless) was an optional field. Some tx have it some have not. It also have issue on verification of the signature

The idea for the code is that if i get those tx and verification failed, we try to set or remove this field and try to verify again signature and recompute id.

fix:
ah ok gotcha what you mean.

fix:
Can you find devnet tx that had issues?

fix:
At some point we should put those examples in tests to be sure we dont have regressions on them

fix:
also thinking loudly about some refactoring: Instead of shenanigans to be able to fix v1 serialisations issue mixed with aip11 implementation, what about (2 options):
- having a function passing deserialisation AND original transaction and performing the the needed fix. pro: no issue about injection attack but difficulties to sort out how legit is the original data
- having a function when receiving a new tx (from blocks or single) and perform the analysis. pro: less redundancy, more prone to injection.

alexbarnsley:
@fix I briefly started but Brian wanted to hold off on it, here: https://github.com/ArkEcosystem/core/pull/790

You can find the exact txs here:

https://github.com/ArkEcosystem/core/issues/754
https://github.com/ArkEcosystem/core/issues/715

I think this is what Brian is referring to

fix:
test passing on models

faustbrian:
As @alexbarnsley already linked, in https://github.com/ArkEcosystem/core/issues/754 are 3 transactions from the old devnet that deserialised as type 1.

They were initially signed **with** a recipient ID, which shouldn't be present on type 1, so their IDs are only valid with a recipient ID **before** computing the transaction ID.

All other type 1 transactions on the old devnet were signed **without** a recipient ID so their IDs are only valid if the recipient ID is added **after** computing the transaction ID.

We would basically need to check for type 1 and 4 if they already have a recipient ID which means we need to compute the transaction ID **with** the recipient ID included, if not we need to add it after the transaction ID is computed.

fix:
added those tx on tests

faustbrian:
Those 3 transactions will be **valid with** the recipient address but all others failed on the old devnet.

I will ask tharude if he still has a v1 devnet snapshot that we could use to reproduce the issue with the IDs.

1. sign **type 1** transaction **without** recipient ID
2. **deserialise** it with @arkecosystem/crypto __(it will add the recipient ID)__
3. transaction ID will be different __(because step 2 adds the recipient ID which results in a different byte representation)__

1. sign **type 1** transaction **without** recipient ID
2. modify the **deseralize** method to add the recipient ID **AFTER** computing the ID
3. deserialise it with **@arkecosystem/crypto** __(it will add the recipient ID)__
4. transaction ID will be the same as before deserialising

faustbrian:
Adrian will restore a v1 snapshot that we can use to query the devnet transactions.

fix:
this branch fix this issue. The way i handle it is the following:
- serialize the tx
- deserialize the tx
- if v1 apply compatibility fix
- if verification failed and type is 1 or 4, remove recipientId and recompute

fix:
the last 2 steps are handled here:
https://github.com/ArkEcosystem/core/pull/956/files#diff-20020cdc4aa37467285162d1bf6da2e0R80

fix:
maybe one clearer solution would be to handle different serialization if v1 or v2, at the price of duplicated code

fix:
ie 
```
if tx.version = 1 -> serializeV1(tx)
else serializeAIP11(tx)
```

fix:
well that is likely better actually, because the way this is coded we could build the new AIP11 x (resignation etc...) the v1 way...

fix:
ok i have finally finished the work on this. Just needs now to fix a few tests and we are good to merge

packages/core/lib/config/mainnet/plugins.js
```diff
@@ -43,8 +43,8 @@ module.exports = {
       host: process.env.ARK_DB_HOST || 'localhost',
       port: process.env.ARK_DB_PORT || 5432,
       database: process.env.ARK_DB_DATABASE || `ark_${process.env.ARK_NETWORK_NAME}`,
-      user: process.env.ARK_DB_USERNAME || 'ark',
-      password: process.env.ARK_DB_PASSWORD || 'password'
+      user: process.env.ARK_DB_USERNAME || 'fixcrypt',
```
faustbrian: Those changes shouldn't be committed as they are your personal configuration.

faustbrian:
API tests are failing on this branch because https://github.com/ArkEcosystem/core/blob/develop/packages/core-api/lib/versions/2/transformers/transaction.js#L18 results in a different object then before which is missing data like an ID and Recipient.

packages/core-database/lib/wallet-manager.js
```diff
@@ -125,7 +110,7 @@ module.exports = class WalletManager {
    * @param {void}
    */
   setByPublicKey (publicKey, wallet) {
-    this.byPublicKey = this.byPublicKey.set(publicKey, wallet)
+    this.byPublicKey = this.byPublicKey[publicKey] = wallet
```
supaiku0: Same.

packages/core-database/lib/wallet-manager.js
```diff
@@ -135,7 +120,7 @@ module.exports = class WalletManager {
    * @param {void}
    */
   setByUsername (username, wallet) {
-    this.byUsername = this.byUsername.set(username, wallet)
+    this.byUsername = this.byUsername[username] = wallet
```
supaiku0: Same.

packages/core/lib/config/mainnet/plugins.js
fix: yes somehow that file slipped, pushing a revert

fix:
yes on that aspect there is a breaking change.
- I removed all the 'fixing parts' of serialization/deserialization function, because it is mixing concern.
So now these functions are only performing clean AIP11 compliant ser/der
- the only issue is id because id calculation IS DIFFERENT than v1. so the only way to get this 'metadata' is to use the constructor.
- this is where i have no good solution, to find the right id on v1, i need to calculate signature and if different, i need to fix the deserialization (case type 1 or 3) by adding/removing recipientId
- All this to say that using the constructor is far slower in all cases.

fix:
also another source of fail is rank.
The calculation is done ONLY at change round. Otherwise it is NOT correct in the consensus meaning.
- I have removed the rank calculation from wallet-manager for that very reason.
- I we want to enable it, we need to be SURE that this is NEVER called at turn round. However there is no such data in wallet-manager.
- The only source of reliable ranking is from database.getDelegates()

faustbrian:
If things get removed and behaviour breaks we need to fix this behaviour so all tests go green or we can't approve and merge a PR.

A lot of things were moved or removed so they need to be substituted to provide the same behaviour as before those changes for things like `core-api`.

fix:
That is not the point.
The test is wrong, this is my point, the rank is meaningless because ill calculated, and the test does not catch the fact this is ill calculated.

faustbrian:
99% of the tests that are failing is because transactions have no IDs, not because of a delegate rank.

fix:
yes i am fixing them.

fix:
no please merge i will do it, i need a bit more time

fix:
getting remaining tests fails i have no idea where to start

    ● API 2.0 - Blocks › POST /blocks/search › should POST a search for blocks with the exact specified reward

    expect(received).toHaveLength(length)

    Expected value to have length:
      1
    Received:
      []
    received.length:
      0

      314 |       utils.expectCollection(response)
      315 |
    > 316 |       expect(response.data.data).toHaveLength(1)
          |                                  ^
      317 |
      318 |       const block = response.data.data[0]
      319 |       utils.expectBlock(block)

      at Object.it (packages/core-api/__tests__/v2/handlers/blocks.test.js:316:34)

      ● API 2.0 - Blocks › POST /blocks/search › should POST a search for blocks with the specified reward range

    expect(received).toHaveLength(length)

    Expected value to have length:
      1
    Received:
      []
    received.length:
      0

      333 |       utils.expectCollection(response)
      334 |
    > 335 |       expect(response.data.data).toHaveLength(1)
          |                                  ^
      336 |
      337 |       const block = response.data.data[0]
      338 |       utils.expectBlock(block)

      at Object.it (packages/core-api/__tests__/v2/handlers/blocks.test.js:335:34)

supaiku0:
@fix Travis is green and locally I also can't reproduce it. Are you sure you don't have any local changes which may cause the failure?

fix:
Not completely green on Travis but I have more error locally.

fix:
mmm travis is ok now, but show failed here

fix:
restarted tests, seems its getting through

Issue #945, opened 2018-08-29
Dynamic fees


zillionn:
**Is your feature request related to a problem? Please describe.**
Currently, if I send a transaction with a very low fee, lower than the lowest set by a delegate, there is no way to know that my fee is too low and nobody will forge the tx.

**Describe the solution you'd like**
How about to store the delegate's minimum fee `generator_min_fee` in the `blocks` table and when a node receives a tx, it could check the min fee in the `blocks` table for the last 51 block, and if the tx fee is lower than the lowest, the node could just ignore the tx and return some error response.

**Describe alternatives you've considered**
...

**Additional context**
...

kristjank:
I think that better for this would be to still accept it in the pool. And if it doesnt get forged in one round, then you know it will not get forged and will eventually expire...  but you could still lookit up via unconfirmed transaction endpoint

alexbarnsley:
@kristjank 1 round is not accurate. If all blocks are full, there could be 1 delegate who's capable of forging said transaction, but their block was already full with others. Doesn't mean it will never get forged.

Issue #927, opened 2018-08-24
Dynamic Fee Transaction broadcasted and never forged


faustbrian:
**Describe the bug**
When configuring your node to accept a fee lower then the network minimum accepted this check https://github.com/ArkEcosystem/core/blob/develop/packages/core-transaction-pool/lib/utils/dynamicfee-matcher.js#L29-L32 will reject your transaction and broadcast it even though your node should accept it.

**To Reproduce**
Configure your node to accept fees like 150 while the network minimum accepted is 200. The transaction will be rejected by your own node and be broadcasted.

**Expected behavior**
I would expect that my node accepts the transactions even if it is below the accepted minimum of the network as my node is willing to accept the lower fee.

**Additional context**
I reproduced this on devnet and on testnet with several different configurations. Once there is an accepted minimum on the network it seems to be impossible to go below that without modifying my node. After disabling above mentioned check my node will forge the transaction for fees my node is willing accept, even if they are below what the network accepts.

kristjank:
It is not network minimum but is delegate minimum setting. There is no `network` minimum setting, just delegate minimum and clients fee - willing to pay for the transaction to be processed.

Also the logger change is not correct. So this is misleading information. 

The fee set by the client `transaction.fee` was not accept by your delegate, but was broadcasted to other nodes. 

- yes it can still happen that it will be never forged, if no other delegate accepts it, as it is above their limits... so not sure how to respond to the user as you don't have information about other delegates and their settings, nor their pools...

faustbrian:
I tried the exact same configuration and fees both on devnet and on a local testnet, the local testnet accepts them while devnet rejects them and only broadcasts them. The testnet had 3 nodes so having a single node can't be the issue.

kristjank:
let me enable dynamic fees on my forgers and send again. 

we don't know the rules of all the delegates and what are their settings...  so if you where the only delegate with dynamic fees and you fee was too low, the transaction was not accepted...

the only update I can think of is accept the dynamic fee to pool and then display it somehow... but it is still related to the nodes and peers accepting this...

faustbrian:
Just modified my node so that is solved :trollface:

kristjank:
magic modifications :) hard to keep up with all test scripts, nodes and project.

PR #966, opened 2018-09-04
Feat/transaction pool dynamic fees

kristjank:
## Proposed changes
As it was hard to identify what is happening with dynamic-fee transactions when not accepted by local node, but forwarded to other nodes, the logic has changed a bit, to:

1.  If transactions is not accepted by the delegate's fee criteria, it still goes into the pool, where it can be watched via https://docs.ark.io/api/public/v2/transactions.html#list-all-unconfirmed-transaction endpoint.

2. if it will be forged by other delegates, you will no longer find it there, otherwise it will wait in the pool for 6 hours and then it will expire.

Also cleaned up the transaction-pool-interface and moved all redis stuff down to redis. 

Related to #945 and #927 

## Types of changes

## Checklist
<!--

## Further comments
It is not very aesthetic to do that with redis as some core  table structure is needed and some more advanced queries would simplify things a lot being done with sql like structure. 

Will recheck again for a new refactor to make it less complex.


faustbrian:
Shouldn't the https://github.com/ArkEcosystem/core/pull/966/files#diff-e4d88fe8b570e758f4cdbb25b8724413 file stay in `core-transaction-pool` so custom implementations can re-use it?

packages/core-transaction-pool-redis/lib/connection.js
```diff
@@ -237,6 +245,51 @@ module.exports = class TransactionPool extends TransactionPoolInterface {
...
 
+  /**
+   * Removes any transactions in the pool that have already been forged.
+   * @param  {Array} transactionIds
+   * @return {Array} IDs of pending transactions that have yet to be forged.
+   */
+  async removeForgedAndGetPending (transactionIds) {
+    const forgedIdsSet = new Set(await database.getForgedTransactionsIds(transactionIds))
+
+    await Promise.each(forgedIdsSet, async (transactionId) => {
```
faustbrian: I would just use a simple `for (const transactionId of forgedIdsSet)` instead of adding another dependency in the form of `bluebird`.

kristjank: Yes we add another dep, but gain performance. Would go for this here as it can be time-consuming, e.g. wait for 150 trx in line or wait for 150  trx in parallel to be processed. And its called on couple of places.

roks0n: If you wish to run them in parallel you should use `Promise.all()` which is supported natively. `Promise.each()` is the same as a for loop.

Issue #967, opened 2018-09-04
Implement estimate queries for improved performance


faustbrian:
**Problem**

The problem right now is that we have no proper pagination on the blocks & transaction endpoints because the count queries of PostgreSQL have a massive performance impact on the average server we have on mainnet.

The performance hit is caused by how count queries work, which is that they need to verify that a record exists. That means if you hit the blocks API and use a count query it has to look up almost 6 million records just to count them, this is where estimates and explaining come into play.

Estimates are information that PostgreSQL will always update by itself in the background in a meta table that contains information about our database structure and data. [EXPLAIN](https://www.postgresql.org/docs/9.4/static/using-explain.html) is a query keyword provided by PostgreSQL which allows you to analyse queries, this can also be leveraged to perform multitudes faster queries to get information like a count estimate rather then using the `COUNT` keyword.

My average `COUNT` query time on a low-tier or mid-tier server was around 60-100ms for only 115.000 records. If we scale this up to mainnet it would take us several seconds to load the blocks endpoint which is not feasible.

My average `EXPLAIN` query time on the same servers was around `650µs` which would require me to run the `EXPLAIN` query over 100 times to consume the same time of a single `COUNT` query.

_Using estimates & explain will allow us to serve larger data sets in shorter time while still providing meaningful meta data like counts and a proper pagination._

**Solution**

```sql
# Function
CREATE FUNCTION count_estimate(query text) RETURNS integer AS $$
DECLARE
  rec   record;
  rows  integer;
BEGIN
  FOR rec IN EXECUTE 'EXPLAIN ' || query LOOP
    rows := substring(rec."QUERY PLAN" FROM ' rows=([[:digit:]]+)');
    EXIT WHEN rows IS NOT NULL;
  END LOOP;
  RETURN rows;
END;
$$ LANGUAGE plpgsql VOLATILE STRICT;

# Query
SELECT count_estimate ('SELECT * FROM transactions WHERE type = 1');
```

fix:
What about estimation in the stat table

https://stackoverflow.com/a/20574069

Also for verification of the database at start-up the counts should be exact

PR #969, opened 2018-09-05
feat: pure PostgreSQL integration

faustbrian:
## Proposed changes

This PR integrates https://github.com/vitaly-t/pg-promise which provides superior performance and resolves https://github.com/ArkEcosystem/core/issues/967.

This paves the way for future optimisations on the database layer without having to do any comprises because we have to keep in mind how mysql, sqlite or god knows what work as our main focus should be one engine and that is postgres.

The performance impact on the API has already been quite nice after some initial testing as count queries are now possible even on low-end servers which allows us to provide end-users with better pagination of data and serve that data faster.

**This is a work in progress as some adjustments are still left to be done but syncing with devnet just finished so that already works.**

**Todo**
- [x] Create migrations
- [x] Create sql query files
- [x] Create models
- [x] Create repositories
- [x] Create helpers to interact with sql files
- [x] Adjust query builder to pg-promise placeholder syntax
- [x] Adjust query builder to make use of pg-promise helpers for formatting
- [x] Decouple core-api from any database specifics
- [ ] ~~Decouple core-graphql from any database specifics~~ **core-graphql currently has no test suite so that will be done once @air1one implemented a test suite**

## Types of changes

## Checklist


faustbrian:
@alexbarnsley @supaiku0 This is soon ready for testing. I just have to resolve an issue where `estimate` queries fail if they contain any `WHERE` clauses here https://github.com/ArkEcosystem/core/blob/feat/pg-promise/packages/core-api/lib/repositories/repository.js#L22.

faustbrian:
@supaiku0 @alexbarnsley ready for testing

vitaly-t:
Perhaps the best-done example of integrating `pg-promise` into a serious product: [Lisk](https://github.com/LiskHQ/lisk).

fix:
@vitaly-t are you trolling here?

vitaly-t:
@fix What's that supposed to mean? 😕 

This PR refactors the code to use `pg-promise`. I gave the link to a project that illustrates the best integration approach for it, as a perfect example.

fix:
I know very well lisk and how it was integrated at that time.

I was meaning, do you imply ARK is not serious?

vitaly-t:
> I was meaning, do you imply ARK is not serious?

I didn't even look at this project, that's beside the point. I only look at popular projects that try to integrate with `pg-promise` and give valuable advises.

You are looking at it the wrong way.

vitaly-t:
Oh, good to know. I'm still assisting them with the database layer when I have time. May I ask why you left?

fix:
We had complete divergent view on what blockchain should be, and how this should be implemented

supaiku0:
There's at least one place where we occasionally pass an empty array to a query with an `WHERE ... IN` clause, producing faulty sql: `SELECT * FROM table WHERE id IN ()` resulting in an exception.

This happens in`core-transaction-pool-redis` at `removeForgedAndGetPending`. It was fine before, because joining an empty array returns an empty string and `SELECT * FROM table WHERE id IN ('')` is valid.

I think we should add additional checks to each repository right before we pass an array to any query with an `WHERE ... IN` clause and return early with an empty result just to be safe. We don't have many of them anyway.

Regarding votebalance:
Since `votebalance` is lowercase and not camelCase, the function `camelize-columns` doesn't apply here. I opened a PR #981 to rename it, so we don't have to add any workarounds.

vitaly-t:
@supaiku0 See [WHERE col IN Query with empty array as parameter](https://stackoverflow.com/questions/49188085/where-col-in-query-with-empty-array-as-parameter).

PR #970, opened 2018-09-05
feat: client peers override

alexbarnsley:
## Proposed changes
This gives the client package the ability to work with other networks. Passing in the network name previously meant that only mainnet and devnet were possible.

## Types of changes

## Checklist


j-a-m-l:
An alternative approach would be changing the `connect` and `findPeers` methods to accept an `Object` instead of relying on the argument order:

`findPeers({ network: 'mainnet", version: 2 })`

Instead of the `peersOverride`, it could be named `hosts` or peers`:

`findPeers({ hosts, version: 2 })`

PR #985, opened 2018-09-11
chore: remove sequelize

faustbrian:
## Proposed changes

Removes **core-database-sequelize** which can now be found here https://github.com/ArkEcosystem/core-plugins so we can focus on postgresql to optimise database performance and get rid of things like mysql and sqlite.

## Types of changes

## Checklist


fix:
i disagree with this, please resubmit a PR to add it back

faustbrian:
https://git-scm.com/docs/git-revert

fix:
Yes i am sure you know how to do it.
Either that or either have a `core-database-sqlite` plugin before deprecating. This is highly important, we had already this discussion but we do need no install for database option out of the box

fix:
ok alternatively, is it possible to have:
- a config able to use the new core-database-sequelize
- ditch the 'non supported' part on this new repository
@faustbrian

Issue #991, opened 2018-09-12
Improve crypto by leveraging secp256k1-node


supaiku0:
Currently we have ECPair, ECPoint and ECSignature, re-implementing a lot of ECC in JavaScript. But [secp256k1-node](https://github.com/cryptocoinjs/secp256k1-node) can do all of that as well, natively, removing the need to have our own slow ECC implementation.

In essence, by removing our own ECC from `crypto.verify` and using `secp256k1` directly...

this:
https://github.com/ArkEcosystem/core/blob/66ba069a93e93c74a706520f6adf734134b9656f/packages/crypto/lib/crypto/crypto.js#L253-L259

becomes:
```javascript
const secp256k1 = require('secp256k1')
...
const hash = this.getHash(transaction, true, true)
const signatureBuffer = Buffer.from(transaction.signature, 'hex')
const senderPublicKeyBuffer = Buffer.from(transaction.senderPublicKey, 'hex')
const parsedSignature = secp256k1.signatureImport(signatureBuffer)
return secp256k1.verify(hash, parsedSignature, senderPublicKeyBuffer)
```
It is approximately ***more than twice*** as fast as before. I calculated the average of each verify call when syncing devnet by wrapping the verify call in `console.time("verify") ... console.timeEnd("verify")`

Average of verify:
before: 1.56ms
after: 0.67ms

Since most blocks are empty right now, the effect is negligible, but it's still absolutely advised to let go of our ECC in favor of secp256k1-node where possible, since we also use it in more places than crypto.verify and the performance difference isn't small.

@fix 
I'm preparing a branch where we exclusively use secp256k1, would like your feedback @fix

zillionn:
> This implementation is super experimental, use it at your own risk.

fix:
well, i have made some tests yesterday and today actually.
Verifying one block is 1-2 ms (there is a signature too), so i think it will be noticeably faster
on the details: 
- great catch! i did not expect improvements here because that part of code was taken from bitcoin-js.
- there are a few things to verify, above all on the data passed to the secp256k1 lib, if it can be tampered with (that is the meaning of 'superexperimental', secp256k1 is used in production for bitcoin)

vRobM:
https://github.com/gonetwork-project/RNFastSecp256k1

PR #992, opened 2018-09-12
perf: use secp256k1 for verifying and signing

supaiku0:
## Proposed changes

As mentioned in #991 , by using secp256k1 natively the time spent verifying and signing is cut down by half or more on average.

I have been working on a bigger refactor of `crypto`, but I don't feel comfortable pushing these two changes together. So let's first get this one in.

## Types of changes

## Checklist

### DIFF
--- a/packages/crypto/lib/crypto/crypto.js
+++ b/packages/crypto/lib/crypto/crypto.js
@@ -1,13 +1,11 @@
-// const createHash = require('create-hash')
 const bs58check = require('bs58check')
 const crypto = require('crypto')
-// const { Buffer } = require('buffer/')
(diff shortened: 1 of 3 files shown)


fix:
ok it was already there https://github.com/ArkEcosystem/core/blob/develop/packages/crypto/package.json#L48

supaiku0:
> ok it was already there https://github.com/ArkEcosystem/core/blob/develop/packages/crypto/package.json#L48

Also see here, that's why I checked it in the first place:
https://github.com/ArkEcosystem/core/blob/develop/packages/crypto/lib/crypto/ecpair.js#L192

PR #994, opened 2018-09-13
chore: run tests first

faustbrian:
## Proposed changes

Tests are more important then linting or depchecks, there is no point in waiting 10 minutes for tests to run just because of those.

## Types of changes

- [x] Build (changes that affect the build system)

## Checklist


fix:
was thinking the same thing actually ahah

PR #997, opened 2018-09-14
refactor: crypto package

supaiku0:
## Proposed changes

This is a refactor and cleanup of `@arkecosystem/crypto`. Most of the classes carried over from [bitcoinjs](https://github.com/bitcoinjs/bitcoinjs-lib) (ECPair, ECPoint, ECSignature, ...) are not needed (#992) and only add complexity without value. That's why they are now being removed.

My first attempt was to use bitcoinjs directly, since there have been a lot of changes. But I ran into issues with `ECPair.sign`, where the same unit tests would randomly fail or succeed. :man_shrugging: Since it's essentially just calling the `secp256k1` lib and we only need it for signing and verifying  I stopped bothering and decided against it.

A few of our packages were using those classes mentioned above for signing hashes and WIF. Thus I introduced new functions to `crypto.js` and changed the other packages accordingly.

Changes include:
- Removed ECC classes
- Removed other obsolete classes and files
- Lesser dependencies
- New functions `keysFromWIF`, `keysToWIF`, `signHash` and `verifyHash`

An interesting side effect is that all unit tests now complete in under 2 minutes on my machine.

Please check if the changes are reasonable and I didn't miss anything @fix @faustbrian 
Once everything is okay, I will tag a new release for `@arkecosystem/crypto`

## Types of changes

## Checklist


fix:
just sure to understand,
Is there a replacement of HDNode? This is necessary for HD wallet. How this will be used from now on?

fix:
The idea of ECPair is to be able to apply derivation path to it in order to be bip43/44 compatible.

This is not used yet (a part in ledger key). But this is needed. Couple of applications:
- in ARKPay, being able to generate a new public key/address without even generating the private key, so even the server is hacked, there is no security issue on the wallets
- simplification of desktop/mobile wallet where you need only one key and generate many wallets.
- backup method in emergency if ledger key is stolen

supaiku0:
> The idea of ECPair is to be able to apply derivation path to it in order to be bip43/44 compatible.
> 
> This is not used yet (a part in ledger key). But this is needed. Couple of applications:
> 
>    in ARKPay, being able to generate a new public key/address without even generating the private key, so even the server is hacked, there is no security issue on the wallets
> 
>    simplification of desktop/mobile wallet where you need only one key and generate many wallets.
> 
>    backup method in emergency if ledger key is stolen

Hmm, wasn't aware of that. I have to read up a bit to comment on that. Unless it's urgent, can we consider it a non-blocker for now so this can be merged?

fix:
Well deprecated arkjs can be used in case. But we will need it. Might be bip32 enough to do this. 

Actually a great exercise would be to add a simple backup ledger nano s app

fix:
But I confirm it is non blocker. Just arkjs is not deprecated until we sort out this aspect

PR #1003, opened 2018-09-16
[WIIP] Feat: core-snapshots-cli

kristjank:
## Proposed changes

The purpose of this plugin is to provide local snapshot functionality, so in case of issues the blockchain can be rebuild locally from own exported data.
The plugin provides a cli interface, with the following available commands:

## Types of changes

## Features
- [x] Create db independent snapshot
- [x] Append new blocks to existing snapshot
- [x] Truncate existing snapshot to lower height
- [x] Slice existing snapshot
- [x] Truncate ark-node to smaller height (import node from snapshot)
- [x] Verify local snapshot
- [x] Import of local snapshot to database
 
## Checklist

## Further improvements/Comments
Speed will be an issue with this logic, as nothing native to specific database is used, but normal bulk inserts are performed (comparing to pg_dump logic). 
So first snapshot will take a while, then next incremental will just add to existing one, also allowing to have history. Estimated time of processing/importing 260K blocks is 2 minutes!

Tests are open for a new PR.


packages/core-snapshots-cli/CHANGELOG.md
```diff
@@ -0,0 +1,13 @@
...
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
+and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
+
+## Unreleased
+
+
+## 0.0.1 - 2018-09-05
```
faustbrian: Starts at `0.1.0` as every other changelog shows.

packages/core-database-postgres/lib/connection.js
```diff
@@ -361,8 +361,6 @@ module.exports = class PostgresConnection extends ConnectionInterface {
     if (block.transactions.length > 0) {
       this.asyncTransaction.push(this.db.transactions.create(block.transactions))
     }
-
-    await this.db.tx(t => t.batch(this.asyncTransaction))
```
faustbrian: Why was this removed? The transaction will no longer be executed and no data be stored.

packages/core-snapshots-cli/package.json
```diff
@@ -0,0 +1,53 @@
+{
+  "private": true,
+  "name": "@arkecosystem/core-snapshots",
+  "description": "Provides live local streamed snapshots functionality for ARK Core",
+  "version": "0.0.1",
```
faustbrian: Starts at `0.1.0` as all other `package.json` show.

packages/core-snapshots-cli/package.json
```diff
@@ -0,0 +1,53 @@
...
+  "contributors": [
+    "Kristjan Košič <chris@ark.io>"
+  ],
+  "license": "MIT",
+  "main": "lib/index.js",
+  "bin": {
+    "ark:snapy": "./bin/snapy"
```
faustbrian: Should be `snapshot` instead of `snapy`.

faustbrian:
How is this programmatically invoked?

packages/core-database-postgres/lib/connection.js
kristjank: this is for fast rebuild only... transaction is executed when calling saveblockcommit... this is related.

And the method is no longer async - which is not in line with the core-database interface. this should be a separate PR anyways. just fixed for the import speed purpose.. now

kristjank:
for now only from cli

packages/core-database-postgres/lib/connection.js
faustbrian: JavaScript itself has no interfaces so the "fake interface" is not a guarantee as implementations will differ between libraries. We shouldn't and can't take those interfaces as guarantees as some libraries are async and some sync.

kristjank: ok, then will leave it as is. thx

fix: i think this should be left as is.
However if sync is not needed (ie waiting for blocks being saved), we can call it without await.

as for transactions, the tx should be saved as well. (only header is saved on blocks table)

fix: i mean remove await here https://github.com/ArkEcosystem/core/blob/7e88a572fd09fb8bb8edd21746c340e17de725ee/packages/core-blockchain/lib/blockchain.js#L273

fix: but imo this is not the bottleneck

kristjank:
@faustbrian can you check/confirm added stuff to postgres...  and overall review. Naming of deleteGtHeight - maybe some better idea for that 👍

packages/core-database-postgres/lib/connection.js
kristjank: actually I call this for get blocks

https://github.com/ArkEcosystem/core/blob/194fd3e2f9e2685460d5e41b972cfbec7bfe44c4/packages/core-database-postgres/lib/connection.js#L536-L560

kristjank:
closing this, will send a new one, found solution for drastic speed improvements

kristjank:
keep the branch alive...

PR #1007, opened 2018-09-19
fix: add enabled check to webhook

galperins4:
Currently webhook logic does not check to see if the subscription is enabled or disabled. This adds that check into the webhook logic.

## Proposed changes
Add enabled/disabled check for webhook event match. Not sure if it was considered a bug or should have been a feature so checked both just in case 😄 

## Types of changes

## Checklist
- [X] I have read the [CONTRIBUTING](https://docs.ark.io/developers/guidelines/contributing.html) documentation


packages/core-webhooks/lib/manager.js
```diff
@@ -50,23 +50,24 @@ class WebhookManager {
     const matches = []
 
     for (const webhook of webhooks) {
-      if (!webhook.conditions) {
-        matches.push(webhook)
+      if (webhook.enabled == 1) {
```
faustbrian: This should be one of the following.

`if (webhook.enabled)`
`if (webhook.enabled ===true)`
`if (!!webhook.enabled)`

faustbrian: Also, instead of wrapping the whole block into an `if` you could just do an early exit.

```
if(!webhook.enabled) {
  continue;
}
```

galperins4: enabled is stored as 1 or 0 in sqlite (as tinyInt and not boolean). Enabled = 1, Disabled = 0

galperins4: done, this is simpler. Updated code with change

packages/core-webhooks/lib/manager.js
```diff
@@ -50,22 +50,23 @@ class WebhookManager {
...
-        if (!webhook.conditions) {
-          matches.push(webhook)
-
-          continue
-        }
-
-        for (const condition of webhook.conditions) {
-          const satisfies = require(`./conditions/${condition.condition}`)
+      
+      if(webhook.enabled == 0) {
```
faustbrian: If you want to assert that it is false you should do `if(!webhook.enabled)`.

PR #1014, opened 2018-09-20
use a mock server to test forger Client http calls

paroxysm:
## Proposed changes
Makes testing core-forger classes more efficient and reduce code-smell by isolating them behind a mock http server.
Part 2 of this will be to refactor manager.js in core-forger to also leverage the mock server and improve the tests around it. I'll also change the setup.js config to not load, the now unnecessary, modules such as p2p or database to further speed up those tests.

## Types of changes

## Checklist
<!--


faustbrian:
Please use https://github.com/ctimmerm/axios-mock-adapter to mock requests. All other packages use it if they need to mock something as we use axios everywhere to send requests, no point in using another mocking package.

You can check https://github.com/ArkEcosystem/core/blob/develop/packages/core-json-rpc/__tests__/blocks.test.js for some examples.

paroxysm:
I tried this out before using nock. axios-adapter doesn't appear to have a way to match the parameters to an endpoint invocation. ie.
I can do `m.onPost('/interval/blocks').reply(200)`
but cannot `m.onPost('internal/blocks', { block:{ id === <some.value> } }').reply(200)`
So I can test that a POST was sent, but I cannot apply assertions on the data that was sent.

packages/core-forger/__tests__/client.test.js
```diff
@@ -46,8 +54,13 @@ describe('Client', () => {
 
     describe('when the host is available', () => {
       it('should be truthy if broadcasts', async () => {
+        // arrange
```
faustbrian: Please remove all those arrange, act, assert comments.

packages/core-forger/lib/client.js
```diff
@@ -159,4 +151,7 @@ module.exports = class Client {
   async __get (url) {
     return axios.get(url, { headers: this.headers, timeout: 2000 })
   }
+  async __post (url, body) {
```
faustbrian: Add a space before this line.

paroxysm: do u mean a newline?

faustbrian:
Haven't needed param assertions in some time but if I remember right you can do this with axios-mock.

```
axiosMock.onPost(/.../).reply(({ params }) => {
    expect(params.published).toBeTrue()

    return [200, data]
})
```

packages/core-forger/lib/client.js
faustbrian: Yeah

paroxysm:
It ends up being really convoluted ....
```
describe('when the host is available', () => {
      it('should be truthy if broadcasts', async () => {
        mockAxios.onPost(`${host}/internal/blocks`).reply(c => {
          expect(JSON.parse(c.data).block.id).toEqual(block.data.id)
          return [200, true];
        })
        // nock(host).post('/internal/blocks', body => body.block.id === block.data.id).reply(200)

        await client.__chooseHost()

        const wasBroadcasted = await client.broadcast(block.toRawJson())
        expect(wasBroadcasted).toBeTruthy()
      })
    })
```
Not exactly a fan, but it gets the job done. Just hope it doesn't become tech-debt

paroxysm:
Well, I'd rather avoid creating more work for you and write it once. If using axios-mock-adapter does that, I'm fine with it.

paroxysm:
Well, it seems that it lacks the ability to enforce that a matcher happened, and subsequently fail the test if it didn't. @faustbrian  it might be well-served long term to choose the more robust tool. This adapter seems to have been designed to be as simple as possible

PR #1018, opened 2018-09-20
Fix how the client connects to endpoints depending on the API version

j-a-m-l:
## Proposed changes

https://github.com/p-meier/hapi-api-version checks the `api-version` header and then it rewrites the URL to the versioned version (`/api/blocks` to `/api/v2/blocks`). Since during the `OPTIONS` request, `axios` doesn't include custom headers (such as the necessary `api-version`), those API client requests were always resolved to a V1 endpoint.

When the endpoint was existing on V1 and V2 (like `/api/delegates`) the behaviour was successful, but on V2-only endpoints, this was provoking a problem, because the pre-flight `OPTIONS` request was redirected to an nonexistent endpoint.

This PR removes the usage of the `api-version` header, but adds the version to the API endpoint URL.

## Types of changes
- [x] Docs (documentation only changes)

## Further comments

Other option is using an [`Accept` header](https://github.com/p-meier/hapi-api-version#accept-header), but it wouldn't work unless renaming the current `vendorName` (https://github.com/ArkEcosystem/core/blob/46e550628f421539e50caaab97ba5fea44e9f176/packages/core-api/lib/server.js#L51) to remove the "-" or sending a PR request to https://github.com/p-meier/hapi-api-version to accept them.

### DIFF
--- a/packages/client/__tests__/client.test.js
+++ b/packages/client/__tests__/client.test.js
@@ -5,7 +5,7 @@ const { arrayContaining } = expect
 
 const Client = require('../lib')
 const HttpClient = require('../lib/http')
-const ApiResource = require('../lib/resources/v1/transactions')
+const TransactionsResource = require('../lib/resources/v2/transactions')
 const initialPeers = require('../lib/peers')
 
 // https://github.com/facebook/jest/issues/3601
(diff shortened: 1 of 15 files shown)


faustbrian:
This is a breaking change and it is no longer possible to use domains as hosts.

If you have an nginx virtual host that routes your domain to `http://localhost:4003/api/v2` and you try to use that with this client your request will go to `http://localhost:4003/api/v2/api/v2`.

Issue #1023, opened 2018-09-24
Improve Handling Low Disk Space


vRobM:
When the logging goes nuts and writes 34GB of logs in a day, it kills the node.

After cleanup, I noticed that the peers.json is corrupt and needed a re-download for the relay to start, and the forger wouldn't resume until it was reconfigured in commander.

How do we make the Ark Core more resilient to these failure modes?
Check before file writes so we don't get 0 byte files and lose configurations.
Check for corruptness and restore from backup.
Database may also fail to write, and if so, should fail gracefully to be ready to resume operations when the situation is resolved.
What else?

paroxysm:
Not sure what the solution would be if the database fails to write because your node has run out of free space, I think the node-operator should be monitoring their free space and generate alerts.
As far as logging, I believe, doing a size roll over in addition to daily roll over should fix that issue.

vRobM:
The code could attempt to detect such an issue with the database, and if so roll back and wait in a consistent state.

Many such operations in commander don't have checks for success, for example service start. It will issue the start command and regardless of the outcome report success. Log feedback would be useful to report success only if the application reports success for that operation as intended. I realize this could be a PR for commander, but I am using it as an analogy here instead.

As for node operators, you can't count on them to do the right thing, so making the code safer against a disaster is a better option. Ark is user friendly right?

faustbrian:
Ark being end-user friendly doesn't mean that the people that operate the network should run nodes with a _set and forget_ motto. There is a big difference between providing an end-user friendly experience and what the people that operate the network can expect.

If you run a node, no matter if a forger or relay, you should have proper resource monitoring on your system instead of relying on a single piece of software to do **everything** for you.

faustbrian:
Will close this as this is an issue outside the scope of core itself.

PR #1027, opened 2018-09-24
Add size-based log rolling in addition to daily rotation

paroxysm:
## Proposed changes
We already rotate logs daily. This also adds a size-based rolling set to 100M

## Types of changes

## Further comments
This would partly address the "crazy logging" pointed in issue #1023


faustbrian:
This should be added to all configuration files, not only mainnet and devnet.

paroxysm:
1 sec.

paroxysm:
@faustbrian  done.

PR #1030, opened 2018-09-25
feat: HDWallet

### LINKED ISSUE #1001: manage HD Wallets
In order to completely deprecate ark-js the package crypto needs to deal with HDWallet with the following features:
- given a key pair, apply whatever derivation path in the form explained in bip32
- support bip44 (ie default bip32 to m/44'/111'/.... + have simple functions to navigate to the wallets)

supaiku0:
## Proposed changes

`arkjs` uses the `HDNode` class to deal with HD wallets. It originated from the bitcoinjs-lib, but eventually got [removed upstream](https://github.com/bitcoinjs/bitcoinjs-lib/commit/884f3fd57d910637c7ae8ea8219d261571458cd9). The new way to deal with HD wallets is to use the `bip32` module.

A few features are not provided by `bip32`, including the ability to convert a keypair to a node and vice versa. This PR adds a utility class `HDWallet` with a  few convenience functions to get a BIP32 node or turn one into a keypair.

A BIP32 node can be derived by path or manually, see tests for usage. It's basically the same as before.

Resolves #1001.

## Types of changes

## Checklist

- [X] I have added necessary documentation (if appropriate)
<!--


fix:
- it should be possible to derive only the publickey without the knowledge of privatekey
ie given a public key you can derive any sub path.

use case: 
- you have created hd wallet from a seed, offline
- you calculate the publickey deriving `/44'/111'/0'` offline
- you put that public key on a connected computer
- you generate whatever wallet on demand deriving that public key with `<wallet number>'/0`
- in case o tx signing you perform offline derivation and signature of your tx with the associated public key path

fix:
AIP 20 draft for reference https://github.com/ArkEcosystem/AIPs/issues/29

supaiku0:
I'll look into it.

supaiku0:
@fix How about this:
Add new network config key `aip20` which is read by HDWallet to derive the network node.

Then the equivalent to your example would be:
```javascript
const node = HDWallet.deriveNetwork()
    .deriveHardened(0) // account
    .derive(0) // address_index
```

Also, how should the testnet be identified?

---

Now regarding your first comment, you can create a node from a [public key](https://github.com/bitcoinjs/bip32/blob/master/index.js#L271) with `bip32` without a private key. Example:
```javascript
const node = bip32.fromPublicKey(buffer, chainCode)
...
```
The use case you mentioned should be supported.

PR #1032, opened 2018-09-26
fix(client): use the accept header instead of api-version

luciorubeens:
## Proposed changes

As explained by @j-a-m-l in #1018, we need to use the accept header to make it work in the browser.

## Types of changes


j-a-m-l:
@luciorubeens Once that https://github.com/p-meier/hapi-api-version/pull/22 is merged, we should change the `vendorName` to `ark.core-api` instead of `arkCoreApi`

faustbrian:
@j-a-m-l @luciorubeens do you want to keep this open until the vendor can be adjusted or do you adjust it later on?

luciorubeens:
I leave it for later, since it is not something that will break the project.

PR #1042, opened 2018-09-30
use else if, return if delegate reg

JeremiGendron:
## Proposed changes

small fix, improves serialization

## Types of changes

- [x] Other... Please describe:

todo

## Checklist

### DIFF
--- a/packages/crypto/lib/models/transaction.js
+++ b/packages/crypto/lib/models/transaction.js
@@ -243,7 +243,6 @@ module.exports = class Transaction {
     }
 
     else if (transaction.type === TRANSACTION_TYPES.DELEGATE_RESIGNATION) {
-      return
     }
 
     if (transaction.signature) {
--- a/packages/crypto/lib/models/transaction.js
+++ b/packages/crypto/lib/models/transaction.js
@@ -243,6 +243,7 @@ module.exports = class Transaction {
     }
 
     else if (transaction.type === TRANSACTION_TYPES.DELEGATE_RESIGNATION) {
+       // delegate resignation - empty payload 
     }
 
     if (transaction.signature) {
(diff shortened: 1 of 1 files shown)


packages/crypto/lib/models/transaction.js
```diff
@@ -223,28 +222,28 @@ module.exports = class Transaction {
...
       transaction.asset.payments.forEach(p => {
         bb.writeUInt64(p.amount)
         bb.append(bs58check.decode(p.recipientId))
       })
     }
 
-    if (transaction.type === TRANSACTION_TYPES.DELEGATE_RESIGNATION) {
-      // delegate resignation - empty payload
+    else if (transaction.type === TRANSACTION_TYPES.DELEGATE_RESIGNATION) {
+      return
```
supaiku0: Why do you `return` here? It completely break the serialization for delegate resignations. What `empty payload` means is that a delegate resignation has no asset. The `if` is just there for the sake of completeness. So please keep the comment and remove the return. Otherwise it's fine.

faustbrian: What is the reason for returning here?

JeremiGendron: says in previous comment that there is no payload to be serialized for delegate registration

JeremiGendron: my mistake, should instead leave it empty, applying changes now

packages/crypto/lib/models/transaction.js
```diff
@@ -243,7 +243,6 @@ module.exports = class Transaction {
     }
 
     else if (transaction.type === TRANSACTION_TYPES.DELEGATE_RESIGNATION) {
-      return
```
faustbrian: The comment should be added back.

packages/crypto/lib/models/transaction.js
faustbrian: As @supaiku0 explained, the payload being empty only means that there is no asset to be processed but you still need to continue the processing otherwise.

PR #1046, opened 2018-10-02
fix: idHex calculation and getHeader

supaiku0:
## Proposed changes

`idHex` was calculated in the wrong order, so that it was always overwritten in `Block.serialize` when calling `applyV1Fixes` while the Block has no `id` resulting in `0000000000000000`.

Instead of
```javascript
this.data.idHex = Block.getIdHex(this.data)
this.data.id = Block.getId(this.data) // <-- overwrites idHex
```
it should be:
```javascript
this.data.id = Block.getId(this.data)
this.data.idHex = Block.getIdHex(this.data)
```

But since `idHex` is not (yet?) used anywhere, it went by unnoticed. This leads to the original issue I was after: passing undefined into `bigi` would yield a zero buffer, while `bignumber.js` yields NaN causing idHex to turn into `0000000000000NaN`. Luckily, this is a non-issue because it is not used for anything. I used this chance to remove some unnecessary code.

I also believe that `getHeader` should not remove the transactions from `this.data`:
```javascript
getHeader () {
  const header = this.data
  delete header.transactions
  return header
}
```
Removing the transactions could potentially lead to weird bugs when called in the wrong moment and should be avoided, since `getHeader` is being called on `getLastBlock`.

## Types of changes

## Checklist


packages/crypto/lib/models/block.js
```diff
@@ -223,7 +212,7 @@ module.exports = class Block {
         }
       }
 
-      if (!constants.reward === block.reward.toNumber()) {
+      if (!(constants.reward === block.reward.toNumber())) {
```
faustbrian: Couldn't that simply be `if (constants.reward !== block.reward.toNumber())`?

PR #1054, opened 2018-10-03
Fix: Saving to database only when saveBlockCommit is called

kristjank:
## Proposed changes
Fix: Saving to database only when saveBlockCommit is called. This is related when fastRebuild is active, when we download 10000 of blocks, save them in memory with `saveBlockAsync` and then call saveBlockCommit.

Current implementation was calling insert on every block added to `this.asyncTransaction`.

Increased limit from 10000 to 20000 as postgres and node memory can handle this with ease.

## Types of changes

## Checklist

### DIFF
--- a/packages/core-database-postgres/lib/connection.js
+++ b/packages/core-database-postgres/lib/connection.js
@@ -346,7 +346,7 @@ module.exports = class PostgresConnection extends ConnectionInterface {
   }
 
   /**
-   * Stores the bloc in memory, in the this.asyncTransaction, to be later saved to the database by calling saveBlockCommit.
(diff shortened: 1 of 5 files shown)


packages/core-database-postgres/lib/connection.js
```diff
@@ -351,7 +351,7 @@ module.exports = class PostgresConnection extends ConnectionInterface {
    * @param  {Block} block
    * @return {void}
    */
-  async saveBlockAsync (block) {
+  saveBlockAsync (block) {
```
supaiku0: Should be renamed and the doc updated, it doesn't save anything. Maybe something like `queueSaveBlockAsync`?

kristjank: even better

PR #1058, opened 2018-10-04
Refactor tx serialized to hexstring

supaiku0:
## Proposed changes

This patch aligns the type of `tx.serialized` with `block.serialized` and changes it from a `Buffer` to a hexstring. Most places called `toString('hex')` on  `tx.serialized` anyway and this change allows to drop some code which was needed, because `Block` and `Transaction` were inconsistent.

## Types of changes

## Checklist

<!--


faustbrian:
Tests pass locally but currently fail on Travis, will look into it tomorrow if they still fail then.

PR #1066, opened 2018-10-05
feat: core-transaction-pool-mem

vasild:
## Proposed changes

Replace the redis transactions pool with a hybrid one that stores the data in memory and serves all queries from there and occasionally syncs/saves to a permanent storage (postgresql).

## Types of changes

## Checklist


packages/core-database-postgres/lib/connection.js
```diff
@@ -589,6 +589,52 @@ module.exports = class PostgresConnection extends ConnectionInterface {
...
   }
 
+  /* Transaction pool aux methods @{ */
+
+  /**
+   * Add transactions to the transaction pool.
+   * @param {Array of Transaction objects} transactions
+   * @return {void}
+   */
+  async transactionPoolAdd (transactions) {
```
faustbrian: This should not be part of `core-database-postgres` because you are hardcoding transaction pool logic into a package that is supposed to be decoupled from it.

packages/core-database-postgres/lib/migrations/20180928100000-create-pool.sql
```diff
@@ -0,0 +1,22 @@
+-- Table Definition
```
faustbrian: This should not be part of `core-database-postgres` because you are hardcoding transaction pool logic into a package that is supposed to be decoupled from it.

If you want to store data in the postgres database and add migrations you should do that in your transaction pool package.

packages/core-database-postgres/lib/models/pool.js
```diff
@@ -0,0 +1,26 @@
+const Model = require('./model')
```
faustbrian: This should not be part of `core-database-postgres` because you are hardcoding transaction pool logic into a package that is supposed to be decoupled from it.

If you want to store data in the postgres database and add migrations you should do that in your transaction pool package.

faustbrian:
All logic that is currently in `core-database-postgres` should be moved out to the transaction pool package itself because the current implementation is to tight coupled to postgres.

The transaction pool is something that should be easily replaceable so if you want to tie it to postgresql you should do that inside your package and use the connection created by core-database-postgres to perform all your postgres specific actions.

packages/core-database-postgres/lib/connection.js
vasild: Thanks for looking into this. Indeed it looks a bit strange to have it there. Where else would you suggest to put it?

Btw, assuming we want to store the data in a table in postgres, some functions must be added to core-database-postgres to fiddle with that table?

faustbrian:
I would say overall postgresql seems a bit overkill for the amount of data a transaction pool will have. Data goes in and out so fast that the benefit of a system like postgresql for persistent data won't come to use.

I would recommend to either use sqlite or leveldb **(if you hate life)** as those are easy to integrate and something like sqlite will provide you with the same query functionality as postgresql while being lightweight and having the added bonus of being unaffected by a node performing a rebuild from scratch where the database gets dropped.

vasild:
The operations that are done with the persistent storage are the following:

- add a new transaction (insert one row in the table)
- delete a given transaction by id (delete one row from the table, given its id)
- delete all transactions
- read all transactions, in insertion order

The first two are usually done in batches, so it would be good to have some optimized way to e.g. insert 500 transactions at once, or delete 500 by id.

I chose PostgreSQL for this solely because it is already up and running for other reasons. So its usage comes for free, from the point of view of extra dependencies and maintenance in the code and maintenance by the node operators. In my previous implementation which was based on the master branch I used the generic sequelize library, with sqlite as an engine. But both sequelize and sqlite were removed in the develop branch.

faustbrian:
SQLite was removed as it wasn't suiteable for running a blockchain on it as it would constantly get a deadlock which isn't exactly something you want.

For something like a transaction pool it is completely fine as the overall size of the database is small and the batches of data that get in fast are already limited by how much data comes in via API which is throttled and has a transaction limit per request.

I don't really see where something like 500 transactions would cause such an overhead that SQLite would cause performance issues and you would need to switch to something like PostgreSQL.

packages/core-transaction-pool-mem/package.json
```diff
@@ -21,6 +21,7 @@
     "depcheck": "depcheck ./"
   },
   "dependencies": {
+    "@arkecosystem/core-api": "^0.1.1",
```
faustbrian: Why is this required by the transaction pool? The transaction pool shouldn't have a dependency on anything other then the container and pool interface 👀

vasild: To convert "number of seconds since genesis block" to "number of seconds since 1970". Now changed in https://github.com/ArkEcosystem/core/pull/1066/commits/777ac2222482164f9c5a9f550e385e896476e275

packages/core-transaction-pool-mem/lib/defaults.js
```diff
@@ -2,6 +2,11 @@
 
 module.exports = {
   enabled: true,
+  // How often to save/sync the transaction pool to a persistent storage.
+  // The number designates count of added or deleted transactions that are not in
+  // the persistent storage. When that number is reached a new sync of the pool is
+  // triggered.
+  syncToDiskMods: 512,
```
faustbrian: I would recommend to rename this to something like `syncInterval` as it will make the comment obsolete because `Interval` indicates that the `sync` will happen periodically at that number.

vasild: Done in 7d536a3d

packages/core-database-postgres/lib/connection.js
vasild: All PostgreSQL traces obliterated in a59ac681.

packages/core-transaction-pool-mem/lib/storage.js
```diff
@@ -0,0 +1,100 @@
...
+        'Transaction pool: could not switch SQLite journal mode to WAL. ' +
+        `The journal mode is '${newMode}'. Expect reduced performance.`)
+    }
+
+    this.db.exec(`
+      CREATE TABLE IF NOT EXISTS ${this.table} (
+        "sequence" INTEGER PRIMARY KEY AUTOINCREMENT,
+        "id" VARCHAR(64) UNIQUE,
+        "senderPublicKey" VARCHAR(66) NOT NULL,
+        "serialized" TEXT NOT NULL
```
faustbrian: I would replace this with `"serialized" bytea NOT NULL` to be consistent with how the `transactions` table stores the serialized hex.

vasild: https://www.postgresql.org/docs/10/static/datatype-binary.html#id-1.5.7.12.9

https://www.sqlite.org/datatype3.html

no `bytea` in SQLite?

faustbrian: I edited my comment shortly after, `BLOB` should be the equivalent in SQLite.

vasild: To illustrate:
```
from db (binary) ->
convert to hex [1] ->
pass to the constructor of Transaction ->
pass to Transaction.deserialize() ->
convert to binary using ByteBuffer.fromHex(hexString, true) [2] ->
play with the binary
```
[1] and [2] can/should be removed. But this is outside of the scope of this tx mem pool / redis replacement.

supaiku0: Yes, I agree. Let's reiterate that once mainnet is on v2.

faustbrian:
@vasild What is the status on this PR, ready to be tested or still work in progress?

vasild:
> @vasild What is the status on this PR, ready to be tested or still work in progress?

It is production ready. The code I write usually has no bugs. However, to be on the safe side, I would advise to test it heavily as much as possible. Just to rule out the _theoretical_ possibility that there may be some problems with it.

My short term todo:
- replay redis changes from a4250d4f to the mem pool
- add method to check if a given sender has already vote/unvote transaction in the pool (as asked by @kristjank)
- increase the test coverage (write more tests)

Also:
- actually plug this into devnet and see how it behaves

faustbrian:
Why do I need to test it if your code is bug free? 🤔

vasild:
> Why do I need to test it if your code is bug free? thinking

Just to rule out the _theoretical_ possibility that there may be some problems with it.

I forgot to mention long term todo:
- rule the world!

packages/core-transaction-pool-mem/lib/connection.js
```diff
@@ -0,0 +1,340 @@
...
+   * the on-disk database, saved there from a previous run.
+   * @return {TransactionPool}
+   */
+  make () {
+    this.mem = new Mem()
+
+    this.storage = new Storage()
+
+    const allSerialized = this.storage.loadAllInInsertionOrder()
+    allSerialized.map(s => this.mem.add(new Transaction(s), this.options.maxTransactionAge, true))
```
supaiku0: Why `.map` instead of `.forEach`?

packages/core-transaction-pool-mem/lib/connection.js
```diff
@@ -0,0 +1,340 @@
...
+
+    this.__syncToPersistentStorageIfNecessary()
+  }
+
+  /**
+   * Add many transactions to the pool.
+   * @param {Array}   transactions, already transformed and verified by transaction guard - must have serialized field
+   */
+  addTransactions (transactions) {
+    transactions.map(t => this.addTransaction(t))
```
supaiku0: Why `.map` instead of `.forEach`?

packages/core-transaction-pool-mem/lib/connection.js
```diff
@@ -0,0 +1,340 @@
...
+    this.__syncToPersistentStorageIfNecessary()
+  }
+
+  /**
+   * Remove multiple transactions from the pool (by object).
+   * @param  {Array} transactions
+   * @return {void}
+   */
+  removeTransactions (transactions) {
+    transactions.map(t => this.removeTransaction(t))
```
supaiku0: Why `.map` instead of `.forEach`?

packages/core-transaction-pool-mem/lib/connection.js
```diff
@@ -0,0 +1,340 @@
...
+
+  /**
+   * Removes any transactions in the pool that have already been forged.
+   * @param  {Array} transactionIds
+   * @return {Array} IDs of pending transactions that have yet to be forged.
+   */
+  async removeForgedAndGetPending (transactionIds) {
+    const forgedIdsSet = new Set(await database.getForgedTransactionsIds(transactionIds))
+
+    Array.from(forgedIdsSet).map(id => this.removeTransactionById(id))
```
supaiku0: Why `.map` instead of `.forEach`?

packages/core-transaction-pool-mem/lib/storage.js
```diff
@@ -0,0 +1,103 @@
...
+    if (data.length === 0) {
+      return
+    }
+
+    const insertStatement = this.db.prepare(
+      `INSERT INTO ${this.table} (id, serialized) VALUES (:id, :serialized);`)
+
+    this.db.prepare('BEGIN;').run()
+
+    data.map(d => insertStatement.run({
```
supaiku0: Why `.map` instead of `.forEach`?

packages/core-transaction-pool-mem/lib/storage.js
```diff
@@ -0,0 +1,103 @@
...
+    if (ids.length === 0) {
+      return
+    }
+
+    const deleteStatement = this.db.prepare(
+      `DELETE FROM ${this.table} WHERE id = :id;`)
+
+    this.db.prepare('BEGIN;').run()
+
+    ids.map(id => deleteStatement.run({ id: id }))
```
supaiku0: Why `.map` instead of `.forEach`?

packages/core-transaction-pool-mem/lib/connection.js
vasild: Oh, well, good catch! Changed all in 56a1f0ec. Thanks!

packages/core-transaction-pool-mem/lib/defaults.js
```diff
@@ -0,0 +1,15 @@
...
+  enabled: true,
+  // How often to save/sync the transaction pool to a persistent storage.
+  // The number designates count of added or deleted transactions that are not in
+  // the persistent storage. When that number is reached a new sync of the pool is
+  // triggered.
+  syncInterval: 512,
+  maxTransactionsPerSender: process.env.ARK_TRANSACTION_POOL_MAX_PER_SENDER || 300,
+  whitelist: [],
+  allowedSenders: [],
+  maxTransactionsPerRequest: 200,
```
zillionn: What if I change `config.getConstants(height).block.maxTransactions` to 250, will this still work?

packages/core-transaction-pool-mem/lib/mem.js
```diff
@@ -0,0 +1,251 @@
...
+   * Get the expired transactions.
+   * @return {Array of Transaction} expired transactions
+   */
+  getExpired () {
+    const now = new Date()
+
+    let transactions = []
+
+    for (let e of this.idsByExpiration) {
+      if (e.expireAt < now) {
```
supaiku0: Mhh not that it makes a big difference, but shouldn't this be semantically `less or equal`?

packages/core-transaction-pool-mem/lib/connection.js
```diff
@@ -0,0 +1,340 @@
...
+    return this.mem.getIdsBySender(senderPublicKey).size
+  }
+
+  /**
+   * Add a transaction to the pool.
+   * @param {Transaction} transaction
+   */
+  addTransaction (transaction) {
+    if (this.transactionExists(transaction.id)) {
+      return logger.debug(
```
supaiku0: While it works just fine and yields the expected `undefined`, let's not couple `return` with `logger`.

packages/core-transaction-pool-mem/lib/connection.js
```diff
@@ -0,0 +1,340 @@
...
+   * @param  {Number} size
+   * @return {(Array|void)} array of serialized transaction hex strings
+   */
+  getTransactions (start, size) {
+    this.__purgeExpired()
+
+    let serializedTransactions = []
+
+    let i = 0
+    for (let transaction of this.mem.getTransactionsInInsertionOrder()) {
```
supaiku0: Minor nitpick, but this `for loop` is the basically same as in `getTransactionIdsForForging` could be moved into it's own function. It would just need to know which `property` to pick.

packages/core-transaction-pool-mem/lib/connection.js
```diff
@@ -0,0 +1,340 @@
...
+   * @param  {Number} size
+   * @return {Array} array of transactions IDs in the specified range
+   */
+  async getTransactionIdsForForging (start, size) {
+    this.__purgeExpired()
+
+    let ids = []
+
+    let i = 0
+    for (let transaction of this.mem.getTransactionsInInsertionOrder()) {
```
supaiku0: Use `for (const ...)` as you already do in a few other places.

packages/core-transaction-pool-mem/lib/defaults.js
faustbrian: @vasild as zillion pointed out, you should use the same value as the maxTransactions per block according to the network config.

packages/core-transaction-pool-mem/lib/storage.js
```diff
@@ -0,0 +1,103 @@
...
+/**
+ * A permanent storage (on-disk), supporting some basic functionalities required
+ * by the transaction pool.
+ */
+class Storage {
+  /**
+   * Construct the storage.
+   */
+  constructor () {
+    this.file = 'transaction_pool.sqlite'
```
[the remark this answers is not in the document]
faustbrian: As @supaiku0 said, you should store it like `storage: process.env.ARK_DB_STORAGE || `${process.env.ARK_PATH_DATA}/database/transaction-pool-${process.env.ARK_NETWORK_NAME}.sqlite`.

packages/core-transaction-pool-mem/lib/connection.js
```diff
@@ -0,0 +1,340 @@
...
+    return this
+  }
+
+  /**
+   * Disconnect from transaction pool.
+   * @return {void}
+   */
+  disconnect () {
+    this.__syncToPersistentStorage()
+    this.db.close()
```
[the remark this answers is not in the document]
vasild: Fixed, also hooked a missing disconnect call at the end of the test (2fb945fc).

packages/core-transaction-pool-mem/lib/defaults.js
supaiku0: @zillionn Yes, the setting is only used on the p2p v1 endpoint. It doesn't make much sense right now either way, so we will make some changes in that regard.

kristjank:
➤ Vasil Dimov commented:

Aha, "completed"... you wish!

PR #1072, opened 2018-10-06
Chore: Docker adjusting the entrypoint script to recent changes

kristjank:
## Proposed changes
Chore: Docker adjusting the entrypoint script to recent changes. We renamed container names in `docker-compose.yml` but did not adjust the `entrypoint.sh`

Tnx @adrian69 

Delete all containers if you where using ark-core from the docker, and start them again.

## Types of changes
- [x] Docker env

## Checklist

### DIFF
--- a/docker/devnet/entrypoint.sh
+++ b/docker/devnet/entrypoint.sh
@@ -2,9 +2,9 @@
 
 sysctl -w net.ipv4.conf.all.route_localnet=1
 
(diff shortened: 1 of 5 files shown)


adrian69:
```Delete all containers if you where using ark-core from the docker, and start them again.```
Most important is to delete the previous ark-core docker image so compose build it again, otherwise it will have the unmodified script inside the new container and fail again.

PR #1078, opened 2018-10-08
fix: update peer status from http status + manage 'OK' special case + improve suspended peer logging

air1one:
## Proposed changes

For p2p, peer status corresponds to the last request' http status, but there was some cases (status codes not in 200 range) where it would not be correctly updated : this fixes it.

Also there are some 'OK' statuses which are used in API, I updated it so we only are sending these 'OK' statuses for API and not for the rest of the core logic.

Finally improved the 'suspended peer' logging so that we can have a human-readable estimate of the suspension time left with moment's humanize() function.

## Types of changes

## Checklist


packages/core-api/lib/versions/2/handlers/peers.js
```diff
@@ -18,7 +18,12 @@ exports.index = {
   async handler (request, h) {
     const allPeers = await blockchain.p2p.getPeers()
 
-    let result = allPeers.sort((a, b) => a.delay - b.delay)
+    let result = allPeers.map(peer => {
+        // just use 'OK' status for API instead of p2p http status codes
```
faustbrian: For the v2 API we can simply return the numerical value as we don't need this OK hack for backwards compatibility.

PR #1090, opened 2018-10-11
misc: add retry delay when calling __chooseHost on forger startup

supaiku0:
## Proposed changes

When starting a forger and relay simultaneously, the forger tries to connect to the relay immediately, even though it is not ready causing this:
image

I added a small delay, so that calling `__chooseHost` during forger startup doesn't flood the poor relay with requests.

## Types of changes

## Checklist

### DIFF
--- a/packages/core-forger/lib/client.js
+++ b/packages/core-forger/lib/client.js
@@ -1,6 +1,7 @@
 'use strict'
 const axios = require('axios')
 const sample = require('lodash/sample')
+const delay = require('delay')
 const container = require('@arkecosystem/core-container')
 const logger = container.resolvePlugin('logger')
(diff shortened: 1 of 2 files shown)


packages/core-forger/lib/client.js
```diff
@@ -144,6 +145,10 @@ module.exports = class Client {
     } catch (error) {
       logger.debug(`${host} didn't respond to the forger. Trying another host :sparkler:`)
 
+      if (wait > 0) {
+        await delay(wait)
+      }
+
       await this.__chooseHost()
```
zillionn: Shouldn't you pass the `wait` in the recursion?

PR #1091, opened 2018-10-11
fix: getTransactionsForForging : use await for IF condition to work as expected

air1one:
## Proposed changes

Fixed IF condition in getTransactionsForForging which was not correct as checkApplyToBlockchain is async and needs to be called with await.

## Types of changes

## Checklist


packages/core-transaction-pool-redis/__tests__/connection.test.js
```diff
@@ -21,6 +23,10 @@ afterAll(async () => {
   await app.tearDown()
 })
 
+beforeEach(async () => {
```
vasild: What is the reason for this change? We already have

```
afterEach(async () => {
  await connection.flush()
})
```
doing the same at `beforeEach` looks unnecessary / redundant. Maybe I miss something?

PR #1097, opened 2018-10-11
refactor: replace toString with toFixed

faustbrian:
## Proposed changes

This PR will replace all toString and toNumber calls with toFixed so that we won't into issues like on v1.

## Types of changes

## Checklist


faustbrian:
I will test this for a bit tomorrow before we merge it to devnet.

packages/core-api/lib/versions/1/transformers/block.js
```diff
@@ -17,10 +18,10 @@ module.exports = (model) => {
     previousBlock: model.previousBlock,
     height: model.height,
     numberOfTransactions: model.numberOfTransactions,
-    totalAmount: Number(model.totalAmount),
-    totalForged: (Number(model.reward) + Number(model.totalFee)).toString(),
-    totalFee: Number(model.totalFee),
-    reward: Number(model.reward),
+    totalAmount: +bignumify(model.totalAmount).toFixed(),
+    totalForged: +bignumify(model.reward).plus(model.totalFee).toFixed(),
```
air1one: You replace `.toString()` by `+.....toFixed()` so instead of a string you get a number, is it ok ?

packages/core-tester-cli/lib/commands/command.js
```diff
@@ -178,7 +178,7 @@ module.exports = class Command {
    */
   static parseFee (fee) {
     if (typeof fee === 'string' && fee.indexOf('-') !== -1) {
-      const feeRange = fee.split('-').map(f => new Bignum(f).toNumber())
+      const feeRange = fee.split('-').map(f => +(new Bignum(f)).toFixed())
```
air1one: no bignumify here ? :scream:

packages/crypto/lib/crypto/crypto.js
```diff
@@ -132,8 +133,8 @@ class Crypto {
       }
     }
 
-    bb.writeLong(+transaction.amount.toString())
-    bb.writeLong(+transaction.fee.toString())
+    bb.writeLong(+(new Bignum(transaction.amount)).toFixed())
```
air1one: no bignumify :scream:

packages/crypto/lib/models/block.js
```diff
@@ -162,7 +162,7 @@ module.exports = class Block {
 
   static getId (data) {
     const idHex = Block.getIdHex(data)
-    return new Bignum(idHex, 16).toString()
+    return new Bignum(idHex, 16).toFixed()
```
air1one: bignumify or with parenthesis like this `(new Bignum(idHex, 16)).toFixed()`, not sure it gives the same result without parenthesis :thinking:

faustbrian: This is the crypto package, `bignumify` is inside the `core-utils` which has a dependency of `crypto` so it would create a circular dependency.

packages/core-api/lib/versions/1/transformers/block.js
faustbrian: Should be fine, we mainly had it as `toString` to prevent any issues.

packages/crypto/lib/models/block.js
air1one: true :ok_hand:

PR #1102, opened 2018-10-12
test: verify that the tx are genesis tx

JeremiGendron:
## Proposed changes

verify that tx ids received from api calls match genesis tx ids (TODO)

## Types of changes

## Checklist


JeremiGendron:
:clinking_glasses:

air1one:
Hey @JeremiGendron could you refactor every time you do this
```
let genesisTransactions = {}
        genesisBlock.transactions.forEach(transaction => {
          genesisTransactions[transaction.id] = true
        })
        const failed = data.some(transaction => {
          if (!genesisTransactions[transaction.id]) return true
        })
         expect(!failed).toBeTrue()
```
And do something like this instead : 
```
expect(genesisBlock.transactions.map(transaction => transaction.id))
  .toEqual(data.map(transaction => transaction.id))
```
Didn't test it but should work as well.

PR #1113, opened 2018-10-14
fix: delete blocks on commit

supaiku0:
## Proposed changes

The same issue as in #1054, but for deleting blocks. Additionally, I refactored a bit, since `saveBlockCommit` and `deleteBlockCommit`did exactly the same and worked with the same array. No point in having two  separate functions.

## Types of changes

## Checklist

<!--


packages/core-blockchain/lib/blockchain.js
```diff
@@ -196,7 +196,8 @@ module.exports = class Blockchain {
       await deleteLastBlock()
     }
 
-    await this.database.deleteBlockCommit()
+    // Commit delete blocks
+    await this.database.commit()
```
kristjank: the only comment I have is maybe naming of commit() method, so it won't mislead to others that this is a general database commit. As it is relate to queuedQueries... maybe just rename this to `commitQueued()` or something similar.

Issue #1061, opened 2018-10-04
Transaction pool sort by fee and limitations


kristjank:
**Is your feature request related to a problem? Please describe.**
Currently there is no sort of transaction by the dynamic fees. This need to be take into account when getting transaction from the pool.

**Describe the solution you'd like**
Select top 150(blocksize) transaction from the pool, and order them by fee DESC. 
So top transactions get forged.

As this can introduce issues for other transactions types, (vote/unvote order) also enforce a limit on the transaction pool by the sender, in the following form:

TO ALLOW ONLY 1 TRANSACTION OF TYPE VOTE/UNVOTE per SENDER_PUB_KEY in Pool
```
IF incoming transaction  === [VOTE,UNVOTE] THEN:
    IF (sender_public_key already has a VOTE/UNVOTE transaction in POOL) THEN:
        REJECT incoming transaction 
        // allowing the vote transaction to be processed as they are sent    
```

**Further improvements**
After this is working we can think of adding overriding options for specific transactions (similar to ETH or BTC)

faustbrian:
Resolved by https://github.com/ArkEcosystem/core/commit/57317354daf0077a84834d548e419fb775adbea1

vasild:
The second part of this has already been implemented in
https://github.com/ArkEcosystem/core/blob/develop/packages/core-transaction-pool/lib/guard.js#L174
?

PR #1120, opened 2018-10-15
feat: guard add limit one per vote in pool

kristjank:
## Proposed changes
Adding limit functionality to pool guard. Limit per sender to have only 1 vote/unvote transaction allowed in pool. This is related to dynamic fees sorting and less mistakes on the client users.

Added the check to the Excess function call to guard as they correlate and avoid duplications in code. So the rejected vote/unvote transaction is then returned in the excess list.

Relates to https://github.com/ArkEcosystem/core/issues/1061 

## Types of changes

## Checklist


packages/core-transaction-pool/__tests__/interface.test.js
```diff
@@ -218,4 +218,10 @@ describe('Transaction Pool Interface', () => {
       expect(poolInterface.checkDynamicFeeMatch).toBeFunction()
     })
   })
+
+  describe('checkIfSenderHasVoteTransactions', () => {
+    it('should be a function', () => {
+      expect(poolInterface.checkIfSenderHasVoteTransactions).toBeFunction()
+    })
+  })
```
vasild: This looks like a duplicate with what is in `packages/core-transaction-pool-mem/__tests__/connection.test.js`:

```
401   describe('checkIfSenderHasVoteTransactions', () => {
402     it('should be a function', () => {
403       expect(connection.checkIfSenderHasVoteTransactions).toBeFunction()
404     })
...
```

Not sure how tests are organized and what are the best practices but I think one of the two can/should be removed.

packages/core-transaction-pool/lib/guard.js
```diff
@@ -187,8 +187,9 @@ module.exports = class TransactionGuard {
   async __determineExcessTransactions () {
     for (let transaction of this.broadcast) {
       const hasExceeded = await this.pool.hasExceededMaxTransactions(transaction)
+      const hasVote = await this.pool.checkIfSenderHasVoteTransactions(transaction.senderPublicKey)
```
vasild: `checkIfSenderHasVoteTransactions()` is not `async`.

packages/core-transaction-pool/__tests__/interface.test.js
kristjank: the interface test should check for method signature and throw when called. 

as for connection.test.js it will test then actual implementation of the pool-mem.

packages/core-transaction-pool/lib/guard.js
kristjank: tnx

packages/core-transaction-pool/lib/guard.js
```diff
@@ -187,8 +187,9 @@ module.exports = class TransactionGuard {
   async __determineExcessTransactions () {
     for (let transaction of this.broadcast) {
       const hasExceeded = await this.pool.hasExceededMaxTransactions(transaction)
+      const hasVote = this.pool.checkIfSenderHasVoteTransactions(transaction.senderPublicKey)
 
-      if (hasExceeded) {
+      if (hasExceeded || hasVote) {
```
faustbrian: If a vote is already present it should be rejected as invalid, not excess.

faustbrian: Excess in this context means that the limit per request was exceeded, meaning throttling kicked in.

Exceeding any other arbitrary limits that are exceeded should be handled as invalid and return an error together with their ID so the user can be informed about why the transaction was rejected.

In your case you would mark the vote transaction as invalid and return an error like `[address/publicKey] already has a pending vote`.

Issue #850, opened 2018-08-12
Calculate new round active delegate list from memory and not from database


fix:
**Is your feature request related to a problem? Please describe.**
Right now the calculation of the delegate list is made from database. This degrades performance  because the wallets that have changed during last round should be synced with database before the calculation occurs. Using very costly upsert.

see the note https://github.com/ArkEcosystem/core/blob/blockchain_fix/packages/core-database-sequelize/lib/connection.js#L323

**Describe the solution you'd like**
The best way would be to to the calculation in wallet manager with a function `calculateDelegateList(delegateMax = 51)` inside https://github.com/ArkEcosystem/core/blob/blockchain_fix/packages/core-database/lib/wallet-manager.js#L16
consisting on 2 separate actions:
- updating each delegate wallets with balance of voters
- sorting top most voted delegate wallets.

and use it instead of the database calculation here https://github.com/ArkEcosystem/core/blob/blockchain_fix/packages/core-database/lib/interface.js#L71 instead of a database implementation

**Describe alternatives you've considered**
Even better would be to update the votes here https://github.com/ArkEcosystem/core/blob/blockchain_fix/packages/core-database/lib/wallet-manager.js inside the 4 functions:
- applyTransaction
- revertTransaction
- applyBlock
- revertBlock

in the end getting the delegate list would be as simple as sorting the most voted wallets

PR #1128, opened 2018-10-16
[WIP] refactor: build delegate list from memory

faustbrian:
## Proposed changes

Start of a solution for https://github.com/ArkEcosystem/core/issues/850.

**With the current changes syncing beyond 441 on devnet doesn't seem to be possible but 178 seems to be longer an issue as my node never gets stuck at it.**

@supaiku0 maybe can continue this today as his workday starts when mine ends :trollface:

## Types of changes

## Checklist


packages/core-database/lib/wallet-manager.js
```diff
@@ -376,6 +376,41 @@ module.exports = class WalletManager {
...
   }
 
+  /**
+   * Determine the list of active delegates for the given round.
+   * @param  {Number} maxDelegates
+   * @return {Array}
+   */
+  determineActiveDelegates (maxDelegates) {
+    const mapWallets = wallets => wallets.map(wallet => ({
+      publicKey: wallet.vote || wallet.publicKey,
```
vasild: What is the point of having a logic like "use the public key of the delegate the wallet has voted for (wallet.vote), or if that is not set, then use wallet's own public key"? Looks wrong to me.

Also, below we filter the array to leave only entries with `wallet.vote` evaluating to true, this means the expression is identical to just `publicKey: wallet.vote`.

packages/core-database/lib/wallet-manager.js
```diff
@@ -376,6 +376,41 @@ module.exports = class WalletManager {
...
+      balance: wallet.balance ? wallet.balance.toFixed() : 0
```
vasild: `toFixed()` will return String. So, to be consistent, maybe replace `0` with `"0"` or use `Number(wallet.balance.toFixed())`

packages/core-database/lib/wallet-manager.js
```diff
@@ -376,6 +376,41 @@ module.exports = class WalletManager {
...
+    }))
+
+    let delegates = mapWallets(this.allByPublicKey().filter(wallet => !!wallet.vote))
```
vasild: Maybe I miss some edge case, but I think the use of `!!` here is just an obfuscation. I think the following two are identical:

```
.filter(wallet => !!wallet.vote))
```
and
```
.filter(wallet => wallet.vote))
```

packages/core-database/lib/wallet-manager.js
```diff
@@ -376,6 +376,41 @@ module.exports = class WalletManager {
...
+    if (delegates.length < maxDelegates) {
+      const publicKeys = delegates.map(delegate => delegate.publicKey)
+
+      let placeholders = this.allByUsername()
+
+      if (publicKeys.length) {
+        placeholders = placeholders.filter(wallet => publicKeys.includes(wallet.publicKey))
+      }
+
+      placeholders = placeholders.sort((a, b) => a.publicKey - b.publicKey)
```
vasild: I presume the `publicKey` property is a String (not a number).
```
> 'abc' - 'xyz'
NaN
```

packages/core-database/lib/wallet-manager.js
```diff
@@ -376,6 +376,41 @@ module.exports = class WalletManager {
...
+
+      delegates = delegates.concat(mapWallets(placeholders))
+    }
+
+    return delegates
+      .sort((a, b) => b.balance - a.balance) // TODO: improve sorting here to avoid collision of equal vote balances
```
vasild: Shouldn't this use the `voteBalance` property instead of `balance`? It looks like this is sorting according to delegate's own balance, not according to the sum of all voters' balances.

packages/core-database/lib/wallet-manager.js
faustbrian: Wallets that get mapped if they are a filler don't have votes themselves so they only have their `publicKey` property.

packages/core-database-postgres/lib/connection.js
```diff
@@ -203,36 +203,19 @@ module.exports = class PostgresConnection extends ConnectionInterface {
...
-
-      const fillerWallets = chosen.length
-        ? await this.db.rounds.placeholdersWithout(maxDelegates - data.length, chosen)
-        : await this.db.rounds.placeholders(maxDelegates - data.length)
-
-      data = data.concat(fillerWallets)
-    }
-
-    // logger.info(`got ${data.length} voted delegates`)
     const round = Math.floor((height - 1) / maxDelegates) + 1
```
vasild: Shouldn't we be worried about `Math.floor(110 / 1.1)` being `99`?
Maybe something like: `bignum(height).minus(1).dividedBy(maxDelegates).plus(1).toNumber()`? (yuck!)

faustbrian:
@supaiku0 Testing a different approach here https://github.com/ArkEcosystem/core/tree/memory-delegates-2 which is a bit cleaner.

faustbrian:
Closing in favour of https://github.com/ArkEcosystem/core/pull/1134

PR #1131, opened 2018-10-16
feat: get transactions from mem pool ordered by fee

vasild:
## Types of changes

## Checklist


vasild:
Beware:

before this patch:
```
      ✓ multiple additions and retrievals (646ms)
      ✓ delete + add after sync (173ms)
```

after this patch:
```
      ✓ multiple additions and retrievals (5779ms)
      ✓ delete + add after sync (1306ms)
```

vasild:
This implements the first part of https://github.com/ArkEcosystem/core/issues/1061

packages/core-transaction-pool-mem/lib/mem-pool-transaction.js
```diff
@@ -0,0 +1,36 @@
...
+    return this.transaction_
+  }
+
+  get sequence () {
+    assert.notEqual(this.sequence_, undefined)
+    return this.sequence_
+  }
+
+  set sequence (seq) {
+    assert.strictEqual(this.sequence_, undefined)
```
faustbrian:
```suggestion
    assert.strictEqual(this._sequence, undefined)
```

packages/core-transaction-pool-mem/lib/mem-pool-transaction.js
```diff
@@ -0,0 +1,36 @@
...
+    }
+  }
+
+  get transaction () {
+    return this.transaction_
+  }
+
+  get sequence () {
+    assert.notEqual(this.sequence_, undefined)
+    return this.sequence_
```
faustbrian:
```suggestion
    return this._sequence
```

packages/core-transaction-pool-mem/lib/mem-pool-transaction.js
```diff
@@ -0,0 +1,36 @@
...
+    this.transaction_ = transaction
+
+    if (sequence !== undefined) {
+      assert(sequence instanceof Number)
+      this.sequence_ = sequence
+    }
+  }
+
+  get transaction () {
+    return this.transaction_
```
faustbrian:
```suggestion
    return this._transaction
```

packages/core-transaction-pool-mem/lib/mem-pool-transaction.js
```diff
@@ -0,0 +1,36 @@
...
+ * A normal transaction + a sequence number used to order by insertion time.
+ */
+class MemPoolTransaction {
+  constructor (transaction, sequence) {
+    assert(transaction instanceof Transaction)
+    this.transaction_ = transaction
+
+    if (sequence !== undefined) {
+      assert(sequence instanceof Number)
+      this.sequence_ = sequence
```
faustbrian:
```suggestion
      this._sequence = sequence
```

packages/core-transaction-pool-mem/lib/mem-pool-transaction.js
```diff
@@ -0,0 +1,36 @@
...
+const { Transaction } = require('@arkecosystem/crypto').models
+
+/**
+ * A mem pool transaction.
+ * A normal transaction + a sequence number used to order by insertion time.
+ */
+class MemPoolTransaction {
+  constructor (transaction, sequence) {
+    assert(transaction instanceof Transaction)
+    this.transaction_ = transaction
```
faustbrian:
```suggestion
    this._transaction = transaction
```

packages/core-transaction-pool-mem/lib/mem-pool-transaction.js
```diff
@@ -0,0 +1,36 @@
...
+  }
+
+  get sequence () {
+    assert.notEqual(this.sequence_, undefined)
+    return this.sequence_
+  }
+
+  set sequence (seq) {
+    assert.strictEqual(this.sequence_, undefined)
+    this.sequence_ = seq
```
faustbrian:
```suggestion
    this._sequence = seq
```

packages/core-transaction-pool-mem/lib/mem-pool-transaction.js
```diff
@@ -0,0 +1,36 @@
+'use strict'
+
+const assert = require('assert')
+const { Transaction } = require('@arkecosystem/crypto').models
+
+/**
+ * A mem pool transaction.
+ * A normal transaction + a sequence number used to order by insertion time.
+ */
+class MemPoolTransaction {
```
faustbrian:
```suggestion
module.exports = class MemPoolTransaction {
```

vasild:
> Beware:
> 
> before this patch:
> 
> ```
>       ✓ multiple additions and retrievals (646ms)
>       ✓ delete + add after sync (173ms)
> ```
> 
> after this patch:
> 
> ```
>       ✓ multiple additions and retrievals (5779ms)
>       ✓ delete + add after sync (1306ms)
> ```

Just tried to sort on demand every time a sorted result is needed instead of storing the data sorted. We need two orders: by fee (to get highest fee transactions) and by expiration (to expire old transactions). This would make sense if we insert a lot (because then we do not do any sorting during insert) and read a little (because reading is expensive as it involves sorting):

```
      ✓ multiple additions and retrievals (29853ms)
      ✓ delete + add after sync (233ms)
```

Not good either.

packages/core-transaction-pool-mem/lib/mem-pool-transaction.js
vasild: Done

packages/core-transaction-pool-mem/lib/mem-pool-transaction.js
vasild: Done

packages/core-transaction-pool-mem/lib/mem-pool-transaction.js
vasild: Done

packages/core-transaction-pool-mem/lib/mem-pool-transaction.js
vasild: Done

packages/core-transaction-pool-mem/lib/mem-pool-transaction.js
vasild: Done

packages/core-transaction-pool-mem/lib/mem-pool-transaction.js
vasild: Done

vasild:
> > Beware:
> > before this patch:
> > ```
> >       ✓ multiple additions and retrievals (646ms)
> >       ✓ delete + add after sync (173ms)
> > ```
> > after this patch:
> > ```
> >       ✓ multiple additions and retrievals (5779ms)
> >       ✓ delete + add after sync (1306ms)
> > ```
> 
> Just tried to sort on demand every time a sorted result is needed instead of storing the data sorted. We need two orders: by fee (to get highest fee transactions) and by expiration (to expire old transactions). This would make sense if we insert a lot (because then we do not do any sorting during insert) and read a little (because reading is expensive as it involves sorting):
> 
> ```
>       ✓ multiple additions and retrievals (29853ms)
>       ✓ delete + add after sync (233ms)
> ```
> 
> Not good either.

Using http://www.collectionsjs.com/sorted-set:
```
      ✓ multiple additions and retrievals (21493ms)
      ✓ delete + add after sync (177ms)
```

vasild:
> @vasild what dataset size are you using that you end up with 21 seconds, millions of records or is the 1024 from the tests?

It is 1024 from the tests (unmodified), but this does not matter so much. What matters is the magnitude of the change: before patch <1s, with array, >5s. Simply not acceptable. Also, I would expect the sorted-map solution to be faster than the array, not 4x slower. I have not fully analyzed this, just sharing intermediate results.

vasild:
Added one more test and changed the sorting to happen on demand / lazily.

Looks on par with the performance as of before order by fee (before this patch):

```
before:
      ✓ multiple additions and retrievals (635ms)
      ✓ delete + add after sync (173ms)
      ✓ add many then get first few (821ms)

after:
      ✓ multiple additions and retrievals (717ms)
      ✓ delete + add after sync (181ms)
      ✓ add many then get first few (865ms)
```

packages/core-transaction-pool-mem/package.json
```diff
@@ -28,7 +28,8 @@
     "better-sqlite3": "^5.0.1",
     "collections": "^5.1.5",
     "delay": "^4.1.0",
-    "fs-extra": "^7.0.0"
+    "fs-extra": "^7.0.0",
+    "random-seed": "^0.3.0"
```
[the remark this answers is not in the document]
vasild: Right, good catch! Moved. Thanks!

PR #1134, opened 2018-10-17
refactor: build delegate list from memory

### LINKED ISSUE #850: Calculate new round active delegate list from memory and not from database

faustbrian:
## Proposed changes

Resolves #850

## Types of changes

## Checklist

- [ ] I have added tests that prove my fix is effective


packages/core-database/lib/wallet-manager.js
```diff
@@ -349,39 +350,91 @@ module.exports = class WalletManager {
...
+   * @param  {Number} maxDelegates
+   * @return {Array}
+   */
+  allActiveDelegates (maxDelegates) {
+    let delegates = this.allByUsername()
+
+    // NOTE: At the launch of the blockchain we may not have enough delegates.
+    // In order to have enough forging delegates we complete the list in a
+    // deterministic way (alphabetical order of publicKey).
+    if (delegates.length < maxDelegates) {
```
fix: it should throw if there is not enough delegates (ie genesisBlock should register at least maxDelegates)

faustbrian: Should we shutdown the node in case there are not enough delegates? Doesn't seem to make sense to keep it running without enough delegates.

packages/core-database/lib/wallet-manager.js
```diff
@@ -349,39 +350,91 @@ module.exports = class WalletManager {
...
+      }
+
+      fillerWallets = orderBy(fillerWallets, ['publicKey'], ['asc'])
+
+      delegates = delegates.concat(fillerWallets.slice(0, maxDelegates - delegates.length))
+    }
+
+    return delegates.sort((a, b) => {
+      const aBalance = +a.voteBalance.toFixed()
+      const bBalance = +b.voteBalance.toFixed()
```
fix: mmm to fixed gives you a string, this aBalance - bBalance is alphabetical order

fix: + is supposed to cast to integer then, true

faustbrian: The unary plus (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Arithmetic_Operators#Unary_plus_()) will turn the `toFixed` result into an integer.

packages/core-database/lib/wallet-manager.js
```diff
@@ -349,39 +350,91 @@ module.exports = class WalletManager {
...
+      if (publicKeys.length) {
+        fillerWallets = fillerWallets.filter(wallet => !publicKeys.includes(wallet.publicKey))
+      }
+
+      fillerWallets = orderBy(fillerWallets, ['publicKey'], ['asc'])
+
+      delegates = delegates.concat(fillerWallets.slice(0, maxDelegates - delegates.length))
+    }
+
+    return delegates.sort((a, b) => {
```
fix: before doing that i would make a function that recalculate all vote balance:
```
this.allByPublicKeys().forEach((w)=> {
  if(w.vote) this.byPublicKey[w.vote].voteBalance += w.balance //missing initialisation
})
```

packages/core-database/lib/wallet-manager.js
```diff
@@ -398,4 +451,25 @@ module.exports = class WalletManager {
     return wallet.balance.isZero() && !wallet.secondPublicKey && !wallet.multisignature && !wallet.username
   }
 
+  /**
+   * Update the vote balance of the delegate the vote is for.
+   * @param  {Object} sender
+   * @param  {Transaction} transaction
+   * @return {void}
+   */
+  __updateVoteBalance (sender, transaction) {
```
fix: that is quite complex operation.
- this is different if the tx is applied or reverted
- the votes change for sender AND receiver

faustbrian:
Syncing on devnet seems to work without any issues now but mainnet is still having some issues so will continue to look into it.

packages/core-database-postgres/lib/connection.js
```diff
@@ -185,7 +183,10 @@ module.exports = class PostgresConnection extends ConnectionInterface {
       currentSeed = crypto.createHash('sha256').update(currentSeed).digest()
     }
 
-    this.activeDelegates = data
+    this.activeDelegates = data.map(delegate => {
+      delegate.round = +round
```
fix: what is the rationale for this?

packages/core-database/lib/wallet-manager.js
```diff
@@ -184,18 +185,61 @@ module.exports = class WalletManager {
...
+      const bBalance = +b.voteBalance.toFixed()
+
+      if (aBalance === bBalance) {
+        // logger.warn(`Delegate ${a.username} (${a.publicKey}) and ${b.username} (${b.publicKey}) have a matching vote balance of ${a.voteBalance.dividedBy(Math.pow(10, 8)).toLocaleString()}.`)
+
+        if (a.publicKey === b.publicKey) {
+          logger.error(`The balance and public key of both delegates are identical! Delegate "${a.username}" appears twice in the list.`)
+        }
+
+        return a.publicKey.localeCompare(b.publicKey, 'en-US-u-kf-lower')
```
fix: why this 'en-US-u-kf-lower' ? 'en' should be ok as there is only '0123456789abcdef' as letters

packages/core-database-postgres/lib/connection.js
faustbrian: There is a comparison somewhere that was always `false` because it compared `string === integer`, this fixes the comparison to `integer === integer`.

packages/core-database/lib/wallet-manager.js
```diff
@@ -184,18 +184,55 @@ module.exports = class WalletManager {
...
+
     Object.values(this.byPublicKey)
       .filter(voter => !!voter.vote)
       .forEach(voter => {
         const delegate = this.byPublicKey[voter.vote]
         delegate.voteBalance = delegate.voteBalance.plus(voter.balance)
       })
+
     Object.values(this.byUsername)
       .sort((a, b) => +(b.voteBalance.minus(a.voteBalance)).toFixed())
```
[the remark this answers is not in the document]
faustbrian: Just to make sure we are actually sorting by the numerical value instead of string that is returned by `toFixed`, and consistency.

packages/core-database-postgres/lib/connection.js
fix: i mean why adding delegate.round property?

fix: ah you mean you override the round property of delegates?

packages/core-database/lib/wallet-manager.js
faustbrian: Was mainly to make sure the string is treated as a lowercase string in case there ever is a mixed-case public key for whatever reason but changed it to `en`.

packages/core-database/lib/wallet-manager.js
```diff
@@ -184,18 +185,61 @@ module.exports = class WalletManager {
     Object.values(this.byAddress).map(wallet => (wallet.dirty = false))
   }
 
+  /**
+   * Load a list of all active delegates.
+   * @param  {Number} maxDelegates
+   * @return {Array}
+   */
+  activeDelegation (maxDelegates, height) {
```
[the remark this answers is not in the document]
faustbrian: changed

packages/core-database/lib/wallet-manager.js
```diff
@@ -189,13 +189,17 @@ module.exports = class WalletManager {
    * @return {void}
    */
   updateDelegates () {
-    Object.values(this.byUsername).forEach(delegate => (delegate.voteBalance = Bignum.ZERO))
+    Object
+      .values(this.byUsername)
+      .forEach(delegate => (delegate.voteBalance = Bignum.ZERO))
+
     Object.values(this.byPublicKey)
       .filter(voter => !!voter.vote)
```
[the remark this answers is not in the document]
faustbrian: changed

packages/core-database-postgres/lib/connection.js
faustbrian: @supaiku0 added it for here https://github.com/ArkEcosystem/core/blob/develop/packages/core-database-postgres/lib/connection.js#L167

supaiku0: @fix See https://github.com/ArkEcosystem/core/pull/1143 and https://github.com/ArkEcosystem/core/pull/1152

packages/core-database/lib/wallet-manager.js
```diff
@@ -183,20 +184,63 @@ module.exports = class WalletManager {
...
+
+    const { round } = roundCalculator.calculateRound(height, maxDelegates)
+    let delegates = this.allByUsername()
+
+    if (delegates.length < maxDelegates) {
+      throw new Error(`Expected to find ${maxDelegates} delegates but only found ${delegates.length}. This indicates an issue with the genesis block & delegates.`)
+    }
+
+    delegates = delegates.sort((a, b) => {
+      const aBalance = +a.voteBalance.toFixed()
```
fix: by the way what is the point of this? i guess keeping a.voteBalance is enough.
- they are numbers
- if there is a rounding issue, since they are all integers, it won't change the order

packages/core-database/lib/wallet-manager.js
```diff
@@ -183,20 +184,63 @@ module.exports = class WalletManager {
...
+      const bBalance = +b.voteBalance.toFixed()
+
+      if (aBalance === bBalance) {
+        // logger.warn(`Delegate ${a.username} (${a.publicKey}) and ${b.username} (${b.publicKey}) have a matching vote balance of ${a.voteBalance.dividedBy(Math.pow(10, 8)).toLocaleString()}.`)
+
+        if (a.publicKey === b.publicKey) {
+          logger.error(`The balance and public key of both delegates are identical! Delegate "${a.username}" appears twice in the list.`)
```
fix: i think we should throw here, because it means something even wronger happened

packages/core-database/lib/wallet-manager.js
```diff
@@ -183,20 +184,63 @@ module.exports = class WalletManager {
...
+    if (delegates.length < maxDelegates) {
+      throw new Error(`Expected to find ${maxDelegates} delegates but only found ${delegates.length}. This indicates an issue with the genesis block & delegates.`)
+    }
+
+    delegates = delegates.sort((a, b) => {
+      const aBalance = +a.voteBalance.toFixed()
+      const bBalance = +b.voteBalance.toFixed()
+
+      if (aBalance === bBalance) {
+        // logger.warn(`Delegate ${a.username} (${a.publicKey}) and ${b.username} (${b.publicKey}) have a matching vote balance of ${a.voteBalance.dividedBy(Math.pow(10, 8)).toLocaleString()}.`)
```
fix: we can keep this imo

packages/core-database/lib/wallet-manager.js
faustbrian: `a.voteBalance` is a bignumber object inside a wallet object.

supaiku0: We could use https://mikemcl.github.io/bignumber.js/#cmp

fix: ah true that, i should keep that in mind now. Is there some kind of comparison stuff in bignumber library, sounds weird to cast down to normal number in order to compare again

fix: http://mikemcl.github.io/bignumber.js/#cmp

fix:
ok sounds good me. If it rebuilds the first 1000s blocks on mainnet/devnet, we can merge

faustbrian:
@supaiku0 you still had a thing you would like to add in this PR right?

packages/core-database/lib/wallet-manager.js
supaiku0: I changed it because it issues warnings for all standby delegates as well unless we make it more strict.

supaiku0:
Will open a separate PR.

PR #1138, opened 2018-10-17
Change API client to use the default `core-api` port

j-a-m-l:
## Proposed changes
On V2, instead of using the `core-p2p` port (4002) provided by the `peers` endpoint, use the default port of `core-api` (4003).

## Types of changes


fix:
mmm not sure to understand. But i think so:
The api end point is or 2 purposes:
- on p2p layer, so nodes discover each others
- client api so lite client can communicate with

They serve 2 purposes. so to p2p layer must NOT be touched.

packages/client/lib/index.js
```diff
@@ -33,19 +32,28 @@ module.exports = class ApiClient {
...
         let responsePeers = []
         if (version === 1 && data.success && data.peers) {
           responsePeers = data.peers
         } else if (version === 2 && data.data.length) {
-          responsePeers = data.data
+          // Use the default port used by `core-api`
+          responsePeers = data.data.map(peer => {
+            return {
+              ...peer,
+              port: 4003
```
fix: that should be taken from config instead of hardcoded

j-a-m-l: I could change that a little, but the API client can be used on a browser, so it shouldn't depend on other packages to be lighter.

j-a-m-l:
This PR doesn't change anything from the `core-p2p` layer, but, since the client needs to connect to the `core-api` endpoints, then, it needs to infer or fetch the `core-api` port.

The `peers` endpoint returns only 1 kind of port (`core-p2p`). So, when using it to fetch peers, the receive response includes the peers and their `core-p2p` port, but not the `core-api` port.

That means, that users, can't access the API with `core-p2p` port. To get the `core-api` port, it's necessary to:
  a) assume that all of those peers are using the default port (4003)
  b) perform another to get the peer configuration at `/api/node/configuration`.

But, if that node doesn't use the default `core-api` port, then it's impossible to reach that endpoint. so the second approach, although is the correct, is useless.

This PR was a quick fix until implementing a better solution.

Ideally, we would have 2 kind of `/peers` enpoint:
 1. Only for `core-p2p` with the same data that is returning now
 2. An extended version, to be used by `core-api` alone, with part of the configuration of each peer.

An alternative would be adding a parameter to the current endpoint, like `/peers?configuration=all` to return all the configuration, the configuration of some packages (`/peers?configuration=core-api`) or something like that.

j-a-m-l:
@fix what do you think about my previous message? I could work on it if you think that any of those options is OK or you propose a better idea.

faustbrian:
You should simply query http://199.247.25.197:4002/config to perform another check if the API is enabled and to find out which port it uses.

j-a-m-l:
Thanks. That removes the 

About the discovery feature, its main purpose is avoiding that every user has to introduce the peers by hand, relying on an IP that could be offline at any moment, or that a lot of users query and depend on the same peer because they have found it on the internet as the only example.

With that feature users don't need to update their codebase if the peers that they are using change. That's essential to make things easier for our users. They shouldn't re-implement this feature for each project.

faustbrian:
@alexbarnsley do you need this on the desktop wallet? If so, this should probably be finished asap.

j-a-m-l:
If the plan is creating a new kind of client that deal with peer discovery among other features, then this could be enough until creating that new project.

faustbrian:
I would just finish this PR now and then we revisit this later on.

PR #1140, opened 2018-10-17
Update dispatch.test.js

kalgoop:
Fixed a mistake that could have led to misinterpretation of the Test Case

## Types of changes

## Checklist


faustbrian:
Thanks for the PR but the meaning of the description is still the same and I don't see how someone would misinterpret it.

kalgoop:
"fails when the dispatch method **is not called** with the argument "
The phrase "is not called" can be misinterpreted as method isn't even called.
But here, we mean that the **dispatch method is called,** but **without the argument**

PR #1141, opened 2018-10-17
Added warning

kalgoop:
cosole.log - Warning in case there are less than 0.1 DARK to cover fees
Since dynamic fees is planned in v2, it is just a warning and doesn't fail the transaction

## Types of changes

## Checklist


faustbrian:
The wallets that get generated by the CLI will always have enough funds if the previous condition doesn't fail.

PR #1142, opened 2018-10-17
Updated for v2

kalgoop:
'delay' is changed to 'latency' in v2

## Types of changes

## Checklist

### DIFF
--- a/packages/client/lib/utils/sort-peers.js
+++ b/packages/client/lib/utils/sort-peers.js
@@ -4,5 +4,10 @@ const orderBy = require('lodash.orderby')
  * Sorts the peers, in place, by block height and delay
  */
 module.exports = peers => {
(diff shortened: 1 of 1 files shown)


alexbarnsley:
Thanks for your change. This should be split up into 2 methods though, as currently the client is designed to work with both v1 and v2 networks, so it should check which version to see which column needs to be sorted by.

kalgoop:
> 
> 
> Thanks for your change. This should be split up into 2 methods though, as currently the client is designed to work with both v1 and v2 networks, so it should check which version to see which column needs to be sorted by.

Thanks for the advice. Updated.
It checks if peers object has a property of 'delay' or 'latency', and accordingly uses orderBy operation.

fix:
this might be breaking with v1 protocol p2p layer.

kalgoop:
@fix can you elaborate how it may break with v1 p2p?
I feel this will only improve situation from previous version.

PR #1144, opened 2018-10-17
Remove the `--watch` option of the `test:debugger` script

j-a-m-l:
## Proposed changes
I'm having problems when trying to use it with that option.

@supaiku0 maybe this was the reason that you have problems in the past. Could you try it, please?

## Types of changes


supaiku0:
Yeah, works fine. Should we remove `--watch` from all other packages as well?

j-a-m-l:
Yes, I'll do it this evening, after fixing other thing.

PR #1154, opened 2018-10-18
Improved the Implementation

kalgoop:
Rewrote code for sorting

## Types of changes

## Checklist


packages/crypto/lib/utils/sort-transactions.js
```diff
@@ -7,22 +7,10 @@ module.exports = (transactions) => {
...
-    if (a.id < b.id) {
-      return -1
-    }
-
-    if (a.id > b.id) {
-      return 1
-    }
-
-    return 0
+    return a.id - b.id
```
fix: this does not work as intended since they are not integer

PR #1157, opened 2018-10-19
Optimize vote balance calculation

### LINKED ISSUE #1010: Useless database request for each block save
auditing pg sql interaction there is for each block save the following useless request, decreasing performance
```
2018-09-17 04:48:29 EDT [29941-36] node@ark_mainnet LOG:  statement: BEGIN
2018-09-17 04:48:29 EDT [29941-37] node@ark_mainnet LOG:  statement: COMMIT
2018-09-17 04:48:29 EDT [29941-38] node@ark_mainnet LOG:  statement: SELECT * FROM rounds WHERE round = 49443 ORDER BY balance DESC, public_key ASC
```

as the result wont change during a round, this should be cached

supaiku0:
## Proposed changes

This PR changes how vote balances are calculated. Instead of calculating them from scratch all the time, use a "delta update" approach.

The most relevant changes can be found here:
- `applyTransaction/revertTransaction`
- `applyBlock/revertBlock`

Additionally I changed the order of `buildVotes` and `buildDelegates` during SPV. By doing so the
delegate vote balances can be easily validated.

~~Marked as WIP as my node just got stuck around height 8416 with generator mismatch (sync from 0 on devnet), but recovered... and then got stuck a few more times at later heights. I still need to investigate it.~~

Now also resolves #1010 

## Types of changes

## Checklist

### DIFF
--- a/packages/core-database-postgres/lib/connection.js
+++ b/packages/core-database-postgres/lib/connection.js
@@ -158,9 +158,10 @@ module.exports = class PostgresConnection extends ConnectionInterface {
   /**
    * Get the top 51 delegates.
    * @param  {Number} height
+   * @param  {Array} delegates
    * @return {Array}
    */
-  async getActiveDelegates (height) {
+  async getActiveDelegates (height, delegates) {
     const maxDelegates = config.getConstants(height).activeDelegates
     const round = Math.floor((height - 1) / maxDelegates) + 1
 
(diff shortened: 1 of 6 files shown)


faustbrian:
The issue that causes the hiccups seems to be that some wallets have 1 ark less in this branch then they do on develop.

See http://www.mergely.com/IxrRHszK/ - left is this branch, right is develop.

faustbrian:
Fully synced mainnet from 0, should be good 👍

PR #1186, opened 2018-10-23
chore: adhere to naming and readme conventions

dated:
## Proposed changes

 - Changes `ARK` to `Ark` where appropriate
 - Changes readme structure and replaces any reference to imgur with local images

## Types of changes

- [x] Docs (documentation only changes)

## Checklist


dated:
> Copying the banner.png into every directory is unnecessary duplication as the banner will never change. You should simply put the banner into the root of the repository and link to that.

Changed that back for all packages except `client` and `crypto` which have custom banners.

alexbarnsley:
It's worth noting that the banners may need to be within packages for when they are published to npm. Although that being said, it looks like it doesn't work for the same directory either:

https://www.npmjs.com/package/@arkecosystem/crypto

Could it be worth doing the full URL to the image?

dated:
My personal preference would be to add `?sanitize=true` if that works, but I'll wait for your input.

faustbrian:
Do the `?sanitize=true`, looks the cleanest.

alexbarnsley:
I agree, `?sanitize=true` looks like a good way of doing it 👌

dated:
I've looked at this again and I think the sanitize method does not work after all. The only fix seems to be to use absolute URLs. This issue affects all relative paths, so also the `All Contributors` and `MIT` links are broken on the npm website.

dated:
Yep, doesn't work. I'll change the links and images to absolute paths later today.

PR #1199, opened 2018-10-24
test(mempool): ignore index.js and defaults.js in coverage report

vasild:
index.js and defaults.js do not contain any real code, so skip them from
the output of `yarn test:coverage`

## Types of changes

## Checklist


vasild:
There you go. Patch generated with:

```
sed -i '' -E 's|("test:coverage": "cross-env ARK_ENV=test jest --coverage)|\1 --coveragePathIgnorePatterns='"'"'/\(defaults.js\|index.js)$'"'"'|' `git grep -l test:coverage`
```

Maybe there is a smarter way to define this command once at the top level and reuse it, instead of repeat it in every package.json?

PR #1221, opened 2018-10-26
refactor: reduce number of loops when reading/writing transaction bytes

supaiku0:
## Proposed changes

Writing transaction bytes in `Block.serializeFull` requires three loops and two loops when reading.
Both are replaced with a single loop which achieves the same.

## Types of changes

## Checklist

### DIFF
--- a/packages/crypto/lib/models/block.js
+++ b/packages/crypto/lib/models/block.js
@@ -353,15 +353,19 @@ module.exports = class Block {
     block.transactions = []
     if (hexString.length === transactionOffset * 2) return block
 
-    for (let i = 0; i < block.numberOfTransactions; i++) {
-      block.transactions.push(buf.readUint32(transactionOffset))
-      transactionOffset += 4
-    }
+    // A serialized block stores transactions like this:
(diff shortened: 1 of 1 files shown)


packages/crypto/lib/models/block.js
```diff
@@ -374,15 +378,20 @@ module.exports = class Block {
    * @static
    */
   static serializeFull (block) {
+    const serializedBlock = Block.serialize(block, true);
+    const transactions = block.transactions
+
     const buf = new ByteBuffer(1024, true)
```
faustbrian: The size of `1024` seems to cause issues on mainnet because of the size of the genesis block while it doesn't happen on testnet or devnet.

`new ByteBuffer(serializedBlock.length + transactions.length * 4, true)` seems to resolve the mainnet issue.

faustbrian: Results of logging `serializedBlock.length + transactions.length * 4` for the genesis block.

```
mainnet = 6156
devnet = 395
testnet = 799
```

PR #1226, opened 2018-10-28
refactor(core-http-utils): pass name to whitelist plugin

faustbrian:
## Proposed changes

Makes it possible to pass a name to the whitelist plugin as it was always showing `Public API`.

## Types of changes

## Checklist


packages/core-p2p/lib/server/index.js
```diff
@@ -20,7 +20,8 @@ module.exports = async (p2p, config) => {
   await server.register({
     plugin: require('./plugins/accept-request'),
     options: {
-      whitelist: config.whitelist
+      whitelist: config.whitelist,
+      name: 'P2P API'
```
supaiku0: The name here is useless since it's a different plugin.

PR #1256, opened 2018-10-31
Number validation

dated:
## Proposed changes

#1251 added stricter validation on numerical inputs in the core-api package. This PR fixes some of those, reasonal below, and adds the same kind of validation to the schemas of the remaining packages.

## Types of changes

## Checklist

## Further comments

`totalAmount` & `payloadLength`: both can have the value `0`
`balance`: the genesis wallet has a negative value

### DIFF
--- a/packages/core-api/lib/versions/2/schema/wallets.js
+++ b/packages/core-api/lib/versions/2/schema/wallets.js
@@ -16,7 +16,7 @@ exports.index = {
       secondPublicKey: Joi.string().hex().length(66),
       vote: Joi.string().hex().length(66),
       username: Joi.string(),
-      balance: Joi.number().integer().min(0),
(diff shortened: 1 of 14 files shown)


packages/core-api/lib/versions/2/schema/wallets.js
```diff
@@ -16,7 +16,7 @@ exports.index = {
       secondPublicKey: Joi.string().hex().length(66),
       vote: Joi.string().hex().length(66),
       username: Joi.string(),
-      balance: Joi.number().integer().min(0),
+      balance: Joi.number().integer(),
```
faustbrian: Why remove the `min` here, because of the negative genesis balance?

dated: Exactly.

PR #1259, opened 2018-11-01
fix: Uppercase publicKey/signature fields when read from sqlite

### LINKED ISSUE #1252: Handler:canApply can fail a transaction because of a case-sensitive check on publicKey
**Describe the bug**
crypto/lib/handlers/transactions/handler.js line 27.
There is  a comparison :
`applicable = (transaction.senderPublicKey === wallet.publicKey) && enoughBalance`

This check, will fail if either transaction.senderPublicKey is UPPERCASE and wallet.publicKey is lowercase or vice-versa, even if they're equal.
I ran into a scenario on my relay, where somehow the in-memory wallets found in Pool-Wallet-Manager(byPublicKey) became keyed by uppercase versions of the publicKey. So, on a subsequent transaction send, it would fail.
 This might have implications elsewhere where we do such comparisons.

**To Reproduce**
I'm not sure how the in-memory wallets got keyed in uppercase, but I'd start in the transaction-pool and dig there. At any case, that shouldn't cause the transaction to fail.

**Expected behavior**
The two properties are equal in meaning, but not in casing. That shouldn't be a deal breaker..

**Screenshots**
uppercase_by_publickey

paroxysm:
## Proposed changes
Fixes #1252 

## Types of changes

## Checklist

- [x] I  have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/packages/crypto/lib/handlers/transactions/handler.js
+++ b/packages/crypto/lib/handlers/transactions/handler.js
@@ -21,7 +21,7 @@ module.exports = class Handler {
     } else {
       const balance = +(wallet.balance.minus(transaction.amount).minus(transaction.fee)).toFixed()
       const enoughBalance = balance >= 0
-      applicable = (transaction.senderPublicKey === wallet.publicKey) && enoughBalance
+      applicable = (transaction.senderPublicKey.toLowerCase() === wallet.publicKey.toLowerCase()) && enoughBalance
 
       // TODO: this can blow up if 2nd phrase and other transactions are in the wrong order
       applicable = applicable && (!wallet.secondPublicKey || crypto.verifySecondSignature(transaction, wallet.secondPublicKey, configManager.config)) // eslint-disable-line max-len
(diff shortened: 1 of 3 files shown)


packages/crypto/lib/handlers/transactions/handler.js
```diff
@@ -21,7 +21,7 @@ module.exports = class Handler {
     } else {
       const balance = +(wallet.balance.minus(transaction.amount).minus(transaction.fee)).toFixed()
       const enoughBalance = balance >= 0
-      applicable = (transaction.senderPublicKey === wallet.publicKey) && enoughBalance
+      applicable = (transaction.senderPublicKey.toLowerCase() === wallet.publicKey.toLowerCase()) && enoughBalance
```
supaiku0: Not really a fan of calling`.toLowerCase()`everywhere... we need to ensure that it is normalized else it is a bug. Is this even still necessary with the SQLite change?

paroxysm: SQLite change does fix the source of the uppercase fields. We should be able to get by for now, albeit that code will still be vulnerable to that defect, hence why I put it in there for good measure.

faustbrian: This could be an option on the transaction and wallet models to be 100% the keys are always lower-case.

```js
get senderPublicKey () {
    return this._senderPublicKey
}

set senderPublicKey (value) {
    this._senderPublicKey = value.toLowerCase()
}
```

paroxysm: For transaction model, it could work, although there are places where we reference a transaction-like object(such as tests), w/o instantiating a Transaction object.

The Wallet model doesn't even have a constructor that accepts parameters like how it's used in some places.

Look at the test for the handler.js file in this changeset, it creates Transaction-like & Wallet-like objects but never actually constructs them. I have no idea if the same happens in non-test code such that it would render getter/setter unused.

PR #1267, opened 2018-11-01
refactor(core-transaction-pool): return error if transaction fails verification

zillionn:
## Proposed changes
I'm using PHP client/crypto to send transactions and when I set some long `vendorField`, `php-crypto` is verifying the transaction but the `ark-core` will not and the API will not even return any error.

The max `vendorField` I'm able to send with is 63 instead of 64.

Before:
`Sending 3 transactions...
array(2) {
  ["data"]=>
  array(4) {
    ["accept"]=>
    array(0) {
    }
    ["excess"]=>
    array(0) {
    }
    ["invalid"]=>
    array(0) {
    }
    ["broadcast"]=>
    array(0) {
    }
  }
  ["errors"]=>
  NULL
}`

After the patch:
`Sending 3 transactions...
Error sending transactions!
array(2) {
  ["data"]=>
  array(4) {
    ["accept"]=>
    array(0) {
    }
    ["excess"]=>
    array(0) {
    }
    ["invalid"]=>
    array(3) {
      [0]=>
      string(64) "89726b20f053981e756193c782acc14c25a9f76fb1901fbb9b12d345e06b3b74"
      [1]=>
      string(64) "f8493f9807c28ae4d06c1dfa361bbe11ab2bacd118f0550690f3cce0ec09536d"
      [2]=>
      string(64) "17d7fc629565353dc0e83af507378392b2e9bd07dfce56b19c0bfadb58b7a2e5"
    }
    ["broadcast"]=>
    array(0) {
    }
  }
  ["errors"]=>
  array(3) {
    ["89726b20f053981e756193c782acc14c25a9f76fb1901fbb9b12d345e06b3b74"]=>
    array(1) {
      [0]=>
      string(49) "Transaction didn't pass the verification process."
    }
    ["f8493f9807c28ae4d06c1dfa361bbe11ab2bacd118f0550690f3cce0ec09536d"]=>
    array(1) {
      [0]=>
      string(49) "Transaction didn't pass the verification process."
    }
    ["17d7fc629565353dc0e83af507378392b2e9bd07dfce56b19c0bfadb58b7a2e5"]=>
    array(1) {
      [0]=>
      string(49) "Transaction didn't pass the verification process."
    }
  }
}`

## Types of changes
I'm returning an error if the transaction didn't pass the verification.

## Checklist


zillionn:
I have no idea why the tests fail.

faustbrian:
If the issue you are facing is `The max vendorField I'm able to send with is 63 instead of 64.` then it seems to be an issue with the crypto in javascript as 64 is the correct value.

vasild:
@zillionn give me the exact steps to reproduce this, together with a description of observed and expected behavior.

zillionn:
I'm sending a trx with 64 chars `vendorField`:
`array(1) {
  [0]=>
  array(11) {
    ["amount"]=>
    int(1)
    ["fee"]=>
    int(300)
    ["id"]=>
    string(64) "ec502438443ddae2f176954f91c7f80fe04111ae4c26780d3eacccee5d5225cc"
    ["network"]=>
    int(30)
    ["recipientId"]=>
    string(34) "D8wgoDnRNAW1tnjeTy5a39zMp8Wa3sJSFZ"
    ["senderPublicKey"]=>
    string(66) "03318a51e35b8fa96108a2643b17819e4f6d5c7c1df30fbee2102bb18151e09dc4"
    ["signature"]=>
    string(142) "3045022100eec3f0c468e996d3259f5e099966d543110819970b45738060e3a9ebb86d2f66022053df6719941a43b43b4f3408bac4c53e292c7d8d85ffc41a37f195a5ffb471ec"
    ["timestamp"]=>
    int(51047908)
    ["type"]=>
    int(0)
    ["vendorField"]=>
    string(64) "test test test test test test test test test test test test|1234"
    ["version"]=>
    int(1)
  }
}`

The response is empty because the `guard.js` does not return any errors if `!trx.verified` [L132](https://github.com/ArkEcosystem/core/blob/056b70a939196622080839995278df6ff3b1d481/packages/core-transaction-pool/lib/guard.js#L132):
`array(2) {
  ["data"]=>
  array(4) {
    ["accept"]=>
    array(0) {
    }
    ["excess"]=>
    array(0) {
    }
    ["invalid"]=>
    array(0) {
    }
    ["broadcast"]=>
    array(0) {
    }
  }
  ["errors"]=>
  NULL
}`

I just added an `else` and return an error message.

zillionn:
@faustbrian two issues:
1. Can't send trx with `vendorField > 63` chars
2. If `vendorField > 63` no error is the response

zillionn:
It's `< 64` everywhere:
[https://github.com/ArkEcosystem/core/blob/056b70a939196622080839995278df6ff3b1d481/packages/crypto/lib/crypto/crypto.js#L112](https://github.com/ArkEcosystem/core/blob/056b70a939196622080839995278df6ff3b1d481/packages/crypto/lib/crypto/crypto.js#L112)

Shouldn't that be `<= 64`?

dated:
> Shouldn't that be `<= 64`?

No, since you're starting with `i = 0`. `<= 64` would write 65 bytes.

zillionn:
@dated You're probably right.

And I believe if the node is out of sync, the transaction-pool will also not return any errors.

zillionn:
What is `32` here, `senderPublicKeyBuffer`?
https://github.com/ArkEcosystem/core/blob/e92ad1548bcf48a3d4a22fd5fe9f681e9617fe63/packages/crypto/lib/crypto/crypto.js#L88

Because `Buffer.from('03318a51e35b8fa96108a2643b17819e4f6d5c7c1df30fbee2102bb18151e09dc4', 'hex').length` returns `33` for me.

Issue #1268, opened 2018-11-01
[dibs] Graphql Scalar Address type


JeremiGendron:
**Describe the bug**

The typedefs for graphql define Scalar Address, thought the composed types (Wallet, TX, Block) all use address: String. I believe in one of the filters it's also recipient: Address, but this one is buggy and yields errors.

```js
console.log __tests__/api/transactions.test.js:87
    { data: { transactions: null },
      errors: 
       [ { message: 'Argument "filter" has invalid value {recipientId: "AHXtmB84sTZ9Zd35h9Y1vfFvPE2Xzqj8ri"}.',
           locations: [Array],
           path: [Array],
           extensions: [Object] } ] }
```

**To Reproduce**
1. Checkout my tests/graphql-completion branch
2. Lerna bootstrap
3. yarn test graphql
4. See logs

**Expected behavior**
The address property uses string input across all composed types, instead of Scalar Address on one and String Address on all others.

**Screenshots**

**Server (please complete the following information):**
 - OS: [e.g. Ubuntu]
 - Version [e.g. 16.04]

**Additional context**

PR #1271, opened 2018-11-01
fix(core-graphql): sorting and limiting of records

JeremiGendron:
## Proposed changes

1. The Transaction type is the only one where the address field (recipientId) has the Address Scalar type, this is unwanted because every other type uses Strings for address property values. The fix is to remove the erroneous Address Scalar typedef and resolvers, which didn't seem to work either.

2. When using orderBy, if mentioning `ASC` as the value it will not be recognized as an AST node by the resolver (tested on transactions endpoint). The reason why is because the database takes queries for ordering in lowercase format (`desc || asc`) only and the helper doesn't convert to lowercase. Easy fix is to convert to lowercase all orderBy direction values (done).

Also instates a collection of graphql tests, most of which work. For transactions, there is still a bug that needs to be figured out where the limit param doesn't get taken into account. I suspect this is because of some hardcoded 100 limit. There was also an issue with `wallets.test.js`, where the limit parameter also is erroneous:  `database.createCondition is not a function`

## Types of changes

## Checklist


JeremiGendron:
closes #1268

JeremiGendron:
Also added: The transaction type test was failing hard because it passed the string value 'TRANSFER' to the DB query, which accepts an Int value; easy fix replace `parameters.type` in the repository (query module for transactions) to `TRANSACTION_TYPES[parameters.type]` in-place. Works good :+1:

JeremiGendron:
Another nasty bug: the resolver for transactions didn't pass the limit to the repository query (db), so it was always undefined and defaulted to 100. Fixed.

JeremiGendron:
Issue potentially averted: the default orderBy string for the wallets resolver is 'height:DESC', this is problematic in many cases as shown with the transactions repository example where the orderBy string is not properly formatted for a DB query.

packages/core-graphql/__tests__/api/wallets.test.js
```diff
@@ -66,10 +66,39 @@ describe('GraphQL API { wallets }', () => {
   })
 
   describe.skip('GraphQL queries for Wallets - testing relationships', () => {
+    it('should verify that relationships are valid', async () => {
+      const query = '{ wallets(limit: 1) { transactions { id } } }'
+      const response = await utils.request(query)
+
+      expect(response).toBeSuccessfulResponse()
 
+      console.log(response.data)
```
faustbrian: Should be either removed or be replaced by an assertion.

packages/core-graphql/__tests__/api/wallets.test.js
```diff
@@ -66,10 +66,39 @@ describe('GraphQL API { wallets }', () => {
...
+      /**
+       *  { data: { wallets: [ [Object] ] },
+            errors: [ { message: 'database.createCondition is not a function',
+              locations: [Array],
+              path: [Array],
+              extensions: [Object] } ]
+          }
+       */
+
+      console.log(response.data.wallets)
```
faustbrian: Should be either removed or be replaced by an assertion.

packages/core-graphql/lib/defs/inputs.js
```diff
@@ -12,7 +12,6 @@ module.exports = `
   scalar JSON
   scalar Limit
   scalar Offset
-  scalar Address
```
faustbrian: Address is the correct type. You can find it in `core-graphql/lib/resolvers/index.js`.

packages/core-graphql/lib/defs/inputs.js
```diff
@@ -35,7 +34,7 @@ module.exports = `
     fee: Float
     blockId: String
     senderPublicKey: String
-    recipientId: Address
+    recipientId: String
```
faustbrian: Address is the correct type. You can find it in `core-graphql/lib/resolvers/index.js`.

packages/core-graphql/lib/repositories/transactions.js
```diff
@@ -47,6 +51,7 @@ class TransactionsRepository extends Repository {
 
     applyConditions([selectQuery, countQuery])
 
+console.log(parameters.limit)
```
faustbrian: `console.logs` don't belong into production code.

packages/core-graphql/lib/resolvers/queries/transaction/transactions.js
```diff
@@ -8,11 +8,10 @@ const { transactions: repository } = require('../../../repositories')
  * @return {Transaction[]}
  */
 module.exports = async (root, args) => {
-  const { orderBy, filter } = args
-
+  const { orderBy, filter, limit } = args
   const order = formatOrderBy(orderBy, 'timestamp:desc')
-
-  const result = await repository.findAll({ ...filter, orderBy: order })
+console.log(limit)
```
faustbrian: `console.logs` don't belong into production code.

packages/core-graphql/__tests__/api/wallets.test.js
```diff
@@ -66,10 +66,39 @@ describe('GraphQL API { wallets }', () => {
...
+      /**
```
faustbrian: This comment should be removed.

packages/core-graphql/__tests__/api/wallets.test.js
```diff
@@ -66,10 +66,39 @@ describe('GraphQL API { wallets }', () => {
...
 
   describe.skip('GraphQL queries for Wallets - testing relationships', () => {
+    it('should verify that relationships are valid', async () => {
+      const query = '{ wallets(limit: 1) { transactions { id } } }'
+      const response = await utils.request(query)
+
+      expect(response).toBeSuccessfulResponse()
 
+      console.log(response.data)
+      /**
```
faustbrian: This comment should be removed.

JeremiGendron:
I found it a bit hard to implement the Address type and also saw all but
one type use String; would it be OK to verify in the resolver instead of
the typedef? @faustbrian

On Nov 1, 2018 11:37 PM, "faustbrian" <notifications@github.com> wrote:

*@faustbrian* requested changes on this pull request.
------------------------------

In packages/core-graphql/__tests__/api/wallets.test.js
<https://github.com/ArkEcosystem/core/pull/1271#discussion_r230262000>:

>
+      console.log(response.data)

Should be either removed or be replaced by an assertion.
------------------------------

In packages/core-graphql/__tests__/api/wallets.test.js
<https://github.com/ArkEcosystem/core/pull/1271#discussion_r230262014>:

>
+      console.log(response.data)
+      /**
+       *  { data: { wallets: [ [Object] ] },
+            errors: [ { message: 'database.createCondition is not a function',
+              locations: [Array],
+              path: [Array],
+              extensions: [Object] } ]
+          }
+       */
+
+      console.log(response.data.wallets)

Should be either removed or be replaced by an assertion.
------------------------------

In packages/core-graphql/lib/defs/inputs.js
<https://github.com/ArkEcosystem/core/pull/1271#discussion_r230262115>:

> @@ -12,7 +12,6 @@ module.exports = `
   scalar JSON
   scalar Limit
   scalar Offset
-  scalar Address

Address is the correct type. You can find it in
core-graphql/lib/resolvers/index.js.
------------------------------

In packages/core-graphql/lib/defs/inputs.js
<https://github.com/ArkEcosystem/core/pull/1271#discussion_r230262121>:

> @@ -35,7 +34,7 @@ module.exports = `
     fee: Float
     blockId: String
     senderPublicKey: String
-    recipientId: Address
+    recipientId: String

Address is the correct type. You can find it in
core-graphql/lib/resolvers/index.js.
------------------------------

In packages/core-graphql/lib/repositories/transactions.js
<https://github.com/ArkEcosystem/core/pull/1271#discussion_r230262156>:

> @@ -47,6 +51,7 @@ class TransactionsRepository extends Repository {

     applyConditions([selectQuery, countQuery])

+console.log(parameters.limit)

console.logs don't belong into production code.
------------------------------

In packages/core-graphql/lib/resolvers/queries/transaction/transactions.js
<https://github.com/ArkEcosystem/core/pull/1271#discussion_r230262161>:

>    const order = formatOrderBy(orderBy, 'timestamp:desc')
-
-  const result = await repository.findAll({ ...filter, orderBy: order })
+console.log(limit)

console.logs don't belong into production code.
------------------------------

In packages/core-graphql/__tests__/api/wallets.test.js
<https://github.com/ArkEcosystem/core/pull/1271#discussion_r230262183>:

>
+      console.log(response.data)
+      /**
+       *  { data: { wallets: [ [Object] ] },
+            errors: [ { message: 'database.createCondition is not a function',
+              locations: [Array],
+              path: [Array],
+              extensions: [Object] } ]
+          }
+       */
+
+      console.log(response.data.wallets)
+      /**

This comment should be removed.
------------------------------

In packages/core-graphql/__tests__/api/wallets.test.js
<https://github.com/ArkEcosystem/core/pull/1271#discussion_r230262192>:

>
+      console.log(response.data)
+      /**

This comment should be removed.

—
You are receiving this because you authored the thread.

Reply to this email directly, view it on GitHub
<https://github.com/ArkEcosystem/core/pull/1271#pullrequestreview-170960754>,
or mute the thread

faustbrian:
1. Why is it hard to implement the `Address` type since it is already implemented?
2. What address fields are using `String` as type instead of `Address`?
3. Why would you verify in the resolver instead of the typedef?

JeremiGendron:
1. It wasn't intuitive how a query should be formatted and the
implementation was incomplete (regex isn't a valid field of
GraphQLTypes.String afaict)
2. In the typedef, Wallet has address: String for example. I think it's the
only place an address is 'used' like that (the implementation doesn't work
out of the box and it's non-obvious why there's an Address type but no
publicKey or vendorfield types).
3. I found it easy to implement: have a regex.test with the address string
and throw an error with a clear message if it's not good (bubbles up nicely
at least in Jest).

I also asked for help initially for this reason (address scalar, how to
present in query), turns out there were more issues (actual bugs) that
needed to be fixed. I admit the Address scalar type should be implemented I
simply really can't find the proper syntax based off documentation and
error messages.

JeremiGendron:
Cleaned it up a lot, also got Address to work (the regex used /g and wasn't working for that reason)

PR #1285, opened 2018-11-04
feat: purge.sh to get rid of docker devnet components only

JeremiGendron:
## Proposed changes

Useful to reset testing plugins or other Core functionality.

Currently, we offer purge_all which removes everything.. This is also useful when having multiple nets on one machine.

## Types of changes

## Checklist


faustbrian:
The purge_all.sh should be sufficient and I wouldn't recommend to run multiple networks at the same time on the same machine to avoid any weird issues.

PR #1289, opened 2018-11-05
fix(crypto): validate the vendor field length

faustbrian:
## Proposed changes

The vendor field length was never validated.

## Types of changes

## Checklist


faustbrian:
Testing sync from 0 at the moment as there are already some transactions that have a vendor field length above 64 characters so we might need to add exceptions.

faustbrian:
Added exceptions for devnet, mainnet should be fine according to this query on a v1 database.

```
ark_mainnet=> SELECT * FROM transactions WHERE length("vendorField") > 64;
 id | rowId | blockId | type | timestamp | senderPublicKey | senderId | recipientId | amount | fee | signature | signSignature | requesterPublicKey | vendorField | signatures | rawasset
----+-------+---------+------+-----------+-----------------+----------+-------------+--------+-----+-----------+---------------+--------------------+-------------+------------+----------
(0 rows)
```

PR #1297, opened 2018-11-06
 fix(client): remove webhooks 'events' API, accept payload in 'update'

thomas-neuman:
## Proposed changes

For body parameters expected by the `PUT /webhooks/{id}` API, accept the necessary payload as a parameter in the client and pass it to the HTTP request.

In addition, the unit tests as well as the mock are adjusted to return the exact status codes as given in the [server's API handler](https://github.com/ArkEcosystem/core/blob/master/packages/core-webhooks/lib/server/handler.js), and the unhandled `events` API has been removed from the client and test suite.

## Types of changes

_`events` API has been removed_

## Checklist


thomas-neuman:
Whoops -- been pulling from 'master' rather than 'develop', so still had these files in the core repo. Will resubmit to the proper js-client

PR #1301, opened 2018-11-06
fix: do not handle all !=TRANSFER transactions as VOTE

vasild:
## Types of changes

## Checklist

### DIFF
--- a/packages/core-transaction-pool/lib/guard.js
+++ b/packages/core-transaction-pool/lib/guard.js
@@ -160,20 +160,38 @@ module.exports = class TransactionGuard {
    * Determines valid transactions by checking rules, according to:
    * - if recipient is on the same network
    * - if sender has enough funds
-   * - if sender has more than one vote/unvote transaction in pool
+   * - if sender already has another transaction of the same type, for types that
+   *   only allow one transaction at a time in the pool (e.g. vote)
    * Transaction that can be broadcasted are confirmed here
    */
   __determineValidTransactions () {
     this.transactions.forEach(transaction => {
-      if (transaction.type === TRANSACTION_TYPES.TRANSFER) {
+      switch (transaction.type) {
+      case TRANSACTION_TYPES.TRANSFER:
         if (!isRecipientOnActiveNetwork(transaction)) {
           this.__pushError(transaction, `Recipient ${transaction.recipientId} is not on the same network: ${configManager.get('pubKeyHash')}`)
           return
         }
-      }
-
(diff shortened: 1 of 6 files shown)


vasild:
Notice: this patch will most likely break things as it would invalidate any non-TRANSFER and non-VOTE transaction. I guess the invalidation added is too strict and should be relaxed.

packages/core-transaction-pool/lib/guard.js
```diff
@@ -165,15 +165,28 @@ module.exports = class TransactionGuard {
...
-      if (this.pool.checkIfSenderHasVoteTransactions(transaction.senderPublicKey)) {
-        this.__pushError(transaction, `Sender ${transaction.senderPublicKey} already has a pending vote transaction in pool`)
+        break
+      case TRANSACTION_TYPES.VOTE:
+        if (this.pool.checkIfSenderHasVoteTransactions(transaction.senderPublicKey)) {
+          this.__pushError(transaction, `Sender ${transaction.senderPublicKey} already has a pending vote transaction in pool`)
+          return
+        }
+        break
+      case TRANSACTION_TYPES.SECOND_SIGNATURE:
```
faustbrian: `default` should be sufficient if none of the other types are handled in another case.

vasild: Yes, technically `default:` suffices, but I added them in order to be explicit at what is being invalidated. I presume nobody remembers all transaction types on top of his head, so reading them here helps to skip looking up the definition of `TRANSACTION_TYPES`.

dated:
```suggestion
      case TRANSACTION_TYPES.SECOND_SIGNATURE:
        if (this.pool.checkIfSenderHasSecondSignatureTransactions(transaction.senderPublicKey)) {
          this.__pushError(transaction, `Sender ${transaction.senderPublicKey} already has a pending second signature transaction in pool`)
          return
        }
        break
```

packages/core-transaction-pool/lib/guard.js
```diff
@@ -165,15 +165,28 @@ module.exports = class TransactionGuard {
...
+      case TRANSACTION_TYPES.DELEGATE_REGISTRATION:
```
dated:
```suggestion
      case TRANSACTION_TYPES.DELEGATE_REGISTRATION:
        if (this.pool.checkIfSenderHasDelegateRegistrationTransactions(transaction.senderPublicKey)) {
          this.__pushError(transaction, `Sender ${transaction.senderPublicKey} already has a pending delegate registration transaction in pool`)
          return
        }
        break
```

Issue #1313, opened 2018-11-07
Transaction Pool flooding and possible solutions (e.g. Throttling)


faustbrian:
The issue with the transaction posting endpoints getting slower over time is related to general performance of the node. It basically boils down to performance degradation of a node if it gets spammed which is caused by more CPU getting used which in turn results in the crypto operations like verification of transactions slowing down of which the end-result is a long response time.

For now we changed the default config to allow less transactions per request (https://github.com/ArkEcosystem/core/commit/23494399a630e1715b82559fac94ea5679b5cc72) which helps the issue but is just a bandaid. Another solution is throttling, even on the P2P layer, which can be touchy.

PR #1314, opened 2018-11-07
feat(core-p2p): add hapi-rate-limit plugin

faustbrian:
## Proposed changes

Adds https://github.com/wraithgar/hapi-rate-limit as rate-limiting/throttling plugin. A list of available options can be found here https://github.com/wraithgar/hapi-rate-limit#options.

I marked this PR was **WIP** because we need to play with the value of the `userLimit` variable to find a good number that makes sense for throttling but won't cause any network issues.

Partially resolves https://github.com/ArkEcosystem/core/issues/1313

## Types of changes

## Checklist


faustbrian:
I've set the default rate limit to 800 for now as the average number of requests per minute between 2 peer seems to be around 600 requests.

PR #1315, opened 2018-11-07
refactor(core-api): add number validation to schemata

dated:
## Proposed changes

Some API endpoints were missing proper validation of the input values in their schemata, such as block and transaction ids, allowing for requests such as the following to be valid:

https://dexplorer.ark.io:8443/api/v2/blocks/xxx
https://dexplorer.ark.io:8443/api/v2/transactions/aaa
https://dexplorer.ark.io:8443/api/v2/votes/bbb

## Types of changes

## Checklist

### DIFF
--- a/packages/core-api/lib/versions/2/schema/blocks.js
+++ b/packages/core-api/lib/versions/2/schema/blocks.js
@@ -11,10 +11,10 @@ exports.index = {
     ...pagination,
     ...{
       orderBy: Joi.string(),
-      id: Joi.string(),
+      id: Joi.string().regex(/^[0-9]+$/, 'numbers'),
       version: Joi.number().integer().min(0),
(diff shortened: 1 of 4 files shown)


packages/core-api/lib/versions/2/schema/blocks.js
```diff
@@ -11,10 +11,10 @@ exports.index = {
     ...pagination,
     ...{
       orderBy: Joi.string(),
-      id: Joi.string(),
+      id: Joi.string().regex(/^[0-9]+$/, 'numbers'),
```
faustbrian: Block IDs can probably be validated as `Joi.number().integer().unsafe()`.

dated: That should probably suffice, as there are no side effects with invalid inputs anyway.

PR #1316, opened 2018-11-07
refactor(transaction-pool): remove redundant memory.js

vasild:
With the new mem pool implementation
packages/core-transaction-pool/lib/memory.js is redundant and is not
necessary.

Problems with memory.js:

* It is not part of the core-transaction-pool interface and callers
  access transactionPool.memory directly

* The memorize() method has complexity O(N^2) where N is the number of
  remembered transactions, see below:

* The way it is used, invalid transactions are remembered forever.

## Types of changes

## Checklist


vasild:
> Looks like a `core-forger` test is failing on this branch.

Yes, and I am quite puzzled by that - can't reproduce locally and couldn't nail it down by staring at the surrounding code and the log. Also, it seems this test did not fail before in other branches and that it is failing deterministically on my branch (but only on circle-ci, not on my comp).

kristjank:
same here, no local test failures

PR #1331, opened 2018-11-08
[WIP] chore: update dependencies

faustbrian:
## Proposed changes

The state machine package and immutable.js got new major releases.

## Types of changes

- [x] Build (changes that affect the build system)

## Checklist


packages/core-blockchain/lib/state-storage.js
```diff
@@ -108,7 +108,7 @@ class StateStorage {
    */
   getLastBlocksByHeight (start, end) {
     end = end || start
-    return _mapToBlockData(_lastBlocks.filter(block => block.data.height >= start && block.data.height <= end)).toArray()
+    return _mapToBlockData(_lastBlocks.filter(block => block.data.height >= start && block.data.height <= end)).valueSeq().toArray()
```
fix: I would keep logic of arrays. Where the length of returned array would be end-start.
Here it is end-start+1. 

With default end = start+1

And filter with < end

faustbrian: Will make that change in another PR. The change from `toArray()` to `valueSeq().toArray()` is just because of a breaking change in the `immutable.js` package.

PR #1337, opened 2018-11-08
WIP: perf(core-transaction-pool-mem): Do not call getTransactionIdsForForging(0, pool size)

vasild:
Requesting lots of transactions from getTransactionIdsForForging()
can be slow and we only need a few (usually 150) => avoid calling
getTransactionIdsForForging(0, pool size) unnecessary.

## Types of changes

## Checklist

### DIFF
--- a/packages/core-transaction-pool-mem/__tests__/connection.test.js
+++ b/packages/core-transaction-pool-mem/__tests__/connection.test.js
@@ -360,7 +360,13 @@ describe('Connection', () => {
       expect(connection.getTransactionsForForging).toBeFunction()
     })
 
-    it('should return an array of transactions', async () => {
+    it('should return empty array when 0 is requested', async () => {
+      const transactions = await connection.getTransactionsForForging(0)
+      expect(transactions).toBeArray()
+      expect(transactions.length).toBe(0)
+    })
+
+    it('should skip low fee transactions', async () => {
       connection.addTransaction(mockData.dummy1)
       connection.addTransaction(mockData.dummy2)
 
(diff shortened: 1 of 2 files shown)


vasild:
This patch assumes that most of the transactions in the pool will pass `checkDynamicFeeMatch()` and `checkApplyToBlockchain()`. But if this is not the case, then the performance will be worse than it is before the patch. For example if we have a pool with 30'000 transactions that do not pass those tests, we will end up calling:
```
getTransactionIdsForForging(0, 200)
getTransactionIdsForForging(0, 400)
getTransactionIdsForForging(0, 800)
getTransactionIdsForForging(0, 1600)
getTransactionIdsForForging(0, 3200)
getTransactionIdsForForging(0, 6400)
getTransactionIdsForForging(0, 12800)
getTransactionIdsForForging(0, 25600)
getTransactionIdsForForging(0, 30000)
```
which is worse than before the patch, just:
```
getTransactionIdsForForging(0, 30000)
```

Thus, performance-wise, this patch depends on not allowing too low fee transactions to enter the pool in the first place.

vasild:
Maybe one of these could be better:
* Try (0, N) and if it does not deliver then directly (0, pool size), instead of (0, N*2), (0, N*4), ... (0, pool size)
* Try (0, N) and if it does not deliver then try (N, M) and handle possible inconsistencies due to pool being changed between the calls

kristjank:
@vasild as you said, this is an issue. We are changing this and moving dynamic fee check on the pool entry door, meaning this function can be simplified back to return only first 150 transactions from the pool.

kristjank:
@vasild join this branch here please #1343

vasild:
Overridden by https://github.com/ArkEcosystem/core/pull/1343

PR #1343, opened 2018-11-09
refactor(core-transaction-pool): avoid that transactions below the minimum fee enter the pool

kristjank:
## Proposed changes
If fee doesn't match the recipient pool, it is not allowed to enter the pool. Transaction will be broadcasted to other peers.

This also makes the call for getTransactionsForForging more effective.

## Types of changes

## Checklist

## Further comments
Improve return answer to the client (2.1)


packages/core-transaction-pool/lib/guard.js
```diff
@@ -117,15 +118,22 @@ module.exports = class TransactionGuard {
...
    * @return {void}
    */
   __transformAndFilterTransactions (transactions) {
     this.transactions = []
 
     transactions.forEach(transaction => {
-      const exists = this.pool.transactionExists(transaction.id)
+      if (!dynamicFeeMatch(transactions)) {
+        this.broadcast.push(transaction)
+        // TODO: currently no id is returned to the sender so he doesnt know what is with this id?
```
faustbrian: Why is no fee returned to the sender, shouldn't the ID be in the broadcast array?

Maybe we should also push the transaction as an error with the message `Peer rejected the transaction because of not meeting the minimum accepted fee.` so the user knows it has been broadcasted *but* also gets an indication of why it was not accepted since *rejection* technically is an error.

PR #1354, opened 2018-11-09
fix: tx builder network sign

alexbarnsley:
## Proposed changes

Use network version in transaction builder when signing with mixin

## Types of changes

## Checklist


.circleci/config.yml
```diff
@@ -39,6 +39,7 @@ jobs:
       - save_cache:
           key: 'core-node10-{{ checksum "checksum.txt" }}-1'
           paths:
+            - ./packages/client/node_modules
```
vasild: I did `mv .git/hooks .git/hooks_` after this was sneaked into my commits.

PR #1360, opened 2018-11-10
fix(crypto): remove maximum payload length check for blocks

zillionn:
Removed some forgotten payload length check.

## Types of changes

## Checklist

### DIFF
--- a/packages/crypto/lib/models/block.js
+++ b/packages/crypto/lib/models/block.js
@@ -250,10 +250,6 @@ module.exports = class Block {
       //   }
       // }
 
-      if (block.payloadLength > constants.maxPayload) {
-        result.errors.push('Payload length is too high')
-      }
-
       let size = 0
       let payloadHash = createHash('sha256')
 


packages/crypto/lib/models/block.js
```diff
@@ -250,10 +250,6 @@ module.exports = class Block {
       //   }
       // }
 
-      if (block.payloadLength > constants.maxPayload) {
```
faustbrian: What is the reasoning for removing that check?

The variable still exists and should be verified as you otherwise can include more data in blocks then allowed. https://github.com/ArkEcosystem/core/blob/develop/packages/crypto/lib/networks/ark/devnet.json#L58

zillionn: There is no `constants.maxPayload`, the correct constant is `constants.block.maxPayload` and the correct check is here: https://github.com/ArkEcosystem/core/blob/3059c238800b3d7b7249f1abaccb35b740084838/packages/crypto/lib/models/block.js#L323

PR #1380, opened 2018-11-12
fix: orderBy for wallet transaction endpoints

alexbarnsley:
## Proposed changes

Allows the orderBy query string parameter for wallet transactions

## Types of changes

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -39,6 +39,7 @@ jobs:
       - save_cache:
           key: 'core-node10-{{ checksum "checksum.txt" }}-1'
           paths:
+            - ./packages/client/node_modules
             - ./packages/core/node_modules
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -65,6 +66,7 @@ jobs:
             - ./packages/core-tester-cli/node_modules
             - ./packages/core-transaction-pool/node_modules
             - ./packages/core-transaction-pool-mem/node_modules
+            - ./packages/core-transaction-pool-redis/node_modules
```
boldninja: Don't think we have redis anymore :thinking:

alexbarnsley: Already reverted those :danstop: it got auto-added because i still had the folder in my local, which i've now removed

PR #1383, opened 2018-11-12
feat(core-transaction-pool): log if sender is blocked

zillionn:
## Proposed changes
If the sender is blocked no trxs are accepted from that sender, but there is no indication in the logs. This PR shows a debug message if sender is blocked:
`[2018-11-12 14:43:55][DEBUG] : Transaction 94c5ce3d3df219dfb409392a41a5133ba0da9fc09858f354e2bb5d3518ad157f rejected. Sender 0366da7ed85cc1e73a13fafb8007d2609d92239355847343cc7b12d85a65d25502 is blocked until Mon Nov 12 2018 15:22:51 GMT+0000 ⏱`

## Types of changes

## Checklist

### DIFF
--- a/packages/core-transaction-pool/lib/guard.js
+++ b/packages/core-transaction-pool/lib/guard.js
@@ -133,7 +133,7 @@ module.exports = class TransactionGuard {
     transactions.forEach(transaction => {
       const exists = this.pool.transactionExists(transaction.id)
 
-      if (!exists && !this.pool.isSenderBlocked(transaction.senderPublicKey)) {
+      if (!exists && !this.pool.isSenderBlocked(transaction.senderPublicKey, transaction.id)) {
         try {
           const trx = new Transaction(transaction)
 
--- a/packages/core-transaction-pool/lib/interface.js
+++ b/packages/core-transaction-pool/lib/interface.js
@@ -150,7 +150,7 @@ module.exports = class TransactionPoolInterface {
    * @param  {String} senderPublicKey
    * @return {Boolean}
    */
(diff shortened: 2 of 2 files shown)


packages/core-transaction-pool/lib/guard.js
```diff
@@ -133,7 +133,7 @@ module.exports = class TransactionGuard {
     transactions.forEach(transaction => {
       const exists = this.pool.transactionExists(transaction.id)
 
-      if (!exists && !this.pool.isSenderBlocked(transaction.senderPublicKey)) {
+      if (!exists && !this.pool.isSenderBlocked(transaction.senderPublicKey, transaction.id)) {
```
kristjank: in this case pass the whole transaction object to the function, so something like isSenderBlocked(transaction)

packages/core-transaction-pool/lib/interface.js
```diff
@@ -150,7 +150,7 @@ module.exports = class TransactionPoolInterface {
    * @param  {String} senderPublicKey
    * @return {Boolean}
    */
-  isSenderBlocked(senderPublicKey) {
+  isSenderBlocked(senderPublicKey, transactionId) {
```
kristjank: update to comment above

packages/core-transaction-pool/lib/guard.js
kristjank: this things are our of log for the reason of not filling the logs too much, but I see the usecase, and as @dated suggested - we can then move this up higher

PR #1413, opened 2018-11-16
 feat(core-transaction-pool): implement pool cap

vasild:
## Types of changes

## Checklist


packages/crypto/lib/networks/ark/mainnet.json
```diff
@@ -28,26 +28,32 @@
...
-        "delegateRegistration": 500,
-        "vote": 100,
-        "multiSignature": 500,
-        "ipfs": 250,
-        "timelockTransfer": 500,
-        "multiPayment": 500,
-        "delegateResignation": 500
+        "dynamicFees": {
+          "minWhatever1": 1000,
+          "minWhatever2": 1000,
```
vasild: Devise real names for `minWhatever1` and `minWhatever1`. One is "minimum fee to be eligible to enter the pool (arktoshi/byte)" and the other is "minimum fee to broadcast (arktoshi/byte)".

faustbrian:
Re-added https://github.com/ArkEcosystem/core/pull/1413/commits/cdee3f43cfd7cc06ac227c08675811123364413d because it is the P2P endpoint nodes use to send transactions to each other. https://github.com/ArkEcosystem/core/tree/develop/packages/core-p2p/lib/server/versions/peer is not yet in use.

_Must have been my bad a few days ago in slack because I confused 2 links you send, sorry about that. Keep in mind though that we cannot throttle transactions or alter the response format on this endpoint as it is used by nodes to talk to each other, not humans._.

packages/core-p2p/lib/server/versions/peer/handlers/transactions.js
```diff
@@ -37,44 +37,43 @@ exports.store = {
...
-      request.payload.transactions,
-    )
-
     const guard = new TransactionGuard(transactionPool)
 
-    for (const ne of notEligible) {
-      guard.invalidate(ne.transaction, ne.reason)
+    const result = await guard.validate(request.payload.transactions)
+
+    if (result.invalid.length > 0) {
```
zillionn: So, if even only 1 trx in the payload is `invalid`, you'll reject all of them?

vasild: The code before this patch:
```
  async handler(request, h) {
    const { eligible, notEligible } = transactionPool.checkEligibility(
      request.payload.transactions,
    )

    const guard = new TransactionGuard(transactionPool)

    for (const ne of notEligible) {
      guard.invalidate(ne.transaction, ne.reason)
    }

    await guard.validate(eligible)

    if (guard.hasAny('invalid')) {
      return Boom.notAcceptable(
        'Transactions list could not be accepted.',
        guard.errors,
      )
    }
```
and after the patch:
```
    const guard = new TransactionGuard(transactionPool)

    const result = await guard.validate(request.payload.transactions)

    if (result.invalid.length > 0) {
      return Boom.notAcceptable('Transactions list could not be accepted.', guard.errors)
    }
```
So, I am mimicking the behavior as of before the patch. With such big patches that change lots of things it is best to avoid behavioral changes.

packages/crypto/lib/networks/ark/mainnet.json
supaiku0: How about `minFeePool`/`minFeeEnterPool`  and `minFeeBroadcast`?

PR #1424, opened 2018-11-17
feat: POST payload validation

supaiku0:
## Proposed changes

Adds payload validation to the respective block and transaction `store` endpoints

Notes:
- Strips properties which are not defined in the schema (`stripUnknown`)
- When POSTing multiple txs, if one fails Joi validation, the request fails
- Joi aborts validation on the first error, so you don't know for sure if only one is invalid (`abortEarly`)
- Returned are the default Joi error messages which can be hard to interpret... I experimented briefly with the error calback to return custom errors, but needs a bit more thought
- Tests are green and I could use core-tester-cli to send tx without issues, but there are probably some edge cases
- Runs fine on my forger

## Types of changes

## Checklist

### DIFF
--- a/packages/core-api/lib/versions/2/schema/transactions.js
+++ b/packages/core-api/lib/versions/2/schema/transactions.js
@@ -1,5 +1,5 @@
-const Joi = require('joi')
 const container = require('@arkecosystem/core-container')
+const Joi = require('@arkecosystem/crypto').validator.engine.joi
 const pagination = require('./pagination')
 
 /**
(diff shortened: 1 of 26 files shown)


faustbrian:
`When POSTing multiple txs, if one fails Joi validation, the request fails.` - This should be fine as failed validation via Joi means that the structure of a transaction is completely borked.

`Returned are the default Joi error messages which can be hard to interpret... I experimented briefly with the error calback to return custom errors, but needs a bit more thought` - The vague error messages of Joi are something that needs to and will be improved overall in 2.1.

PR #1430, opened 2018-11-19
refactor: log reason why transaction fails

### LINKED ISSUE #1365: Add useful logging around why transaction validation fails
Right now, when a transaction fails, a generic "transaction failed" is logged. It's a lot more useful for diagnosing purposes to log WHY the transaction failed based on those conditions we test.

Log all the reasons why a transaction has failed validation.
This is a little tricky because the validation stuff is in the "crypto" package which doesn't have a reference to the logger package because it might cause a circular dependency.

paroxysm:
## Proposed changes

resolves #1365 

## Types of changes

## Checklist

<!--

### DIFF
--- a/packages/crypto/__tests__/handlers/transactions/transfer.test.js
+++ b/packages/crypto/__tests__/handlers/transactions/transfer.test.js
@@ -18,11 +18,11 @@ describe('TransferHandler', () => {
       expect(handler.canApply).toBeFunction()
     })
 
-    it('should be truthy', () => {
+    it('should be true', () => {
       expect(handler.canApply(wallet, transaction, [])).toBeTrue()
     })
 
-    it('should be falsy', () => {
+    it('should be false', () => {
       transaction.senderPublicKey = 'a'.repeat(66)
 
       expect(handler.canApply(wallet, transaction, [])).toBeFalse()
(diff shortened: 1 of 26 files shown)


paroxysm:
Missed a few spots.

packages/crypto/__tests__/handlers/transactions/transfer.test.js
```diff
@@ -18,14 +18,14 @@ describe('TransferHandler', () => {
       expect(handler.canApply).toBeFunction()
     })
 
-    it('should be ok', () => {
-      expect(handler.canApply(wallet, transaction)).toBeTrue()
+    it('should be truthy', () => {
```
faustbrian: Should be described as `true` instead of `truthy` as `truthy` can mean multiple types that are not empty rather then a `bool`.

packages/crypto/__tests__/handlers/transactions/transfer.test.js
```diff
@@ -18,14 +18,14 @@ describe('TransferHandler', () => {
...
+      expect(handler.canApply(wallet, transaction, [])).toBeTrue()
     })
 
-    it('should not be ok', () => {
+    it('should be falsy', () => {
```
faustbrian: Should be described as `true` instead of `truthy` as `truthy` can mean multiple types that are empty rather then a `bool`.

paroxysm: Changed

PR #1439, opened 2018-11-19
refactor(core-transaction-pool): apply transaction after it has been added

### LINKED ISSUE #1349: Add check for crypto verified upon loading transaction from pool db
Add check for crypto verified upon loading transaction from pool db to memory.

When loading transactions from local storage db - check if crypto verify was a success. The 
`const trx = new Transaction()` sets flag `verified`

**Describe alternatives you've considered**
Similar functionality i used when entering the pool here: https://github.com/ArkEcosystem/core/blob/develop/packages/core-transaction-pool/lib/guard.js#L131-L133

supaiku0:
## Proposed changes

Refactors the filter logic from the guard and moves the `applyTransaction` on the pool wallet manager to the transaction pool. This way a transaction is added first and then applied. So the guard no longer has to check wether a tx is already in the pool, after it applied a transaction, because of the async nature.

Couldn't test it properly yet, since devnet is broken.

Also fixes #1349 

## Types of changes

## Checklist

### DIFF
--- a/packages/core-api/lib/versions/2/handlers/transactions.js
+++ b/packages/core-api/lib/versions/2/handlers/transactions.js
@@ -55,42 +55,23 @@ exports.store = {
 
     const result = await guard.validate(request.payload.transactions)
 
-    // key=id, val=transaction, used to remove entries with sub-linear complexity
-    const broadcast = new Map(result.broadcast.map(t => [ t.id, t ]))
-
     if (result.accept.length > 0) {
-      const addResult = transactionPool.addTransactions(result.accept)
-
-      result.accept = addResult.added
-
-      for (const notAdded of addResult.notAdded) {
-        result.invalid.push(notAdded.transaction)
-        const id = notAdded.transaction.id
-
-        if (result.errors[id] === undefined) {
-          result.errors[id] = []
-        }
-        result.errors[id].push({ type: 'ERR_FULL_POOL', message: notAdded.reason })
-
-        broadcast.delete(id)
-      }
-
       const len = result.accept.length
       logger.info(`Accepted ${len} new ${pluralize('transaction', len)}`)
     }
 
-    if (broadcast.size > 0) {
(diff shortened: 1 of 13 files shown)


zillionn:
> 0|ark-core | Debug: internal, implementation, error 
0|ark-core |     TypeError: Cannot read property 'length' of undefined
0|ark-core |     at DynamicFeeManager.calculateFee (/home/darknodeuser/ark-core/packages/crypto/lib/managers/dynamic-fee.js:23:59)
0|ark-core |     at module.exports.transaction (/home/darknodeuser/ark-core/packages/core-transaction-pool/lib/utils/dynamicfee-matcher.js:25:47)

zillionn:
I believe I found out where the whole problem is. Now with your patch, I send only about 40-80 trxs like this:
`~/ark-core/packages/core-tester-cli/bin/tester transfer --skip-validation --base-url http://127.0.0.1 -p "sender passphrase" -a 0.00000001 --transfer-fee 0.10000000 -n 40 --smart-bridge 'spam'`

and I got thousands of lines like this one:
> Transaction pool: allowing sender public key: 0366da7ed85cc1e73a13fafb8007d2609d92239355847343cc7b12d85a65d25502 (listed in options.allowedSenders), thus skipping throttling.

Check the log [guard.validate.log](https://github.com/ArkEcosystem/core/files/2597422/guard.validate.log)

The cause is that `guard.validate()`, for some reason, is called hundreds of times here:
https://github.com/ArkEcosystem/core/blob/d0be57584c17f30b4384368570235626a2aa30b4/packages/core-p2p/lib/server/versions/1/handlers.js#L316

For some reason `v1/postTransactions` ends up in some loop.

To test this I just changed this line to return error and no more spam/connection issues:
https://github.com/ArkEcosystem/core/blob/d0be57584c17f30b4384368570235626a2aa30b4/packages/core-p2p/lib/server/versions/1/handlers.js#L286

So, I hope you'll find the culprit cause I need to sleep :).

Issue #1464, opened 2018-11-22
Network does not adapt to diverse maxTransactionsPerRequest settings


vasild:
Each relay node has a setting maxTransactionsPerRequest which determines the maximum number of transactions it would accept in one request. If an acceptable request is received with N transactions (N <= maxTransactionsPerRequest) it is broadcasted to other peers also as one request with N transactions.

This is a problem if the recipient of the broadcast has set maxTransactionsPerRequest to a value smaller than N. Then the entire request is rejected and no attempt is made by the sending peer to adapt and send less transactions per request to the recipient.

To fix this:
* The rejection should contain a hint as to what is the maximum accepted. E.g. "You sent me 200, but I would accept maximum 160 per request".
* The broadcaster should interpret that, reduce the number of transactions in the request and retry.

zillionn:
Why do this and send several requests, when the receiver can just slice the request and accept only as much as it's `maxTransactionsPerRequest`?

zillionn:
Oh, you'll send requests until the other relay gets all the txs?

PR #1468, opened 2018-11-22
refactor(core-p2p): broadcast transactions in chunks to meet peers requirements

faustbrian:
## Proposed changes

Follow up to https://github.com/ArkEcosystem/core/pull/1466. Resolves https://github.com/ArkEcosystem/core/issues/1464.

## Types of changes

## Checklist


packages/core-p2p/lib/peer.js
```diff
@@ -65,16 +66,29 @@ module.exports = class Peer {
...
+        },
+      )
+
+    try {
+      await broadcast(transactions)
+    } catch (err) {
+      if (err.response && err.response.status === 413) {
+        const items = chunk(transactions, err.response.data.error.allowed)
+
+        await Promise.all(items.map(item => broadcast(item)))
```
zillionn: Hm, are you sure this is a good idea? What if sender has 200 txs and receiver allows max 10, then that's 20 requests one after another. How about a little delay between the requests, like 200ms or something?

vasild: I think that sender should be sending as much as possible as faster as possible. If that does not suit the receiver he should respond accordingly. Arbitrary delay on the sender side has the problems that 1. most of the time it will cause unnecessary slowdown and 2. maybe it will not meet the requirements on the receiver - thus be useless.

alexbarnsley: @vasild 	what if receiver only allows 1 tx per request? that will hold up the sender node because it won't continue until it's finished, which could also affect other areas of the sending node.

vasild: If that would be a problem, we can check if `err.response.data.error.allowed` is too low. But I do not think this is the case - the overhead for sending 1 transaction per request and 100 transactions per request is comparable (I did not do the exact calculations).

kristjank: This opens possible attack window, and also await will delay it.
This is critical p2p and nodes should be in sync with the same protocol. Either it recipient accepts the payload or not, they dont want to have a chat, and also not every pool is required to be the same.

faustbrian:
@air1one 😭

packages/core-p2p/lib/peer.js
zillionn: You're probably right, but I just don't like such loops and I'm also worried because the main problem with the network was the broadcasting, because of too many connections peers were blocking each other and the network loses quorum. I hope this will not cause the same issues.

PR #1488, opened 2018-11-23
perf: don't always purge expired txs when checking for existence

vasild:
Before we check whether a transaction exists in the pool we remove
expired ones to avoid false positive - pretending that an expired
transaction exists in the pool.

However, if a transaction is not in the pool there is no need to purge
in order provide a definite answer.

## Types of changes

## Checklist


supaiku0:
Approved, but in this case you can just `return true` at the end.

vasild:
> Approved, but in this case you can just `return true` at the end.

No, because we do not know whether the purge removed the tx or not.

PR #1499, opened 2018-11-26
fix(core-blockchain): remove invalid transactions from pool when receiving a bad block

supaiku0:
## Proposed changes

When `processBlock` encounters a bad block, purge all transactions from any sender with invalid transactions in that block.

## Types of changes

## Checklist

### DIFF
--- a/packages/core-transaction-pool-mem/__tests__/connection.test.js
+++ b/packages/core-transaction-pool-mem/__tests__/connection.test.js
@@ -692,9 +692,9 @@ describe('Connection', () => {
     })
   })
 
-  describe('purgeInvalidTransactionsFromBlock', () => {
+  describe('purgeSendersWithInvalidTransactions', () => {
     it('should be a function', () => {
-      expect(connection.purgeInvalidTransactionsFromBlock).toBeFunction()
+      expect(connection.purgeSendersWithInvalidTransactions).toBeFunction()
     })
 
     it('should purge transactions from sender when invalid', async () => {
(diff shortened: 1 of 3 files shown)


packages/core-transaction-pool-mem/__tests__/connection.test.js
```diff
@@ -691,4 +691,48 @@ describe('Connection', () => {
...
+        1,
+      )
+
+      const block = {
+        transactions: [...transfersA, ...transfersB],
+      }
+
+      block.transactions.forEach(tx => connection.addTransaction(tx))
+
+      expect(connection.mem.all).toHaveLength(6)
```
vasild: There is `connection.getPoolSize()` method.

```
       expect(connection.getPoolSize()).toBe(6) 
```

packages/core-transaction-pool/lib/interface.js
```diff
@@ -280,6 +280,21 @@ module.exports = class TransactionPoolInterface {
...
 
+  /**
+   * Purges all transactions from senders with at least one
+   * invalid transaction.
+   * @param {Block} block
+   */
+  purgeSendersWithInvalidTransactions(block) {
+    const publicKeys = new Set(
+      block.transactions
+        .filter(tx => !tx.verified)
```
vasild: Isn't it better to prevent transactions with `.verified == false` from entering the pool in the first place, rather than letting them in and removing them afterwards?

kristjank: this is done in the guard

PR #1507, opened 2018-11-26
chore(core-p2p): add secDecimalDigits option to prettyms call

dated:
## Proposed changes

Adds the `secDecimalDigits` to certain `prettyms` calls where the precision is unnecessary, for instance:

```
x.x.x.x still suspended for 1 minute 59.3 seconds because of "Timeout".
```

## Types of changes

## Checklist


faustbrian:
I like some good precision 👀

PR #1522, opened 2018-11-27
fix(core-transaction-pool): apply tx to recipient wallet when sender unknown

supaiku0:
## Proposed changes

We only checked if a sender wallet is present, so if it is unknown to the pool, the recipient wallet would disconnect from the blockchain wallet state as we never applied the transaction.

## Types of changes

## Checklist


packages/core-transaction-pool/lib/interface.js
```diff
@@ -199,20 +199,30 @@ module.exports = class TransactionPoolInterface {
...
         } catch (error) {
           logger.error(`AcceptChainedBlock in pool: ${error}`)
-          this.purgeByPublicKey(transaction.senderPublicKey)
+          // Purge sender
+          this.purgeByPublicKey(senderPublicKey)
+
+          // Purge recipient if not sender
+          if (senderPublicKey !== recipientWallet.publicKey) {
+            this.purgeByPublicKey(recipientWallet.publicKey)
+          }
```
vasild: Why remove also recipient's transactions?

This way somebody may deprive somebody else from sending anything by constantly purging all his transactions via this code. For example, A periodically generates a transaction A->B such that it would end up here, doing line 224. We remove all B's transactions from the pool. The innocent B is never able to send anything to anybody because his transactions are constantly being removed.

Issue #1494, opened 2018-11-25
p2p: broadcast in chunks


zillionn:
**Describe the bug**
This is not working:
https://github.com/ArkEcosystem/core/blob/df3f087b887b22f15f40c3aa791b29451235fd5b/packages/core-p2p/lib/peer.js#L76-L78

Because `__broadcastTransactions `/`__post` doesn't throw any errors:
https://github.com/ArkEcosystem/core/blob/df3f087b887b22f15f40c3aa791b29451235fd5b/packages/core-p2p/lib/peer.js#L257-L275

**Expected behavior**
To broadcast in chunks.

Also imo this is not the correct way to do it. You're making 1 extra request. For example:
- 1st req: send 45 txs, the receiver will reject the whole request and return `allowed: 40`
- 2nd req: send 40 txs
- 3rd req: send 5 txs

IMO this is the correct way:
- 1st req: send 45 txs, the receiver should accept only the first `40` and return the excess `5`
- 2nd req: send the last 5 txs

**Server (please complete the following information):**
 - OS: Ubuntu
 - Version: 18.04
 - Ark Core: df3f087b

zillionn:
And btw I still believe this rebroadcasting is absolutely unnecessary. I don't get it, how exactly `maxTransactionsPerRequest` is helping, if you're going to send all the txs in chunks anyway? IMO it only hurts the performance because the sender will make several requests, when it could be just 1. `maxTransactionsPerRequest` should be hardcoded and receiver should rejects the request from the client in the first place. If you change your relay's hardcoded value, others will just reject your large broadcasting requests.

paroxysm:
I agree with your suggested solution, it's more elegant and efficient.

PR #1523, opened 2018-11-28
refactor(core-p2p): remove chunk broadcasting

mvega3:
## Proposed changes
This is a first implementation of the p2p algorithm specified by @zillionn in #1494. I'm not familiar enough with the rationale behind the limit to suggest which solution is best -- the current code, Zillion's first solution (which is what this PR attempts to implement), or his second solution: to forgo rebroadcasting.

If the decision is to forgo rebroadcasting, most of this code can be rolled back and replaced with the previous error code commented out in this PR.

## Types of changes

## Checklist

### DIFF
--- a/packages/core-p2p/lib/peer.js
+++ b/packages/core-p2p/lib/peer.js
@@ -310,7 +310,7 @@ module.exports = class Peer {
 
     if (
       response &&
-      response.status === 202 &&
+      response.status === 200 &&
       !response.data.excessTransactions
     ) {
       return response
--- a/packages/core-p2p/lib/peer.js
+++ b/packages/core-p2p/lib/peer.js
@@ -308,24 +308,6 @@ module.exports = class Peer {
       throw new Error(err.message)
     }
 
-    if (
-      response &&
-      response.status === 200 &&
-      !response.data.excessTransactions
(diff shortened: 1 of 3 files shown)


kristjank:
personally - I recommend removing this and hardcoding the maxRequestValue. We will do more harm otherwise.

mvega3:
@kristjank noted, I'll make those changes

packages/core-p2p/lib/peer.js
```diff
@@ -300,15 +291,23 @@ module.exports = class Peer {
...
-      },
-      {
-        headers: this.headers,
-        timeout: 8000,
-      },
-    )
+    let response
+
+    try {
+      response = await this.__post(
```
faustbrian: We can just return here, there is no point in assigning it to a variable if we are not going to do anything with it afterwards besides returning it.

packages/core-p2p/lib/peer.js
```diff
@@ -76,16 +76,7 @@ module.exports = class Peer {
     try {
```
faustbrian: This try/catch won't do anything anymore because __broadcastTransactions already catches any exceptions.

PR #1528, opened 2018-11-28
chore(core-p2p): pluralize discovered peers log

dated:
## Proposed changes

Pluralize discovered peers log.

## Types of changes

- [x] Other... Please describe: chore, pluralization

## Checklist


faustbrian:
There will never be only 1 peer, it will discover peers until there are enough.

dated:
Although unlikely, it should still be possible that only one peer is on one specific version, no?

PR #1537, opened 2018-11-28
fix(core-snapshots-cli) fix app being declared twice

vmunich:
## Proposed changes

`const app` was being declared twice, preventing the snapshot script to run. This fixed the issue and I was able to create, rollback and restore snapshots again.

## Types of changes

## Checklist

- [ ] I have read the [CONTRIBUTING](https://docs.ark.io/developers/guidelines/contributing.html) documentation (link is not working btw)


arkmoon:
This is actually incorrect. The offending line is below, we actually need line 3 to function.

PR #1538, opened 2018-11-29
fix(core-snapshots-cli) fix app being declared twice

arkmoon:
## Proposed changes
`const app` declared twice, removing the second as it appears it's not needed.


faustbrian:
Resolved by https://github.com/ArkEcosystem/core/pull/1537

Issue #1402, opened 2018-11-15
AssertionError


zillionn:
**Describe the bug**

> 0|ark-core-relay  | [2018-11-15 19:33:29][INFO]    : Received new block at height 755,813 with 0 transactions from 185.170.115.40
0|ark-core-relay  | [2018-11-15 19:33:29][DEBUG]   : Delegate proxima_centauri_b (03120f521f7025f76341a09112f88a6c072411c549e4bfa8c92946fcf1c57cdf1e) allowed to forge block 755,813 👍
0|ark-core-relay  | [2018-11-15 19:33:29][DEBUG]   : event 'PROCESSFINISHED': "newBlock" -> "idle" -> actions: [{"type":"checkLater"},{"type":"blockchainReady"}]
0|ark-core-relay  | [2018-11-15 19:33:30][INFO]    : Broadcasting block 755,813 to 57 peers
1|ark-core-forger | [2018-11-15 19:33:38][INFO]    : Next forging delegate zillion (0366da7ed85cc1e73a13fafb8007d2609d92239355847343cc7b12d85a65d25502) is active on this node.
1|ark-core-forger | [2018-11-15 19:33:38][DEBUG]   : Sending wake-up check to relay node http://127.0.0.1:4002
0|ark-core-relay  | [2018-11-15 19:33:38][DEBUG]   : Blockchain sync check WAKEUP requested by forger 🛏
0|ark-core-relay  | [2018-11-15 19:33:38][DEBUG]   : event 'WAKEUP': "idle" -> {"syncWithNetwork":"syncing"} -> actions: [{"type":"checkLastDownloadedBlockSynced"}]
0|ark-core-relay  | [2018-11-15 19:33:38][DEBUG]   : Queued blocks for rebuildQueue: 0
0|ark-core-relay  | [2018-11-15 19:33:38][DEBUG]   : Queued blocks for processQueue: 0
0|ark-core-relay  | [2018-11-15 19:33:38][DEBUG]   : event 'SYNCED': {"syncWithNetwork":"syncing"} -> {"syncWithNetwork":"downloadFinished"} -> actions: [{"type":"downloadFinished"}]
0|ark-core-relay  | [2018-11-15 19:33:38][INFO]    : Block download finished 🚀
[...]

PR #1542, opened 2018-11-29
refactor: add "valid-as-of-block" in reply when getting txs to forge

vasild:
When getting the top transactions from the pool with the intention to
forge them in the next block we don't want already forged transactions
to be in that list. However, the pool cannot guarantee that there will
not be forged transactions because of the nature of forging - new blocks
are forged every 8 seconds. So even if the reply from the pool is
"correct" (it does not contain forged transactions), then it could happen
that 0.1 seconds after that some of those transactions get forged.

We can guarantee that the list of returned transactions does not contain
forged ones as of some block height (H). That is - none of the returned
transactions is forged in block with height H or earlier blocks. So we
append H to the list of transactions in the reply. The forger knows the
height of the block he is forging and can decide if the reply suits him
or not. This patch changes the forger to retry a few times if the reply
is outdated and if still outdated - try to forge a block nevertheless.

This fixes https://github.com/ArkEcosystem/core/issues/1402

## Types of changes

## Checklist


kristjank:
Just had another idea. So forgers know when is his turn. Befor his turn, when the previous forger is going to forge, forger already sends a wake call to the node. We have enough time 12s in this window to also call removeForgedTransactions. In this way we could then completely remove the call from getTransactions, as they would be removed before the actual forging starts.

Then the block of previous delegate is received and its transactions removed from pool.

packages/core-blockchain/lib/blockchain.js
```diff
@@ -528,20 +528,28 @@ module.exports = class Blockchain {
...
    */
-  getUnconfirmedTransactions(blockSize) {
-    const transactions = this.transactionPool.getTransactionsForForging(
-      blockSize,
-    )
+  async getUnconfirmedTransactions(n, asOfHeight = undefined) {
+    let transactions = this.transactionPool.getTransactions(0, n)
+    const transactionsIds = transactions.map(t => t.id)
+
+    const forgedIds = await this.database.getForgedTransactionsIds(transactionsIds, asOfHeight)
```
kristjank: change to lastForged Height where this forger has succesfully forged. 
`asOfHeight=lastForgedBlockByDelegate.height`

packages/core-forger/lib/manager.js
```diff
@@ -202,7 +202,7 @@ module.exports = class ForgerManager {
     //
     // this.client.emitEvent('forger.started', delegate.publicKey)
 
-    const transactions = await this.__getTransactionsForForging()
+    const transactions = await this.__getTransactionsForForging(round.lastBlock.height)
```
kristjank: same comment here - maybe round.lastForgedBlock by delegate

faustbrian:
@vasild This can now be adjusted to TypeScript on the `develop` branch.

faustbrian:
@kristjank do we close this as was discussed to look for a simpler solution?

faustbrian:
Will close this for, can revisit it later on if needed as simpler solutions were discussed.

PR #1547, opened 2018-11-30
fix: Transaction pool wallet manager apply logic change

supaiku0:
## Proposed changes

TL;DR: Don't apply transaction to recipient when entering pool.

The issue in the pool-wallet-manager was related to the pool balance apply logic. Since we were tracking the balance spent, and the balance to be spent, we also applied not yet forged funds to the recipient(in pool only), allowing users to send "future" spend transactions to the pool and thus allowing chaining of transfers tx1:A->B, tx2:B->C to the pool, meaning even if tx1 was not forged, we allowed tx2 to enter the pool. This caused a ripple effect when network was spamming since the state begins to differ and forger could produce invalid blocks.

Due to this we changed some of the logic and rules of AIP17, where we now only apply transaction to the sender - when entering pool, and to recipient when block is forged (if present in pool, and transaction not in pool).

This should also close edge cases on possible spamming of network with address tumbling or just cold wallet spending in general.

## Types of changes

## Checklist

- [X] I have read the [CONTRIBUTING](https://docs.ark.io/developers/guidelines/contributing.html) documentation
- [X] I have added necessary documentation (if appropriate)


packages/core-transaction-pool-mem/lib/connection.js
```diff
@@ -246,8 +266,7 @@ class TransactionPool extends TransactionPoolInterface {
         fetchSize,
       )) {
         const transaction = this.mem.getTransactionById(id)
-
-        if (transaction && this.checkApplyToBlockchain(transaction)) {
+        if (transaction) {
```
vasild: If this is removed and I think transaction will always be defined, then the current function `getTransactionsForForging()` can be simplified to `this.getTransactionIdsForForging(0, blockSize).map(id => this.mem.getTransactionById(id).serialized)`

packages/core-transaction-pool-mem/lib/mem.js
```diff
@@ -154,6 +171,12 @@ class Mem {
       delete this.bySender[senderPublicKey]
     }
 
+    const recipientId = memPoolTransaction.transaction.recipientId
+    this.byRecipient[recipientId].delete(memPoolTransaction)
+    if (this.byRecipient[recipientId].size === 0) {
+      delete this.byRecipient[recipientId]
+    }
+
```
vasild: Removal happens in reverse order to addition. E.g. add to member1, add to member2, add to member3. Remove from member3, remove from member2, remove from member1. That is, move this new code to line 168.

packages/core-transaction-pool/lib/guard.js
```diff
@@ -78,22 +78,6 @@ module.exports = class TransactionGuard {
...
-    // NOTE: will be refactored in 2.1
-    transactions = transactions.filter(transaction => {
-      const errors = []
-      if (!this.pool.walletManager.canApply(transaction, errors)) {
-        this.__pushError(transaction, 'ERR_APPLY', JSON.stringify(errors))
-        return false
-      }
-
-      return true
-    })
```
vasild: Removal of this code would reintroduce the bug it was fixing, described in the comment?

packages/core-transaction-pool/lib/interface.js
```diff
@@ -53,6 +53,15 @@ module.exports = class TransactionPoolInterface {
     throw new Error('Method [getSenderSize] not implemented!')
   }
 
+  /**
+   * Get the number of transaction in the pool for specific recipient
```
vasild: s/transaction/transactions/

packages/core-transaction-pool/lib/interface.js
```diff
@@ -209,34 +219,51 @@ module.exports = class TransactionPoolInterface {
...
-        senderWallet &&
-        senderWallet.balance === 0 &&
-        this.getSenderSize(senderPublicKey) === 0
-      ) {
-        this.walletManager.deleteWallet(senderPublicKey)
+      if (exists) {
+        this.removeTransaction(transaction)
+      } else if (senderWallet) {
+        const errors = []
+        if (senderWallet.canApply(transaction, errors)) {
```
vasild: I think there is on requirement for transactions in a block to be in any specific order. If this is the case, then we should not be upset here if we cannot apply a given tx. Maybe another one in the block would give that sender enough funds.

packages/core-transaction-pool/lib/pool-wallet-manager.js
```diff
@@ -63,125 +63,91 @@ module.exports = class PoolWalletManager extends WalletManager {
...
-    return transaction
+    return errors.length === 0
   }
 
   /**
-   * Apply the given block to a delegate in the pool wallet manager.
-   * We apply only the block reward and fees, as transaction are already be applied
-   * when entering the pool. Applying only if delegate wallet is in pool wallet manager
-   * @param {block}
+   * Remove the given transaction from a a sender only.
```
vasild: from a a sender

packages/core-transaction-pool/lib/pool-wallet-manager.js
```diff
@@ -63,125 +63,91 @@ module.exports = class PoolWalletManager extends WalletManager {
...
+   * @return {Transaction}
    */
-  applyPoolBlock(block) {
-    // if delegate in poll wallet manager - apply rewards
-    if (this.exists(block.data.generatorPublicKey)) {
-      const delegateWallet = this.findByPublicKey(block.data.generatorPublicKey)
-      delegateWallet.applyBlock(block.data)
-    }
-  }
+  revertTransaction(transaction) {
```
vasild: "revertTransaction" is misleading if we are going to revert only the sender balance.

packages/core-transaction-pool/lib/pool-wallet-manager.js
```diff
@@ -63,125 +63,91 @@ module.exports = class PoolWalletManager extends WalletManager {
...
+    const { type, data } = transaction
+    const sender = this.findByPublicKey(data.senderPublicKey) // Should exist
+    const recipient = this.byAddress[data.recipientId]
```
vasild: This variable is not used.

packages/core-transaction-pool-mem/lib/connection.js
```diff
@@ -80,6 +80,17 @@ class TransactionPool extends TransactionPoolInterface {
     return this.mem.getBySender(senderPublicKey).size
   }
 
+  /**
+   * Get the number of transactions in the pool for a specific recipient
+   * @param {String} recipientId
+   * @returns {Number}
+   */
+  getRecipientSize(recipientId) {
```
vasild: This method is not used anywhere.

packages/core-transaction-pool-mem/lib/connection.js
```diff
@@ -38,16 +38,24 @@ class TransactionPool extends TransactionPoolInterface {
...
+    await this.removeForgedTransactions()
+
+    return this
+  }
+
+  /*
+  *Remove transactions that were forged while we were offline.
+   * @return {void}
+
+  */
```
vasild: whitespace messup

packages/core-transaction-pool/__tests__/interface.test.js
```diff
@@ -185,6 +185,20 @@ describe('Transaction Pool Interface', () => {
...
+    it('should be a function', () => {
+      expect(poolInterface.removeForgedTransactions).toBeFunction()
+    })
+
+    it('should throw an exception', async () => {
+      await expect(
+        poolInterface.removeForgedTransactions(),
+      ).rejects.toThrowError(
+        'Method [removeForgedTransactions] not implemented!',
+      )
```
vasild: Use `expect(poolInterface.removeForgedTransactions).toThrow('Method [removeForgedTransactions] not implemented!')`.

packages/core-transaction-pool-mem/lib/connection.js
```diff
@@ -38,16 +38,24 @@ class TransactionPool extends TransactionPoolInterface {
 
     this.__purgeExpired()
 
-    // Remove transactions that were forged while we were offline.
-    const allIds = all.map(
```
vasild: Why move to a separate function? It is called in just one place. This patch removes 4 lines and adds 41!

packages/core-transaction-pool-mem/__tests__/connection.test.js
```diff
@@ -139,8 +206,8 @@ describe('Connection', () => {
       transactions[transactions.length - 1].expiration = expiration
 
       transactions.push(new Transaction(mockData.dummy1))
-      transactions[transactions.length - 1].type =
-        TRANSACTION_TYPES.TIMELOCK_TRANSFER
+      // transactions[transactions.length - 1].type =
+      //   TRANSACTION_TYPES.TIMELOCK_TRANSFER
```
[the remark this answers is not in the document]
kristjank: not yet active in 2.0

packages/core-transaction-pool/lib/guard.js
kristjank: The whole PR is around this bug, as B->C cant enter the pool anymore until A->B forged

packages/core-transaction-pool/lib/interface.js
kristjank: Not true. Block is already applied and forged before we enter this function and the definitive and true state of the chain. Here we apply that state if needed, and if not successful the senders must go out of pool.

Order is important for other things, vote/unvote etc, but not related to this.

vasild:
This change looks overly complicated to me. What problem exactly is it trying to solve? Is it this one:

> ... allowing chaining of transfers tx1:A->B, tx2:B->C to the pool, meaning even if tx1 was not forged, we allowed tx2 to enter the pool. This caused a ripple effect when network was spamming since the state begins to differ and forger could produce invalid blocks.

If yes, then I think it should be fixable by a much simpler patch, modifying solely the forger - to make sure it produces valid blocks.

packages/core-transaction-pool-mem/lib/connection.js
kristjank: Something like the last commit, still testing on forger

kristjank: changes something, but can be played later on...

packages/core-transaction-pool-mem/lib/mem.js
kristjank: ok please check

packages/core-transaction-pool-mem/__tests__/connection.test.js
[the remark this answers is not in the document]
vasild: I mean why is it necessary to comment it in this patch? Did it start failing due to changes in this patch?

packages/core-transaction-pool-mem/lib/connection.js
vasild: > the current function `getTransactionsForForging()` can be simplified to `this.getTransactionIdsForForging(0, blockSize).map(id => this.mem.getTransactionById(id).serialized)`

I mean that the entire body of the function can become 1 line.

packages/core-transaction-pool/lib/guard.js
vasild: I could have missed something but I think after this patch we have the following problem:

* New transaction tx1 A->B is received and at that time A does not have enough funds, so it is not allowed to enter the pool, but is however added to the cache.
* At a later time A receives funds by some other transactions, which are also forged.
* We receive tx1 again. This time A has enough funds. However we reject tx1 because it is in the cache.

packages/core-transaction-pool-mem/__tests__/connection.test.js
```diff
@@ -124,6 +162,35 @@ describe('Connection', () => {
...
+        mockData.dummy4,
+        mockData.dummy5,
+        mockData.dummy6,
+      ]
+
+      container
+        .resolvePlugin('database')
+        .walletManager.findByPublicKey(
+          '000000000000000000000000000000000000000420000000000000000000000000',
+        )
```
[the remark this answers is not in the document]
supaiku0: The wallet is otherwise a cold wallet, so by calling `findByPublicKey` on the database wallet manager it makes sure the test can work as intended.

packages/core-transaction-pool/lib/guard.js
zillionn: How about when you receive a new block with transactions, extract all senders and receivers and delete them from the cache. Is this even possible or you're caching only the txid?

vasild: Only ids :/

packages/core-transaction-pool-mem/__tests__/connection.test.js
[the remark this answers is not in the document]
supaiku0: Yes, the tx validation would fail (it is not added to the pool), as we don't have the new tx types enabled yet.

PR #1655, opened 2018-12-05
perf(core-api): Use a faster alternative to derive an estimate

vasild:
## Types of changes

## Checklist


faustbrian:
I would look at https://wiki.postgresql.org/wiki/Count_estimate and create a function via migration to then wrap queries in it to get the proper estimate that is based on the full query with all conditions.

vasild:
> I would look at https://wiki.postgresql.org/wiki/Count_estimate and create a function via migration to then wrap queries in it to get the proper estimate that is based on the full query with all conditions.

That would be one solution. However in some cases we can derive the exact result very fast and thus don't need to bother with estimates. One such is:
`SELECT COUNT(*) FROM blocks TABLESAMPLE SYSTEM (100) WHERE height = ...`
This one randomly selects 100% of the rows in the table, then searches inside this random pile (indexes invalidated) for a row with the given height. If instead we use
`SELECT COUNT(*) FROM blocks WHERE height = ...`
it would use the unique index on `height` and come up with the result "fast" (exact result, not an estimate).

So, I am looking into how to do that.

packages/core-api/__tests__/v2/handlers/transactions.test.js
```diff
@@ -168,7 +168,6 @@ describe('API 2.0 - Transactions', () => {
         expect(response.data.data).toBeArray()
 
         expect(response.data.data).toHaveLength(100)
-        expect(response.data.meta.totalCount).toBe(153)
```
vasild: Notice: that count is supposed to be an estimate. So we may return 100 rows, totalCount=100, but however there may be e.g. 110 rows in the database.

faustbrian:
@vasild This can now be adjusted to TypeScript on the `develop` branch so we can include it in the next release.

PR #1625, opened 2018-12-04
refactor: migrate to typescript

faustbrian:
## Proposed changes

- Resolves https://github.com/ArkEcosystem/core/milestone/3
- Fixes a bug in `core-container` and `resolveOptions` that could result in null resolutions or providing options for plugins that are not even registered in the container
- Fixed some return and value assignment issues where for example a `BigNumber` object was expected but a `Number` was assigned **(JS didn't catch this and we had some tests that were passing even though they should have been failing)**
- Moved the `peers.json` config into the defaults of `core-p2p` as they are only used there and not by anything else (https://github.com/ArkEcosystem/core/pull/1625/commits/dc582feb0467d63b7ae2f9da277051ffe25cf51b). This means `peers.json` is now solely used to store the peer lists, more changes will follow in an upcoming PR.
- Resolved various minor bugs that showed up during the migration
- Added the `yarn setup` command to install, bootstrap and build all packages
- Merged `core-transaction-pool` and `core-transaction-pool-mem`

Have been forging for some time with this branch on devnet now and had no errors occur so far.

## Types of changes

- [x] Refactoring (improve a current implementation without adding a new feature or fixing a bug)
- [x] Build (changes that affect the build system)

## Checklist

- [x] I have read the [CONTRIBUTING](https://docs.ark.io/developers/guidelines/contributing.html) documentation
- [x] Lint and unit tests pass locally with my changes

PR #1670, opened 2018-12-06
fix: reply, on node configuration API endpoint, with integer ports

j-a-m-l:
## Proposed changes
Devnet and Mainnet were returning a different type on the node configuration endpoint (http://167.114.29.54:4003/api/v2/node/configuration). It may be caused by loading the ports from the environment, configuration files, or other causes.

This PR enforces that, independently of the cause, the API ports of the response, are always integers.

## Types of changes

## Checklist

### DIFF
--- a/packages/core-api/__tests__/v2/handlers/node.test.js
+++ b/packages/core-api/__tests__/v2/handlers/node.test.js
@@ -61,6 +61,10 @@ describe('API 2.0 - Loader', () => {
         expect(response.data.data.symbol).toBeString()
         expect(response.data.data.explorer).toBeString()
         expect(response.data.data.version).toBeNumber()
+
+        Object.values(response.data.data.ports).forEach(port => {
+          expect(port).toBeNumber()
+        })
       })
     })
   })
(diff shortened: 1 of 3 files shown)


faustbrian:
Should be done in the transformer rather then method. Will be fixed anyways in https://github.com/ArkEcosystem/core/pull/1625,

j-a-m-l:
Reopen to fix the problem while the TypeScript migration is being done.

faustbrian:
The typescript migration will be done mid next week and there won't be a new release before that for such a small change.

j-a-m-l:
So, should we stop working on the current PRs or create new ones?

faustbrian:
Would be advised unless it is a critical bug fix or security issue, that is why all other PRs are open and not getting merged.

j-a-m-l:
Wouldn't it be better to include this small fix anyway? If one of those PR is merged because it's a critical fix, we would fix both problems.

faustbrian:
It won't be merged until TS is done next week so just leave it closed.

PR #1676, opened 2018-12-08
refactor(core-api): lower blocks index timeout

dated:
## Proposed changes

When fetching the latest block the explorer would occasionally fetch the wrong cached block.

## Types of changes

## Checklist

### DIFF
--- a/packages/core-api/src/versions/2/blocks/methods.ts
+++ b/packages/core-api/src/versions/2/blocks/methods.ts
@@ -50,7 +50,7 @@ const search = async request => {
 export function registerMethods(server) {
   server.method("v2.blocks.index", index, {
     cache: {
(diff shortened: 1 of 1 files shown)


packages/core-api/src/versions/2/blocks/methods.ts
```diff
@@ -50,7 +50,7 @@ const search = async request => {
 export function registerMethods(server) {
   server.method("v2.blocks.index", index, {
     cache: {
-      expiresIn: 8 * 1000,
+      expiresIn: 4 * 1000,
```
faustbrian: I would set it to 6 seconds, that still gives 2 seconds room.

dated: should be enough, i agree.

PR #1713, opened 2018-12-13
fix: prevent the list of peers to go become too short

vasild:
* Reschedule a call to updateNetworkStatusIfNotEnoughPeers() at the end
  of updateNetworkStatus(), even if successful.

* Add peers from config.peers.list if the list is too short

* Remove loop by recursion in updateNetworkStatus()

* Remove loop by recursion in discoverPeers() and cycle through the
  peers in deterministic fashion after shuffling them. The previous
  implementation would pick one random peer and if unsuccessful, then
  pick random peer and retry, but this way we could end up retrying with
  the same peer over and over again. For example if 70% of the peers
  result in failure and 30% not, we could loop a lot of times over the
  bad portion of the peers.

## Types of changes

## Checklist


packages/core-p2p/lib/monitor.js
```diff
@@ -413,25 +399,29 @@ class Monitor {
...
-
-      peers.forEach(peer => {
-        if (
-          Peer.isOk(peer) &&
-          !this.getPeer(peer.ip) &&
-          !this.guard.isMyself(peer)
-        ) {
-          this.__addPeer(peer)
+    const shuffledPeers = shuffle(this.getPeers())
+    const nNewPeersNeeded = 4
```
fix: why 4?

packages/core-p2p/lib/monitor.js
```diff
@@ -103,47 +103,33 @@ class Monitor {
...
     if (this.config.disableDiscovery) {
-      logger.warn(
-        'Skipped peer discovery because the relay is in non-discovery mode.',
-      )
+      logger.warn('Skipped peer discovery because the relay is in non-discovery mode.')
       return
     }
 
     try {
-      if (process.env.ARK_ENV !== 'test') {
```
supaiku0: I think we need to keep this check.

packages/core-p2p/lib/monitor.js
```diff
@@ -103,47 +103,33 @@ class Monitor {
...
-      if (process.env.NODE_ENV !== 'test') {
-        await delay(5000)
-      }
+      nextRunDelaySeconds = 5
 
-      return this.updateNetworkStatus()
+      logger.info(`Couldn't find enough peers, trying again in ${nextRunDelaySeconds} seconds`)
     }
+
+    setTimeout(() => await this.updateNetworkStatusIfNotEnoughPeers(), nextRunDelaySeconds * 1000)
```
supaiku0: This also needs to be wrapped into the check from above. Unit tests are not supposed to to call it.

supaiku0: Or wait, probably wrong what I just said. You are missing an `async` .

packages/core-p2p/lib/monitor.js
faustbrian: Unit tests run with the testnet config which has 0/1 peers so this will loop forever, that is why the check exists there.

packages/core-p2p/lib/monitor.js
vasild: It was an arbitrary number to fix this deficiency in `discoverPeers()` before the patch:
`discoverPeers()` would be satisfied as long as it fetches all peers from some of its peers. E.g.
ME <-> P (P is my peer). If I fetch all P's peers then it is ok and `discoverPeers()` considers the job done and quits. However it could be that the list of P's peers is quite short - maybe just 1 or 2 entries.

To fix this I added this magic number "4" - at least 4 new peers must be added, even if we have to cycle through more than one of our peers (e.g. fetch 2 from P and 2 others from Q).

However, this can be done better - I changed the code to consider the job done and quit once `this.hasMinimumPeers()` becomes true.

packages/core-p2p/lib/monitor.js
vasild: I did not mean to disable the check. I noticed that there is such a check already in the very beginning of `updateNetworkStatus()`: `if (test) return`. Which means there is no need to do this check a second time. This is why I removed it.

Now I noticed that the "is test" check at the beginning of `updateNetworkStatus()` is `process.env.NODE_ENV === "test"` whereas later we checked `process.env.ARK_ENV`. I think this must have been a type and it should be `ARK_ENV`. Updated in the patch.

packages/core-p2p/lib/monitor.js
vasild: > This also needs to be wrapped into the check from above. Unit tests are not supposed to call it.

No because we do `return;` if we detect a test environment at the start of the function.

> You are missing an async .

Where?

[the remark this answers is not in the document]
vasild: Changed to 10 min.

PR #1731, opened 2018-12-14
fix(core-database): delegate ordering

dated:
## Proposed changes

Ordering on the delegates API does not seem to work. This PR adjusts the delegates `findAll` method so that is consistent with the respective method for wallets.

## Types of changes

## Checklist


faustbrian:
@dated could you add a test that ensures the ordering works with the default and custom values.

Issue #1782, opened 2018-12-16
Mark __ methods as private and rework tests for them


faustbrian:
Previously we used the `__` prefix as an indicator for `private` methods in JavaScript as it didn't have a concept of visibility. The result of that methodology is that we have a lot of flawed tests that odd naming conventions.

What I mean by flawed tests is that we are testing private methods which is a bad practice. Classes should always be tested in a way a developer is going to use the public API it exposes which means there is no access to those methods and their correct behaviour should be asserted through the public methods that call them.

PR #1825, opened 2018-12-16
[WIP] refactor: mark private methods and remove them from tests

JeremiGendron:
#1782 

## Proposed changes

As described in issue, initially sifted through tests for `(?!__support__|__fixtures__|__stubs)(__[a-z0-9]*\()` under `packages/**/__tests__/**/*.test.ts` and remove those units, their related then unused imports and assignments in `beforeAll()` using those imports.

Now, will run the same query under `packages/**/src/**/*.ts` and mark all `__methods()` with the private prefix.

## Types of changes

## Checklist


faustbrian:
Simply removing those tests serves no purpose other then making the code less tested, they need to be replaced with other tests that do not lower the coverage but cover the behaviour of the private methods through the public API.

`Classes should always be tested in a way a developer is going to use the public API it exposes which means there is no access to those methods and their correct behaviour should be asserted through the public methods that call them.`

As this requires attention to detail and knowledge about how things work together to not lower the coverage I will close this and assign team members to the task.

PR #1831, opened 2018-12-17
fix(core-database): ensure that delegate search deals with "undefined" as username

j-a-m-l:
## Proposed changes
Searching delegates without parameters return the delegate with username "undefined" when it should not return anything.

## Types of changes

## Checklist


j-a-m-l:
Other search endpoints return all the results when no parameter is passed, so I've updated this fix to make it consistent

PR #1833, opened 2018-12-17
Package up core-logger types

paroxysm:
## Proposed changes
Addresses #1807
Exports core-logger types to introduce type-safety to modules that depend on it.
According to https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html
We have two options, we can package up the types in the module itself leveraging https://basarat.gitbooks.io/typescript/docs/tips/barrel.html to export all types that other dependencies might reference OR publish a separate @types/<ark_module> to npm for each module. The former just works and resolves locally(w/o having to go through NPM to resolve things) so we can adopt that for the rest of the modules.

## Types of changes

## Checklist

<!--

### DIFF
--- a/packages/core-logger-winston/src/index.ts
+++ b/packages/core-logger-winston/src/index.ts
@@ -1,3 +1,4 @@
+import { LogManager } from "@arkecosystem/core-logger";
 import { defaults } from "./defaults";
 import { Logger } from "./driver";
 
@@ -7,8 +8,7 @@ export const plugin = {
     alias: "logger",
     extends: "@arkecosystem/core-logger",
     async register(container, options) {
-        const logManager = container.resolvePlugin("logManager");
-        // @ts-ignore
+        const logManager: LogManager = container.resolvePlugin("logManager");
         await logManager.makeDriver(new Logger(options));
 
         return logManager.driver();
--- a/packages/core-logger/package.json
+++ b/packages/core-logger/package.json
@@ -6,8 +6,8 @@
         "Brian Faust <brian@ark.io>"
     ],
     "license": "MIT",
-    "main": "index",
-    "types": "index",
+    "main": "dist/index",
+    "types": "dist/index",
     "files": [
         "dist"
     ],
--- a/packages/core-logger/package.json
+++ b/packages/core-logger/package.json
@@ -6,7 +6,8 @@
         "Brian Faust <brian@ark.io>"
     ],
     "license": "MIT",
-    "main": "dist/index.js",
+    "main": "dist/index",
+    "types": "index",
     "files": [
         "dist"
     ],
--- a/packages/core-logger/package.json
+++ b/packages/core-logger/package.json
@@ -6,7 +6,7 @@
         "Brian Faust <brian@ark.io>"
     ],
     "license": "MIT",
-    "main": "dist/index",
+    "main": "index",
     "types": "index",
     "files": [
         "dist"
--- a/.gitignore
+++ b/.gitignore
@@ -62,3 +62,6 @@ packages/**/dist/
 
 # Random
 peers_backup.json
+
+#Webstorm/Intellij
+.idea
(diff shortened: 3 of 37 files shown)


packages/core-logger-winston/src/index.ts
```diff
@@ -7,8 +8,7 @@ export const plugin = {
     alias: "logger",
     extends: "@arkecosystem/core-logger",
     async register(container, options) {
-        const logManager = container.resolvePlugin("logManager");
-        // @ts-ignore
+        const logManager = container.resolvePlugin("logManager") as LogManager;
```
faustbrian: `const logManager: LogManager = container.resolvePlugin("logManager");`

packages/core-logger/package.json
```diff
@@ -6,7 +6,8 @@
         "Brian Faust <brian@ark.io>"
     ],
     "license": "MIT",
-    "main": "dist/index.js",
+    "main": "dist/index",
+    "types": "dist/index",
```
faustbrian: A proper type definition should be created instead of just using the `dist` files.

paroxysm: Typescript already produces declarations for us, but barrel index.ts in the root folder of each module will work too.
So rather than "dist/index.d.ts"
I'll create an "./index.ts" that uses the barrel pattern to export the types. Will that work?

paroxysm: Done. Can I safely assume that this will be the pattern to adopt across all modules?

faustbrian:
Tests are failing because of missing exports.

packages/core-logger/package.json
paroxysm: I'm getting strange issues using the approach. it doesn't like a ./index.d.ts exporting types, which cause some test failures since it can't resolve the types.

However, if I export the types via ./src/index.ts, then the tests run. /src/index.ts automatically generates a /dist/index.d.ts. Since this works, is there any reason why we can't use 'types: dist/index'?

paroxysm: Nevermind, I've got it working. Will be committing shortly..

paroxysm:
Tests are now passing, the issue was that I hadn't included `./index.ts` to be compiled in `tsconfig.json`

faustbrian:
Have you tried starting a relay after those changes?

packages/core-logger/package.json
```diff
@@ -6,7 +6,8 @@
         "Brian Faust <brian@ark.io>"
     ],
     "license": "MIT",
-    "main": "dist/index.js",
+    "main": "index",
```
faustbrian: This will result in an error that the module cannot be found if you try to start a relay as the compiled files are not what is being served for an import.

paroxysm: Yes, this is because now the resulting compiled files are in `/dist/src` as opposed to `/dist` previously. This is because we're now compiling `./index.ts`, ts is preserving those folder structures when outputting to `/dist`
This causes that line `pkg : require('../package.json')` to fail because the file is now located in `/dist/src/index.js` as opposed to `/dist/index.js`.

It's fixable by going up 1 more level, so `pkg : require('../../package.json')` but it just reads weirdly...

paroxysm: We can also impose a new pattern entirely, where in each module, the `/src/index.ts` is a barrel file for exporting types. Then we rename the file that exports the plugin descriptor object to `plugin.ts`
That way, leave the `tsconfig.js` unchanged. All compiled sources explode into `/dist` instead of `dist/src` and it still works because the `plugin.ts` is still exporting the `plugin` property

paroxysm: I've applied the changes I suggeste @faustbrian, the relay now runs and all tests pass.
Lemme know if you're cool w/ this new pattern. The alternative is to combine both the barrel pattern and the existing code for exporting the plugin object into one `index.ts` but I wanted to keep the two separate

faustbrian:
Looks ok for now, ultimately I would aim for proper declaration files though.

paroxysm:
@faustbrian Based on the typescript compiler options that we're inheriting from `@sindresorhus/tsconfig`. Declaration files are being automatically generated. So if we we're adopting a barrel file pattern, the corresponding .d.ts is generated for us.
This frees us from the manually maintaining it. If we need to make tweaks, we can tweak the barrel file to exclude/include types

faustbrian:
Feel free to send PRs for other packages, the foreseeable future that solution is good enough.

paroxysm:
Okay cool, I wasn't sure whether it was an acceptable pattern to start applying, thanks for the go-ahead. I'll proceed with the others!

PR #1873, opened 2018-12-20
refactor: add minimum of minFeeBroadcast for getFeeStatistics query

### LINKED ISSUE #1741: Default feestatistics should match default node settings
[getFeestatistics()](https://github.com/ArkEcosystem/core/blob/fad5a259b1b1c074e7cf35d8279371ac78a47062/packages/core-graphql/src/repositories/transactions.ts#L257) should have a minimum for the query for fees. This would reflect a non-adjusted node's fee structure with offset. Currently this includes delegates with modified fee structures self-packaging transactions in their blocks, which drags the minimum down and probably the average. The current model reflects on delegates as they have a very high throughput on the network, however only they are able to make these transactions. I'm not sure how feestatistics are useful for delegates, because of the aforementioned method.

vulet:
## Proposed changes 
Resolves #1741.

## Types of changes

## Checklist

### DIFF
--- a/packages/core-api/src/repositories/transactions.ts
+++ b/packages/core-api/src/repositories/transactions.ts
@@ -261,7 +261,7 @@ export class TransactionsRepository extends Repository implements IRepository {
             .from(this.query)
             // @ts-ignore
             .where(this.query.timestamp.gte(slots.getTime(dayjs().subtract(30, "days"))))
-            .and(this.query.fee.gte(minFeeBroadcast))
+            .and(this.query.fee.gte(this.transactionPool.options.dynamicFees.minFeeBroadcast))
             .group(this.query.type)
             .order('"timestamp" DESC');
 
--- a/packages/core-api/src/repositories/transactions.ts
+++ b/packages/core-api/src/repositories/transactions.ts
@@ -1,5 +1,6 @@
 import { app } from "@arkecosystem/core-container";
 import { constants, slots } from "@arkecosystem/crypto";
+import { config as localConfig } from "@arkecosystem/core-transaction-pool";
 import dayjs from "dayjs-ext";
 import { IRepository } from "../interfaces/repository";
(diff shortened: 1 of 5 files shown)


packages/core-graphql/package.json
```diff
@@ -29,6 +29,7 @@
     "dependencies": {
         "@arkecosystem/core-container": "^2.1.0",
         "@arkecosystem/core-http-utils": "^2.1.0",
+        "@arkecosystem/core-transaction-pool": "^2.1.0",
```
faustbrian: Packages itself should never have a dependency on anything but interfaces and the container.

packages/core-api/src/repositories/transactions.ts
```diff
@@ -248,6 +249,7 @@ export class TransactionsRepository extends Repository implements IRepository {
      * @return {Object}
      */
     public async getFeeStatistics(): Promise<any> {
+        const minFeeBroadcast = localConfig.get("dynamicFees.minFeeBroadcast");
```
faustbrian: This should be resolved from the container, those config objects are only for plugin internal use.

packages/core-graphql/src/repositories/transactions.ts
```diff
@@ -255,6 +256,7 @@ class TransactionsRepository extends Repository {
      * @return {Object}
      */
     public async getFeeStatistics() {
+        const minFeeBroadcast = localConfig.get("dynamicFees.minFeeBroadcast");
```
faustbrian: This should be resolved from the container, those config objects are only for plugin internal use.

vulet:
I see, something like this?

boldninja:
@vulet Can you provide me with your ARK address for GitHub bounty? You can also contact me on ARK's Slack - https://ark.io/slack (username : boldninja) and direct message me your address there.

PR #1887, opened 2018-12-20
chore: export core-logger-winston types.

paroxysm:
## Proposed changes
Further addresses #1795. While we're exporting the `Logger` type defined in this module. The rest of the modules will only depend on `core-logger` module for its `AbstractLogger` type. This ensures that, even though we're introducing a `core-logger` dependency to all modules that need to log(so as to benefit from type-safety), we still retain the ability to supplant the `core-logger-winston` implementation with a different implementation as long as that implementation extends from `core-logger`'s `AbstractLogger` 

i.e. Override the winston logger with a logstash json encoded logger or RDMS logger or even a network streaming logger. As long as all these extend from the `AbstractLogger` in `core-logger` module, they still benefit from the type-safety but don't restrict the rest of the modules to one specific implementation.

## Types of changes

## Checklist

### DIFF
--- a/packages/core-logger/src/logger.ts
+++ b/packages/core-logger/src/logger.ts
@@ -60,7 +60,13 @@ export abstract class AbstractLogger {
      * @param  {Number} figures
      * @return {void}
      */
-    public abstract printTracker(title: string, current: number, max: number, postTitle: string, figures: number): void;
+    public abstract printTracker(
+        title: string,
+        current: number,
+        max: number,
+        postTitle: string,
+        figures?: number,
+    ): void;
 
     /**
      * Stop the progress tracker.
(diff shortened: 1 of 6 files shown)


packages/core-logger/src/logger.ts
```diff
@@ -60,7 +60,13 @@ export abstract class AbstractLogger {
      * @param  {Number} figures
      * @return {void}
      */
-    public abstract printTracker(title: string, current: number, max: number, postTitle: string, figures: number): void;
+    public abstract printTracker(
```
faustbrian: Did `tslint` format it like that?

paroxysm: Yes :(

faustbrian: Odd, never yet complained to me about that line.

PR #1891, opened 2018-12-20
Export & use core-container types

paroxysm:
## Proposed changes
Address #1789 and also further propagate `core-logger`'s `AbstractLogger` type usage.
We're leveraging ts generics to be able to retrieve typed objects directly from the `Container` class.
For the time-being, the generic is defaulted to `any` to prevent compilation issues, but in the future, as we expose more module typedefs, we can remove that default to enforce that we always get typed values from the `Container`

We're also introducing `core-logger` as a dependency to every module that was previously grabbing a reference via `container.resolvePlugin('logger')`

I'm trying to be careful not to introduce a circular dependency. I know there can definitely be one in `core-container` and `core-logger` when they both try to import each other's types. For this edge-case we can break the pattern by having `core-logger` not import `core-container`'s typedefs but it's not that big of an issue.

## Types of changes

## Checklist

### DIFF
--- a/packages/core-api/package.json
+++ b/packages/core-api/package.json
@@ -29,6 +29,7 @@
     },
     "dependencies": {
         "@arkecosystem/core-container": "^2.1.0",
+        "@arkecosystem/core-logger": "^2.1.0",
         "@arkecosystem/core-http-utils": "^2.1.0",
         "@arkecosystem/core-transaction-pool": "^2.1.0",
         "@arkecosystem/core-utils": "^2.1.0",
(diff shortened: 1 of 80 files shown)


paroxysm:
Looking into them..

paroxysm:
@faustbrian  These failures don't appear to have anything to do w/ my changes. There is brittleness in there somewhere that's surfacing now. The code that was merged doesn't even affect the module whose tests are failing

paroxysm:
@faustbrian I can reproduce these failures on the develop branch w/o the changes in this PR. Seems something else recent is causing these failures.

paroxysm:
It's strange tho, there should've been a 100% failure across all CI runs but only a few failed while some passed....

faustbrian:
@paroxysm you can use https://help.github.com/articles/closing-issues-using-keywords/ if you want to automatically close issues as resolved.

Instead of `Address #1789` write `Resolves #1789`.

paroxysm:
Thanks, that's really nifty @faustbrian!

PR #1900, opened 2018-12-22
chore: format arktoshi

zillionn:
## Proposed changes
Very often I have hard time reading the amount/fee in the logs, for example:
`0.00765 Ѧ`
I'm not sure is this `765,000` arktoshi or is it `76,500`?! Also to figure this out, you must know the network's decimal places.

IMO it should be printed like this:
`Ѧ0.00765000`

## Types of changes

## Checklist

### DIFF
--- a/packages/core-tester-cli/__tests__/commands/command.test.ts
+++ b/packages/core-tester-cli/__tests__/commands/command.test.ts
@@ -309,10 +309,10 @@ describe("Command Base", () => {
 
     describe("static __arktoshiToArk", () => {
         it("should give ark", () => {
-            expect(Command.__arktoshiToArk(1)).toBe("0.00000001 DѦ");
-            expect(Command.__arktoshiToArk(10000000)).toBe("0.1 DѦ");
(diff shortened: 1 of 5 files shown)


zillionn:
Hm, is there an easy way to hide the zeros if it's a whole number when using `toLocaleString .minimumFractionDigits: 8`? Because `765000` becomes `0.00765000` but `10000000000` also becomes `10.00000000`.

paroxysm:
Maybe you can print the Arktoshi value a long side it to help? I don't see how it's confusing.

zillionn:
@paroxysm Currently it's printer like this `0.000007`. How much "toshi" is this?

paroxysm:
I'm not reading that in arktoshi, I'm reading that in Ark. If I wanted the Arktoshi value, I'd multiply that by 1e8. I'm saying seeing the value as Ark is the standard, but you can add supplementary info to display in Arktoshi, but keep the standard.

zillionn:
You don't understand me, but I guess nobody else agrees with me so I'm closing this.

Issue #1573, opened 2018-12-04
Revisit core-database structure and expandability


faustbrian:
Right now the core-database and core-database-postgresql packages have a lot of shared logic that should only be in core-database and never be overwritten or touched unless someone really knows what they are doing.

Revisit these pieces of code and refactor where needed to reduce complexity.

faustbrian:
Ditching in favour of https://github.com/ArkEcosystem/core/issues/2006.

PR #1905, opened 2018-12-24
Export and use core-database-postgres typedefs

### LINKED ISSUE #1791: Add core-database-postgres type definitions

paroxysm:
## Proposed changes
resolves #1791

This refactor has exposed some strange design decisions made. There's some methods that are on the postgres implementation of the `Connection` interface that can be moved into an abstract class.
PS. We probably should break out the business logic in `Connection` out into Services that leverage the connection and repositories so we're separating concerns

## Types of changes

## Checklist

<!--

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -73,8 +73,8 @@ jobs:
           name: core-logger-winston
           command: 'cd ~/ark-core/packages/core-logger-winston && yarn test:coverage'
       - run:
-          name: core-graphql
-          command: 'cd ~/ark-core/packages/core-graphql && yarn test:coverage'
+          name: core-api
+          command: 'cd ~/ark-core/packages/core-api && yarn test:coverage'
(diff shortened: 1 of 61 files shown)


faustbrian:
The mentioned issues will be taken core of by https://github.com/ArkEcosystem/core/issues/1573.. There is still a lot of old code in it written by someone else that will refactored.

In the end the `core-database-postgres` package will barely overwrite anything and mostly just provide data through methods to the underlying `core-database` package which will receive handlers for specific types like blocks and transactions.

paroxysm:
@faustbrian Great! There's definitely some tech debt that needs to be addressed

paroxysm:
@faustbrian I've addressed the conflicts.
I purposefully left out core-graphql because it was referencing some methods that the `PostgresConnection` class doesn't define.
I'll need to track those down and formally declare them

PR #1906, opened 2018-12-24
Export and use core-transaction-pool typedefs

### LINKED ISSUE #1793: Add core-transaction-pool type definitions

paroxysm:
## Proposed changes
resolves #1793 

I excluded core-test-utils because adding a dependency a `core-transaction-pool` creates a circular dependency because `core-test-utils` is a devDependency on a couple of other modules(including `core-transaction-pool` itself)

## Types of changes

## Checklist

<!--

### DIFF
--- a/packages/core-api/__tests__/__support__/setup.ts
+++ b/packages/core-api/__tests__/__support__/setup.ts
@@ -1,7 +1,8 @@
 import { app } from "@arkecosystem/core-container";
+import { PostgresConnection } from "@arkecosystem/core-database-postgres";
 import { setUpContainer } from "@arkecosystem/core-test-utils/src/helpers/container";
 
-import { delegates } from "../../../core-test-utils/src/fixtures/testnet/delegates";
+import { delegates } from "../../../core-test-utils/src/fixtures";
 import { generateRound } from "./utils/generate-round";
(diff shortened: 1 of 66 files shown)


faustbrian:
@paroxysm conflicts

packages/core-api/__tests__/__support__/setup.ts
```diff
@@ -20,7 +21,7 @@ async function setUp() {
         ],
     });
 
-    const connection = app.resolvePlugin("database");
+    const connection = app.resolvePlugin<PostgresConnection>("database");
```
faustbrian: All references to `<PostgresConnection>` should point to `<ConnectionInterface>` instead as this reference is made outside of the `core-database-postgres` package which means the `PostgresConnection` reference could be gone at any time and instead be `<SqliteConnection>` for example.

https://github.com/ArkEcosystem/core/blob/develop/packages/core-database/src/interface.ts#L14

paroxysm: Yes, it'll need to be changed. There's a lot of methods and properties being referenced by their deps that are on the `PostgresConnection` class, and not the abstract class.

PR #1916, opened 2018-12-27
refactor(core-p2p): simplify updatePeersOnMissingBlocks()

vasild:
## Types of changes

## Checklist


faustbrian:
I don't really see the benefit of this change as we lose the logging for different scenarios. @supaiku0 thoughts?

vasild:
> I don't really see the benefit of this change as we lose the logging for different scenarios.

Right. And what I wrote was not actually true. Updated the patch.

PR #1939, opened 2019-01-03
fix(core-p2p): malformed condition for filtering peers

vasild:
Currently we would leave in the list all peers that:
"are not me or have invalid port or have invalid version"
the intention must have been:
"are not me and have valid port and have valid version".

This change was first done in 0c2319649 and was later reverted in
109a4d311 with a scarce explanation: "it randomly causes issues". It is
not clear what those issues were.

However the current code is strikingly and obviously wrong - it would
add peers that have invalid port, for example.

So, fix the wrong code and followup by fixing any issues that arise, but
not by resurrecting the wrong code.

## Types of changes

## Checklist

### DIFF
--- a/packages/core-p2p/src/monitor.ts
+++ b/packages/core-p2p/src/monitor.ts
@@ -843,7 +843,7 @@ export class Monitor implements P2P.IMonitor {
         }
 
         const filteredPeers: any[] = Object.values(peers).filter(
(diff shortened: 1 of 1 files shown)


faustbrian:
This caused issues on testnet and seeds where they would endlessly loop because there were no peers after this change, that is why it was reverted.

vasild:
So, on testnet all peers in `peers.json` have invalid port? How do I reproduce that problem?

PR #1950, opened 2019-01-06
 tests(crypto): Increase coverage and mark private methods

JeremiGendron:
## Proposed changes
 - document validate rules param correct types (Object|Function|String)
 - __validateWith<> methods are now private
 - tests for __validateWith<> changed to validate("...", <>)
 - increase coverage for errors() function
 - modify validate to check for "typeof rules === 'string'" (more below)

Refs: #1782 #1775

The validate function was checking for `rules instanceof String`, which never evaluates to true for a string literal. Changed to `typeof rules === 'string'`

## Types of changes

## Checklist


packages/crypto/src/validation/validator.ts
```diff
@@ -108,7 +115,7 @@ export class Validator {
      * @param  {String} rule
      * @return {void}
      */
-    public __validateWithJoi(attributes, rules) {
+    private __validateWithJoi(attributes, rules) {
```
faustbrian: The `__` prefix has to be removed, if you read the issue.

packages/crypto/src/validation/validator.ts
```diff
@@ -76,13 +76,20 @@ export class Validator {
...
+    }
+
     /**
      * Run the validator's rules against its data using a rule.
      * @param  {*} attributes
      * @param  {String} rule
      * @return {void}
      */
-    public __validateWithRule(attributes, rules) {
+    private __validateWithRule(attributes, rules) {
```
faustbrian: The `__` prefix has to be removed, if you read the issue.

packages/crypto/src/validation/validator.ts
```diff
@@ -98,7 +105,7 @@ export class Validator {
      * @param  {String} rule
      * @return {void}
      */
-    public __validateWithFunction(attributes, validate) {
+    private __validateWithFunction(attributes, validate) {
```
faustbrian: The `__` prefix has to be removed, if you read the issue.

packages/crypto/src/validation/validator.ts
```diff
@@ -76,13 +76,20 @@ export class Validator {
         this.rules[name] = implementation;
     }
 
+    /**
+     * Reset any previous results.
+     */
+    public __reset() {
```
faustbrian: Public methods have no prefixes, if you read the issue.

JeremiGendron:
@faustbrian thank you

packages/crypto/src/validation/validator.ts
JeremiGendron: branch updated

faustbrian:
The core-api changes should not be included in this PR.

JeremiGendron:
@faustbrian not sure if `git revert 0abbee01` was the right move

PR #1959, opened 2019-01-09
 feat: Add Windows CI with AppVeyor

ciband:
## Proposed changes
Added AppVeyor configuration to support CI under Windows.

Historically there have been breaking changes on Windows that were unknown due to the vast majority of Ark developers using Linux or MacOS.  Adding an additional Windows CI check will help find these issues earlier.

## Types of changes

- [X] Build (changes that affect the build system)

## Checklist

- [X] I have read the [CONTRIBUTING](https://docs.ark.io/guidebook/contribution-guidelines/contributing.html) documentation

## Further comments
In my AppVeyor account, all the tests run but some do not pass.  I do not know enough about core to properly diagnose to know if these test failures are expected or directly failing as a result of running on Windows.

Open for discussion as to what the next step should be if we want to introduce Windows CI.


faustbrian:
Core in production will not run on Windows systems so I will close this for now but if we ever plan to recommend to run Core on Windows for production environments then we will revisit this.

Issue #1766, opened 2018-12-16
100% coverage for core-logger-winston


faustbrian:

faustbrian:
Just the custom formatter is missing coverage which isn't critical as we can see if it works by starting a node.

PR #1966, opened 2019-01-11
test(core-logger-winston): increase coverage

JeremiGendron:
## Proposed changes

 - move registerTransports to private, remove __
 - add fixture for package as transport option
 - checks if progress === max vs. current (progress set to 100 if > 100; not current)
 - second tracker print/stop tests for diverse arguments
 - now supports passing logger level option
 - fix tests for .error, .warn, .debug
 - add test for .verbose

Refs: #1766

## Types of changes

## Checklist


packages/core-logger-winston/src/driver.ts
```diff
@@ -133,16 +134,14 @@ export class WinstonLogger extends AbstractLogger {
     public suppressConsoleOutput(suppress: boolean): void {
         const consoleTransport = this.logger.transports.find(t => t.name === "console");
 
-        if (consoleTransport) {
-            consoleTransport.silent = suppress;
-        }
+        consoleTransport.silent = suppress;
```
faustbrian: Revert, if the console logger is not available this will throw an error.

packages/core-logger-winston/src/driver.ts
```diff
@@ -15,9 +15,10 @@ export class WinstonLogger extends AbstractLogger {
      * Make the logger instance.
      */
     public make() {
-        this.logger = winston.createLogger();
+        const { level } = this.options;
+        this.logger = winston.createLogger({ level });
```
faustbrian: Revert, levels are set per transport.

JeremiGendron:
Alright

packages/core-logger-winston/__tests__/logger.test.ts
```diff
@@ -78,10 +96,11 @@ describe("Logger", () => {
     describe("stopTracker", () => {
         it("should stop the tracker", () => {
             logger.stopTracker("test_title", 50, 100);
+            logger.stopTracker("second_tracker", 101, 100);
 
             expect(message).toMatch(/test_title/);
             expect(message).toMatch(/=========================/);
-            expect(message).toMatch(/50/);
+            expect(message).toMatch(/100/);
```
faustbrian: There should be a separate test that asserts it doesn't go over 100% instead of altering this test.

packages/core-logger-winston/__tests__/logger.test.ts
```diff
@@ -1,16 +1,23 @@
...
 import "jest-extended";
+import * as winston from "winston";
 import { WinstonLogger } from "../src";
 
 let logger: AbstractLogger;
 let message;
 
 beforeAll(() => {
     const driver = new WinstonLogger({
+        level: "debug",
```
faustbrian: Should be removed according to the previous comment about the log level.

packages/core-logger-winston/__tests__/logger.test.ts
```diff
@@ -1,16 +1,23 @@
...
         transports: [
             {
                 constructor: "Console",
+                package: "@arkecosystem/core-logger-winston/__tests__/__fixtures__/transport",
```
faustbrian: There is no need to use a fixture for that, simply use the winston console transport.

JeremiGendron: thought I already changed it, guess I forgot to push changes! doing so now

Issue #1967, opened 2019-01-13
Missing api tests for core-jest-matchers


JeremiGendron:
**Describe the bug**
There are no tests for the api/ source files in the core-jest-matchers package

**Additional context**
Additionally, there is reduced coverage for most source files.

PR #1968, opened 2019-01-13
test(core-jest-matchers): increase coverage

JeremiGendron:
## Proposed changes

#1967

I've added tests following the existing implementations for the api/ source files.

I've also added or changed existing tests to also check that upon failure the message is as defined in the source files.

## Types of changes

## Checklist

### DIFF
--- a/packages/core-jest-matchers/__tests__/transactions/types/vote.test.ts
+++ b/packages/core-jest-matchers/__tests__/transactions/types/vote.test.ts
@@ -9,6 +9,6 @@ describe(".toBeVoteType", () => {
     });
 
     test("fails when given an invalid transaction", () => {
-        expect({ type: "invalid" }).not.toBeVoteType();
+        expect(expect({ type: "invalid" }).toBeVoteType).toThrowError("Expected value to be a valid VOTE transaction.");
     });
 });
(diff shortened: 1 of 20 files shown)


JeremiGendron:
I may have included commits from my previous branch

packages/core-jest-matchers/__tests__/transactions/valid.test.ts
```diff
@@ -24,6 +24,6 @@ describe(".toBeValidTransaction", () => {
 
     test("fails when given an invalid transaction", () => {
         transaction.fee = "invalid" as any;
-        expect(transaction).not.toBeValidTransaction();
+        expect(expect(transaction).toBeValidTransaction).toThrowError("Expected value to be a valid transaction");
```
faustbrian: What exactly is this supposed to do?

packages/core-jest-matchers/__tests__/transactions/types/vote.test.ts
```diff
@@ -9,6 +9,6 @@ describe(".toBeVoteType", () => {
     });
 
     test("fails when given an invalid transaction", () => {
-        expect({ type: "invalid" }).not.toBeVoteType();
+        expect(expect({ type: "invalid" }).toBeVoteType).toThrowError("Expected value to be a valid VOTE transaction.");
```
faustbrian: What exactly is this supposed to do?

JeremiGendron:
They're being called by jest, there's no other way of verifying the returned message. If I pass the invocation it'll return a failed test and if I pass the function by reference it'll evaluate it as a sandboxed matcher.

Basically it was needed to verify the result of a failing test without actually creating a failing test, it seems jest is able to evaluate the test without running it with this method.

JeremiGendron:
I played around with it a bit, if it's preferred I'll add back the existing tests for usability references and keep these in for coverage (unless otherwise noted).

Issue #1965, opened 2019-01-11
Generalising the vendor header


alexbarnsley:
Is it possible/worthwhile to generalise the vendorName [here](https://github.com/ArkEcosystem/core/blob/master/packages/core-api/lib/defaults.js#L35) so that it doesn't get changed with each fork/chain. We can cater for it in wallets but I think it will be too confusing to users to understand what its intention is even for. The alternative is to allow us to get it somehow, perhaps from the `/config` endpoint. 

To clarify, this is the header sent when making API requests: `application/vnd.ark.core-api.v2+json`

If it is the case that we can generalise, I think we'll need to allow both for a while until devnet is merged with mainnet at least for backwards compatibility, otherwise the desktop wallet will only work on one of the networks.

faustbrian:
I think we should hold off this change until 2.5.0 as that will be a release that introduces a few breaking changes for the API anyways.

alexbarnsley:
What if we could generalise (`core-api`? 🤷‍♂️) now, but also check for `ark.core-api`. and then remove the `ark.core-api` check in 2.5.0? My only concern with doing it in 2.5.0 is that there may be quite a few chains by then. Currently phantom can't be added to the desktop wallet as a v2 network because they changed this to `phantom.core-api`. if we really need to we can add an option on the wallets to change this, it's just a weird thing to non-technical users to have to specify when connecting to a new network. That being said, I don't want to necessarily write/change core around the wallet

PR #1970, opened 2019-01-14
chore: debrand core by renaming ARK to CORE

faustbrian:
## Proposed changes

This PR debrands core by renaming all instances of `ARK` to `CORE` and the API vendor from `ark.core-api` to `core-api`.

This makes core more generic and easier for sidechains to deploy as they don't need to rename anything, simply generate your custom network config and you are ready to go.

Resolves https://github.com/ArkEcosystem/core/issues/1965

## Types of changes

- [x] Other

## Checklist


geopsllc:
Since 2.1 already changes config structure, it will make sense to also change the data folder now as mentioned below. There are still quite a few more things that need debranding but this is a good start.
1. Default USER from "ark" to "core"
2. CORE_DB_DATABASE from `ark_${process.env.CORE_NETWORK_NAME}` to `core_${process.env.CORE_NETWORK_NAME}`
3. Data folder from .ark/ to .core/
4. All the comments from Ark to Core
5. toBeArkAddress, arkBlock, arkAddress, arkBlockId, arkTransactionArray, arkPublicKey, arkTransfer, arkSecondSignature, arkDelegateRegistration, arkVote, arkMultiSignature, arkUsername, arkDelegateResignation, arkIpfs, arkMultiPayment, arkTimelockTransfer, ark-codec, ArkCodec, arkCrypto, arkEncoders, ark-development, ark-core, ark-devnet, ark-mainnet, ark-testnet
6. module core-snapshots is riddled with "ark" everywhere, including filenames.
7. Maybe also rebrand arkjs to corejs - not a must but will make things consistent.

faustbrian:
1. Makes sense.
2. Makes sense.
3. Makes sense.
4. Makes absolutely no sense.
5. Makes sense.
6. Makes sense.
7. No clue where arkjs has been seen since the release of core.

geopsllc:
Found a few more ark references, some coming from the develop merge:
import { client as ark }
builder = ark.getBuilder().*
expect(expect("invalid-address").toBeArkAddress).toThrowError("Expected value to be a valid address");
expect(expect("invalid-pubkey").toBeArkPublicKey).toThrowError("Expected value to be a valid public key");

4. This was just to rebrand the code comments... not a must of course.
7. You're right ... it's not in the code. It's just in a couple comments:
packages/core-forger/src/manager.ts:                // technically it is possible to compute doing shennanigan with arkjs.slots lib
packages/core-forger/dist/manager.js:                // technically it is possible to compute doing shennanigan with arkjs.slots lib

geopsllc:
I tested this branch with my custom installer on devnet and it works perfectly fine. Thanks!

fix:
Great move.

Just one 'philosophical' point: as soon as the sidechain builder has 'named' his sidechain, it should possible to 'call' it with the name.

Specifically for instance : 
- the database name/user
- the adresses (create an ark address or a dark, phantom, persona address etc...)
- the nodes/network (connect to an ark/mainnet node)



Also loosely linked to this aspect, desktop/mobile wallet should be able to 'reflect' custom sidechain (logo, colors, ticket etc...) Maybe published on autoconfigure endpoint

geopsllc:
Everything is quite generic right now which makes it very flexible to modify for other chains. The database name/user is already set by the installer script so what's in the code doesn't matter - generic 'core' is fine. I can see .core/ and core/ being a problem and I guess it can be reverted to .ark/ and ark-core/ if it is so desired. I'm done with a script already that can change those paths in the code on the fly. The only issue with that approach is updating the core afterwards will fail  if any of the files containing the paths are modified in the repo.  

P.S. The latest merge didn't pass the checks ... there is one hard-coded ark-core path.

PR #1981, opened 2019-01-15
perf(core-blockchain): optimize StateStorage.getCommonBlocks()

vasild:
Avoid a pattern like

  for each element in array1 do: check if it exists in array2 (linearly)
  aka
  for each element in array1 do: for each element in array2 ...

because its complexity is O(array1.length * array2.length). It can be
easily reduced to O(array1.length) if the check for existence is made
O(1) (hash lookup).

## Types of changes

## Checklist

### DIFF
--- a/packages/core-blockchain/src/state-storage.ts
+++ b/packages/core-blockchain/src/state-storage.ts
@@ -149,8 +149,10 @@ export class StateStorage implements Blockchain.IStateStorage {
      * Get common blocks for the given IDs.
      */
     public getCommonBlocks(ids): models.IBlockData[] {
+        const idsHash = {};
+        ids.forEach(id => idsHash[id] = true);
         return this.getLastBlocksData()
-            .filter(block => ids.includes(block.id))
+            .filter(block => idsHash[block.id])
             .toArray() as models.IBlockData[];
     }
 


supaiku0:
Hmm, I wonder if it even makes a difference.
`getLastBlockDatas` returns a [immutable.Seq](https://facebook.github.io/immutable-js/docs/#/Seq) which is doing a lot of optimizations under the hood.

vasild:
> Hmm, I wonder if it even makes a difference.
> `getLastBlocksData` returns a [immutable.Seq](https://facebook.github.io/immutable-js/docs/#/Seq) which is doing a lot of optimizations under the hood.

I did not bother to measure the impact of this patch because it is obvious: [Seq.filter](https://facebook.github.io/immutable-js/docs/#/Seq/filter) applies the given predicate to every element in the sequence. The predicate itself calls [Array.includes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes) which will probe every element in the array for a match, [linearly](https://en.wikipedia.org/wiki/Linear_search).

For example, for an input `ids[]` with 100 elements and `getLastBlocksData()` returning 100 blocks, the old code will do `100 * 100 = 10000` operations. The new code will iterate over `ids[]` once (100 operations) to build the hash and then iterate over the return Seq from `getLastBlocksData()` querying the hash on each iteration (100 operations). Overall 200 operations vs 10000.

For small numbers this may not matter much, e.g. for `ids[]` with 5 elements and `getLastBlocksData()` returning 10 elements we would get `5 * 10 = 50` vs `5 + 10 = 15`. However, the change in the code is not big either, so I think it is justified.

PR #1982, opened 2019-01-15
fix(core-database-postgres): off-by-one error in getBlocks()

vasild:
The getBlocks() function takes offset,limit arguments, presumably with
the intention of returning maximum `limit` number of elements, starting
from `offset`. However it would return `limit + 1` elements. E.g. if
given offset=5,limit=2 it would return three elements: 5,6,7 instead of
the expected two 5,6.

As a side effect of this, the `if (blocks.length !== limit)` check
inside getBlocks() would never be true and we would always retrieve the
blocks from the database, even if they are present in the "state".

The callers of getBlocks() were adapted to this wretched behavior and
did ask for `limit - 1`, knowing they would get one more. Adjust that,
now that it is not needed.

Also, rename a confusingly named variable `blocksToRemove` which
actually does not contain the blocks to be removed (!).

## Types of changes

## Checklist


supaiku0:
Good catch.

Though, why is this confusing? What does it contain, if not the blocks that are going to be removed? :thinking: 
> Also, rename a confusingly named variable blocksToRemove which
actually does not contain the blocks to be removed (!).

```typescript
  const lastBlock = this.state.getLastBlock();
  await this.database.enqueueDeleteBlock(lastBlock);

   const newLastBlock = new Block(blocksToRemove.pop()); // <-- block to remove
   this.state.setLastBlock(newLastBlock);
   this.state.lastDownloadedBlock = newLastBlock;
```
It doesn't contain `Block` instances, because it fetched raw blocks from the database. But imo `auxBlocks` is even more confusing.

vasild:
`blocksToRemove` is confusing because, for example, if the chain is
..., 100, 101, 102, 103 and we are about to remove the last two blocks 102 and 103. Then! Then, `blocksToRemove` will contain blocks 101 and 102 (!?). It took me some staring at the code and some headache to understand that. Notice - not the blocks to be removed 102 and 103, but instead 101 and 102 instead.

So, the name `blocksToRemove` gives the impression that it is something which it is not. Just `blocks` or even `blackBox42` would be better because it would not trick the reader to think something which is wrong.

vasild:
@faustbrian 

> The variable name itself should tell us what it contains

Yes, I completely agree! However I couldn't come up with a meaningful name, other than `blocksToRemoveExceptWithoutTheLastOneAndWithThePreviousOneAttachedAtTheStartWhichWeWillNotRemove`.

To get this correct, the code needs some bigger refactor, which I did not wanted to dive into.

PR #2003, opened 2019-01-19
fix: docker folder was deleted

JeremiGendron:
Not sure if this was intentional or not, anyways the master/docker folder is added back into this PR against develop; which no longer has the docker folder.


faustbrian:
Was removed intentionally because there is now a script to generate it for the correct coin.

JeremiGendron:
alright, good to know

PR #2005, opened 2019-01-21
test(crypto): increase model coverage

air1one:
## Proposed changes

Added tests to crypto models to get coverage to 100% for the models folder.

Also please have a look at the refactor on models/wallet.

## Types of changes

## Checklist

- [x I have read the [CONTRIBUTING](https://docs.ark.io/guidebook/contribution-guidelines/contributing.html) documentation

### DIFF
--- a/packages/crypto/src/models/wallet.ts
+++ b/packages/crypto/src/models/wallet.ts
@@ -123,18 +123,16 @@ export class Wallet {
      * Remove block data from this wallet.
      */
     public revertBlock(block: IBlockData): boolean {
-        this.dirty = true;
-
         if (
             block.generatorPublicKey === this.publicKey ||
             crypto.getAddress(block.generatorPublicKey) === this.address
         ) {
+            this.dirty = true;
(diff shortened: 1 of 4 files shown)


packages/crypto/src/models/wallet.ts
```diff
@@ -123,18 +123,16 @@ export class Wallet {
...
             block.generatorPublicKey === this.publicKey ||
             crypto.getAddress(block.generatorPublicKey) === this.address
         ) {
+            this.dirty = true;
             this.balance = this.balance.minus(block.reward).minus(block.totalFee);
 
             // update stats
             this.forgedFees = this.forgedFees.minus(block.totalFee);
             this.forgedRewards = this.forgedRewards.minus(block.reward);
-            this.lastBlock = block;
```
faustbrian: The logic here is wrong but simply removing this code won't solve the issue. We should probably store the last 2 or N blocks of a delegate and pop the last block and reattach so we know it is the previous block of the one we removed.

PR #2010, opened 2019-01-22
fix(core-p2p): only log error if error object is not empty

alessiodf:
## Proposed changes
Useless "undefined" errors could be written to the log when a malformed block is received because the error object is empty in these circumstances. This fixes it by checking that the error object is not empty before writing to the log.

## Types of changes

## Checklist

### DIFF
--- a/packages/core-p2p/src/server/versions/1/handlers.ts
+++ b/packages/core-p2p/src/server/versions/1/handlers.ts
@@ -180,7 +180,10 @@ export const postBlock = {
 
             return { success: true };
         } catch (error) {
-            logger.error(error);
+            //  Only write the error to the log if the error object is not empty
+            if(!!Object.keys(error).length) {
+                logger.error(error);
+            }
             return { success: false };
         }
     },


packages/core-p2p/src/server/versions/1/handlers.ts
```diff
@@ -180,7 +180,10 @@ export const postBlock = {
 
             return { success: true };
         } catch (error) {
-            logger.error(error);
+            //  Only write the error to the log if the error object is not empty
+            if(!!Object.keys(error).length) {
```
vasild: What is the purpose of `!!`? Would the code behave in an identical way without it?

I would suggest to avoid treating a number like a boolean and do `Object.keys(error).length > 0` instead.

Missing space after `if`.

faustbrian:
This should be solved directly in the logger rather then a single method that uses the logger to avoid those issues in general.

supaiku0:
Given that the error can only originate from the `Block` constructor it appears that this log statement will never yield anything useful. You could also just remove it.

faustbrian:
https://github.com/ArkEcosystem/core/pull/2011

PR #2024, opened 2019-01-28
Added extra checks

adrian69:
## Proposed changes

Minor database setup fix.
Add CentOS 6/7 support.
Detect SystemV / SystemD.
Improve existing user and database handling.

## Types of changes

## Checklist

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -118,6 +118,9 @@ paragraph ()
 DEB=$(which apt-get)
 RPM=$(which yum)
 
+# Detect SystemV / SystemD
+SYS=$([[ -L "/sbin/init" ]] && echo 'SystemD' || echo 'SystemV')
+
 if [[ ! -z $DEB ]]; then
     success "Running install for Debian derivate"
 elif [[ ! -z $RPM ]]; then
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -275,30 +286,36 @@ if [[ "$choice" =~ ^(yes|y|Y) ]]; then
     read -p "Enter the database password: " databasePassword
     read -p "Enter the database name: " databaseName
 
-    userExists=$(sudo -u postgres psql -c "SELECT * FROM pg_user WHERE username = '${databaseUsername}'" | grep -c "1 row")
+    userExists=$(sudo -i -u postgres psql -c "SELECT * FROM pg_user WHERE usename = '${databaseUsername}'" | grep -c "1 row")
+    databaseExists=$(sudo -i -u postgres psql -l | grep "${databaseName}" | wc -l)
```
vasild: This `grep` will match any database that contains the name, e.g. if `databaseName=foo` it will match databases named `foobar` or `testfoo`. Also, because the owner of the database is also displayed it will match if the owner contains `foo`.

install.sh
```diff
@@ -275,30 +286,36 @@ if [[ "$choice" =~ ^(yes|y|Y) ]]; then
...
 
     if [[ $userExists == 1 ]]; then
         read -p "The database user ${databaseUsername} already exists, do you want to overwrite it? [y/N]: " choice
 
         if [[ "$choice" =~ ^(yes|y|Y) ]]; then
```
vasild: `yes` is redundant because the single `y` will match it. I.e. `^(yes|y|Y)` is identical to `^(y|Y)`. Same for `^(no|n|N)` below.

PR #2032, opened 2019-01-29
refactor(core-p2p): more extensive peer state verification

vasild:
## Proposed changes

Verify each peer's claimed state so that we don't feed garbage data to the quorum calculation process.

## Types of changes

## Checklist


packages/core-p2p/src/peer-verifier.ts
```diff
@@ -0,0 +1,553 @@
...
+            const hisBlockObj = await this.verifyPeerBlock(hisBlockJson, h, delegates);
+            if (hisBlockObj === null) {
+                return false;
+            }
+
+            if (includeHisBlocksInOurChain) {
+                // XXX this code is the same as
+                // packages/core-p2p/src/server/versions/1/handlers.ts:postBlock
+
+                /* XXX this.blockchain is null here
```
vasild: Figure out why blockchain is null. Maybe during startup this code is executed before the blockchain is started. If so, see if it is possible to start the blockchain earlier or if not, exec this code only if not null (e.g. only during pings after startup has completed).

vasild: Yes, blockchain is started after p2p and it would be too involved to swap the order as it would require removing the p2p dependencies from blockchain.

kristjank:
what is the status on this PR ?

vasild:
> what is the status on this PR ?

What's left on my side:
* cache verified blocks ids, to avoid downloading one block multiple times
* fix tests - some tests of the `ping()` method fail now, presumably because they mock some responses which need to be extended
* write tests for the new code in peer-verifier.ts
* consider the ping timeout - right now the `ping()` method takes a timeout, but the newly added code does things not governed by this timeout, so `ping()` could take longer than the supplied timeout, which is probably not ok for the final version of the patch.

So, it is still work in progress, however the remaining changes will not alter the code much, so I guess if anybody is interested he can get familiar with the code now (pre-review). And/or give suggestions about the above points.

packages/core-p2p/src/peer-verifier.ts
```diff
@@ -0,0 +1,540 @@
...
+import { CappedSet, roundCalculator } from "@arkecosystem/core-utils";
+import { models } from "@arkecosystem/crypto";
+import assert from "assert";
+import { Peer } from './peer';
+
+/**
+ * A cache of verified blocks' ids. A block is verified if it is connected to a chain
+ * in which all blocks (including that one) are signed by the corresponding delegates.
+ */
+const verifiedBlocks = new CappedSet();
```
air1one: Wondering why `verifiedBlocks` is here as a global variable and not as a property of the class ?

vasild: My intention is to have it shared between all instances of the class. For example - if we verify blockX and blockY during verifying peer1, and after this we start verifying peer2 and he says he has blockX, then we would not need to download blockX from peer2 and run it through verification.

I wonder if there is a better way to have the `verifiedBlocks` shared between the class instances? If not maybe at least I should put a comment explaining the intention.

supaiku0: You can make it private static field inside the class e.g 
```typescript
private static readonly verifiedBlocks = new CappedSet();
```

and then access it from within an instance via `PeerVerifier.verifiedBlocks`.

air1one: Ok thanks resolving conversation, will run it through my tests now.

faustbrian:
What's the status on this?

air1one:
Finally managed to test with e2e framework, couldn't hack as before into being added as a peer, here are the related logs : 
```
[2019-02-05 12:29:49][INFO]: Peer verify 10.0.3.95: peer's claimed state is invalid: {"success":true,"height":3655,"forgingAllowed":true}
[2019-02-05 12:29:49][DEBUG]: Could not accept new peer 10.0.3.95:4002: Error: Peer state verification failed for peer 10.0.3.95, claimed state: {"success":true,"height":3655,"forgingAllowed":true}
[2019-02-05 12:29:49][DEBUG]: Suspended 10.0.3.95 for 5 minutes because of "Invalid Version"
```
So looks OK, not sure about the suspension reason "Invalid Version" but peer was not added which is the more important.

vasild:
> What's the status on this?

Remaining:

* write tests for the new code

* consider the ping timeout - right now the ping() method takes a timeout, but the newly added code does things not governed by this timeout, so ping() could take longer than the supplied timeout, which is probably not ok for the final version of the patch.

Also, I just chopped of the search code into a generic, reusable component and moved it to `core-utils`.

vasild:
I tried to reproduce the "Invalid version" by applying this patch:
```
diff --git i/packages/core-p2p/src/peer.ts w/packages/core-p2p/src/peer.ts
index d77618781..16b730daf 100644
--- i/packages/core-p2p/src/peer.ts
+++ w/packages/core-p2p/src/peer.ts
@@ -191,13 +191,14 @@ export class Peer implements P2P.IPeer {
      */
     public async ping(delay, force = false) {
         if (this.recentlyPinged() && !force) {
             return;
         }

-        const body = await this.__get("/peer/status", delay || localConfig.get("globalTimeout"));
+        let body = await this.__get("/peer/status", delay || localConfig.get("globalTimeout"));
+        body = {"success":true,"height":3655,"forgingAllowed":true};

         if (!body) {
             throw new Error(`Peer ${this.ip} is unresponsive`);
         }

         if (!body.success) {
```

but I only get:
```
[2019-02-05 17:13:24][INFO]: Peer verify 167.114.29.36: peer's claimed state is invalid: {"success":true,"height":3655,"forgingAllowed":true}
[2019-02-05 17:13:24][DEBUG]: Removed peer 167.114.29.36:4002: Peer state verification failed for peer 167.114.29.36, claimed state: {"success":true,"height":3655,"forgingAllowed":true}
```

How to get that "Invalid version"?

air1one:
OK so the "invalid version" is unrelated, I actually sent headers of an invalid version @vasild :smile: (see guard.js for suspension logic)

About testing this in unit tests, yes totally possible, have a look there are already some where we mock peers, would be appropriate for this.

vasild:
IMO this is good for code review and following that - to hit devnet.

Now I will focus on extending the test coverage.

faustbrian:
Any progress on the test coverage?

vasild:
> Any progress on the test coverage?

https://github.com/ArkEcosystem/core/pull/2032/commits/5ee13dc137e6addfefa99a3e9b0032abe79c26d5

`+` I am now writing more tests

packages/core-p2p/src/peer-verifier.ts
```diff
@@ -0,0 +1,512 @@
...
+
+        return true;
+    }
+
+    /**
+     * Check if the state claimed by the peer is definitely invalid.
+     * @param {Object} claimedState peer's claimed state (from `/peer/status`)
+     * @return {Boolean} true if invalid
+     */
+    private isStateInvalid(claimedState: any): boolean {
```
supaiku0: This looks wrong on so many levels.

The proper way would be to have schema validation for the
endpoint `/peer/status` and introduce an interface for the payload.

But it is out of scope of this PR, so just leave it like that for now.

vasild: Right, a schema validation would be definitely superior to manually checking if some fields are present and have the expected types. Not sure about the interface for the payload. I will look into this after doing the tests for this or a followup PR.

vasild:
> Overall it looks good. But it needs to be made less verbose, even just a summary would be fine or only logging bad peers.

Yes, quite a lot of messages end up in the log. I used most of those for debugging and during developing the patch to figure out what is going on. And they are useless as long as things are going normal and verification succeeds. What about some way to enable/disable the extra verbosity? It could be useful if something goes wrong - then we will need all available info to diagnose what happened.

packages/core-p2p/src/peer-verifier.ts
vasild: Actually the problem of manually checking fields instead of using schema validation is present in the code even before this patch. The code is peppered with such checks, for example `Peer.hasCommonBlocks()` has `body && body.success && body.common`. `Peer.getPeers()` assumes that the reply will contain `peers` and that it will be an array. `Peer.ping()` lacks any checks on what is returned.

I think the proper way to handle this is to extend `Peer.__get(endpoint, timeout)` with a 3rd argument - a schema that the reply must match. That would be out of scope for this PR, more suitable for a followup PR.

supaiku0: Yeah, that's why I said it's out of scope. This is a general issue.

faustbrian:
@vasild try to get this ready until tomorrow so we can get it on devnet to ship it together with the CLI asap.

vasild:
> Overall it looks good. But it needs to be made less verbose, even just a summary would be fine or only logging bad peers.

I reduced the number of messages printed at
https://github.com/ArkEcosystem/core/pull/2032/commits/ca88e0baaaeba77ffb561f83d171d14c2ef235d5
and introduced a way to still enable all the printouts, should a debugging be needed.

faustbrian:
@vasild can this be merged and you send the missing tests/schema refactor in another PR?

vasild:
> @vasild can this be merged and you send the missing tests/schema refactor in another PR?

Yes, IMO it is good to go since the time it does not have `wip:` in the title. It is in reviewers' hands.

* tests - more tests can and should be added, however I do not think it is worth the effort getting peer-verifier.ts coverage to 100% given that all tests are going down the drain soonish.

* schema validation - it is out of the scope of this PR because it would affect pieces of code not touched by this patch and I will do a followup patch. Thus the absence of schema validation should not block this patch.

Btw the entire peer verification can be disabled by defining `CORE_SKIP_PEER_STATE_VERIFICATION` in the environment, should some striking problems surface when we run this on devnet. I guess we may decide to increase some timeouts a little, but lets see what real world workload will show.

packages/core-p2p/src/peer-verifier.ts
vasild: Will do schema validation here and in all places (in peer.ts or even all of core-p2p) in a followup patch.

kristjank:
Yes, we should start testing and use a script to disturb the positive flow, like botnet or related. so we can see the workload and how timeouts will behave -> and what is the overall effect on quorum.

PR #2055, opened 2019-02-03
Separate business-logic from data-layer logic in core-database/postgres

### LINKED ISSUE #2006: Separate business logic from data-layer logic in core-database/postgres
The `ConnectionInterface` class is conflating business logic and data-layer logic into one class. This violates the SR(single-responsibility) principle, which makes it hard to extend, test and refactor a component(evident w/ the lack of tests and even difficulty to write tests)

A clear boundary needs to be defined between business logic and data-layer logic.
Introduce two separate components.

- One that's responsible purely for interacting with the database(fetching and persisting data).
This provides an interface containing methods to support the business logic component.
It also makes it easier to replace this component for a different one(if you wish to use a different DB implementation such as NoSQL)

- The second business service, which performs business logic(applying blocks/rounds, verifying transactions, updating del stats, etc) while leveraging the data-layer component only when it needs to interact with the data models

This makes it easier to write tests(stubbing is much easier), refactor, or extend said logic as needed.

paroxysm:
## Proposed changes

Resolves #2006

The PR introduces a hierarchical change to the core-database/postgres.

I've encapsulated the business logic, previously in `ConnectionInterface` and `PostgresConnection`,  into a  new class `DatabaseService`, implements `IDatabaseService`, which references a database specific implementation of  the `IDatabaseConnection` interface via constructor injection.

`IDatabaseConnection` supplies the repos that `DatabaseService` uses in certain methods.

The plugin providing a `IDatabaseConnection` concrete class is responsible for instantiating a `DatabaseService` and supply a `IDatabaseConnection` concrete class, among with other dependencies such as `IWalletManager`

In theory, the plugin can also provide its own implementation of `IWalletManager` and such..

The implication is that we're now exposing an `IDatabaseService` implementation to be used by the other modules. Most of these modules don't care about the underlying data-layer and shouldn't.
The exception would be core-api/core-graphql/core-snapshots which are(oddly) defining repositories that are accessing postgres specific internals(I use 'any' type-casting to get past compilation errors but this tech-debt that should be addressed sooner than later). These external repos need to be relocated into `IDatabaseConnection` repos and w/e additional logic they were doing be merged.

It shouldn't be too hard to convert the search/query code these external repositories are using into a generic implementation on the database specific repos.

I've tested that my relay runs and syncs as usual. I haven't tested some other affected modules such as core-graphql/snapshot outside their tests

The next step would be to start improving the `DatabaseService` tests, now that we can stub easier.

## Types of changes

## Checklist

### DIFF
--- a/packages/core-api/src/repositories/transactions.ts
+++ b/packages/core-api/src/repositories/transactions.ts
@@ -15,7 +15,7 @@ export class TransactionsRepository extends Repository implements IRepository {
      * @param  {Object}  params
      * @return {Object}
      */
-    public async findAll(parameters: any = {}): Promise<any> {
+    public async findAll(parameters: any = {}, sequenceDesc = true): Promise<any> {
         const selectQuery = this.query.select().from(this.query);
 
         if (parameters.senderId) {
@@ -50,7 +50,7 @@ export class TransactionsRepository extends Repository implements IRepository {
         const results = await this._findManyWithCount(selectQuery, {
             limit: parameters.limit,
             offset: parameters.offset,
-            orderBy: this.__orderBy(parameters),
+            orderBy: this.__orderBy(selectQuery, parameters, sequenceDesc),
         });
 
         results.rows = await this.__mapBlocksToTransactions(results.rows);
@@ -93,7 +93,7 @@ export class TransactionsRepository extends Repository implements IRepository {
         const results = await this._findManyWithCount(selectQuery, {
             limit: parameters.limit,
             offset: parameters.offset,
-            orderBy: this.__orderBy(parameters),
+            orderBy: this.__orderBy(selectQuery, parameters),
         });
 
         results.rows = await this.__mapBlocksToTransactions(results.rows);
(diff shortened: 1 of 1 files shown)


packages/core-snapshots/src/plugin.ts
```diff
@@ -14,6 +14,7 @@ export const plugin: Container.PluginDescriptor = {
     async register(container: Container.IContainer, options) {
         const manager = new SnapshotManager(options);
 
-        return manager.make(container.resolvePlugin<PostgresConnection>("database"));
+        const connection = container.resolvePlugin<Database.IDatabaseService>("database").connection as any;
```
faustbrian: This causes some resolution issues. Try `yarn dump:testnet --config ../core/src/config/testnet` through the CLI to reproduce it.

```
TypeError: Cannot read property 'connection' of null
    at Object.register (./packages/core-snapshots/dist/plugin.js:15:63)
```

paroxysm:
I'll take a look after work 👍🏿

packages/core-snapshots/src/plugin.ts
paroxysm: @faustbrian I've solved the issue.

faustbrian:
@paroxysm do you want to take care of extracting the repositories out of core-api/graphql?

paroxysm:
Yea, I'll handle that

Issue #2041, opened 2019-01-31
[core-api] order of transactions with same timestamp


dated:
**Describe the bug**
The API returns transactions with the same timestamp in an unpredictable and wrong manner.

**To Reproduce**
1. Go to https://dexplorer.ark.io/api/v2/wallets/cam/transactions?page=1&limit=5&orderBy=timestamp:desc
2. Go to https://dexplorer.ark.io/api/v2/wallets/cam/transactions?page=2&limit=5&orderBy=timestamp:desc
2. Search for: `64c15a316786216825f68240ce733db052d2788f43ea94acdb5acabe484382c4 `
5. The same transaction is present in both responses

Both JSON responses: https://pastebin.com/UFrG0qhZ

**Expected behavior**
The above mentioned transaction should not be present on the second page, if it is on the first.

**Additional context**
The results should be sorted by an additional fallback parameter with a unique value, e.g. the transaction id.

faustbrian:
Might be solved if we sort by `timestamp` and `sequence`, will try to take a look within the next few days.

dated:
Cool, thanks. All applications that depend on pagination are potentially affected. The impact on the wallets and explorer is minimal, applications that parse transaction data may be returning unexpected or wrong data as some transactions could be missing, while others are parsed multiple times.

vasild:
Maybe something like `...&orderBy=timestamp:desc,id:desc` should work, or even just `...orderBy=timestamp:desc` + we internally sort by "timestamp, some-field-which-is-unique-for-example-id" to make it deterministic.

@dated notice however this: even if that deficiency is resolved, the same issue may manifest for different reasons - if a new transaction is added between the two queries. For example:
1. you fetch https://dexplorer.ark.io/api/v2/wallets/cam/transactions?page=1&limit=5&orderBy=timestamp:desc
// you get the 1-5 top newest transactions, lets say t1, t2, t3, t4, t5
2. you fetch https://dexplorer.ark.io/api/v2/wallets/cam/transactions?page=2&limit=5&orderBy=timestamp:desc
// you get the 6-10 top newest transactions t5, t6, t7, t8, t9, t10. Wait, what? t5 again? Yes, because a new transaction t0 has been added between 1. and 2., so now the list of transactions is (ordered by timestamp, newest first): t0, t1, t2, t3, t4, t5, t6, t7, t8, t9, ...

The solution to that would be to avoid pagination and fetch all transactions needed with one query within the app.

[...]

PR #2058, opened 2019-02-04
fix(core-api): always sort transactions by sequence and the requested field

faustbrian:
## Proposed changes

Resolves https://github.com/ArkEcosystem/core/issues/2041

**Note: Currently the sequence order is ascending like the SPV, might be better to sort descending?**

## Types of changes

## Checklist


packages/core-api/src/repositories/transactions.ts
```diff
@@ -500,7 +500,9 @@ export class TransactionsRepository extends Repository implements IRepository {
         return null;
     }
 
-    public __orderBy(parameters): string[] {
+    public __orderBy(selectQuery, parameters): string[] {
+        selectQuery.order(this.query.sequence.asc);
```
vasild: We would end up doing (in this order):
```
selectQuery.order(this.query.sequence.asc);
...
this._findManyWithCount(selectQuery, { ..., orderBy: [ "timestamp", "desc" ] });
```

I am not sure how this JS-to-SQL library works, but would it produce SQL `... ORDER BY id ASC, timestamp DESC` or  `... ORDER BY timestamp DESC, id ASC`?

vasild: This produces the following query:

`SELECT "transactions"."block_id", "transactions"."serialized", "transactions"."timestamp" FROM "transactions" WHERE (("transactions"."sender_public_key" = $1) OR ("transactions"."recipient_id" = $2)) ORDER BY "transactions"."sequence" DESC, "transactions"."timestamp" DESC OFFSET 0 LIMIT 5}`

which is probably not what the intention was, and is definitely not returning results sorted by timestamp, even though `...&orderBy=timestamp:desc` is present in the URL.

For example:

```
SELECT timestamp, sequence FROM transactions WHERE sender_public_key = '0219c48e4b9c566b9ed26ad8f762e1fcf425c72776fce61e4839705dd35bfd43dd'
ORDER BY sequence DESC, timestamp DESC;
 timestamp | sequence 
-----------+----------
  49221058 |        8
  49220498 |        7
  49220778 |        5
  49225368 |        4
  49219410 |        4
  49226538 |        3
  49225714 |        3
  49226154 |        1
  49225058 |        1
  49224890 |        1
  49126378 |        1
  49104538 |        1
```

```
SELECT timestamp, sequence FROM transactions WHERE sender_public_key = '0219c48e4b9c566b9ed26ad8f762e1fcf425c72776fce61e4839705dd35bfd43dd'
ORDER BY timestamp DESC, sequence DESC;
 timestamp | sequence 
-----------+----------
  49226538 |        3
  49226154 |        1
  49225714 |        3
  49225368 |        4
  49225058 |        1
  49224890 |        1
  49221058 |        8
  49220778 |        5
  49220498 |        7
  49219410 |        4
  49126378 |        1
  49104538 |        1
(12 rows)
```

faustbrian: Submit a PR to improve it rather then simply commenting.

vasild: Will do.

However, why did you merge a flawed patch in the first place? It is a good practice to look into and resolve all review comments before merging a pull request.

PR #2093, opened 2019-02-13
fix(core-api): Proper hex validation

vasild:
Buffer.from(value, "hex") will only throw an exception if value is not a
string, according to
https://nodejs.org/api/buffer.html#buffer_class_method_buffer_from_string_encoding
Indeed Buffer.from('foo', 'hex') does not throw. The schema has defined
`type: "string"`, which means that, in all cases when the validate
method is called, it will return true.

Use a proper validation using regular expression and a length check.

## Types of changes

## Checklist


faustbrian:
That is the legacy API which will be removed soon so don't waste time on that in the future.

PR #2102, opened 2019-02-14
feat(core-p2p): Validate GET replies from other peers

vasild:
## Types of changes

## Checklist


faustbrian:
This is a breaking change in how the **legacy** P2P API responds, some people still rely on it so that is a no go. This is a change that will be made for websockets.

vasild:
> This is a breaking change in how the **legacy** P2P API responds, some people still rely on it so that is a no go. This is a change that will be made for websockets.

We agreed that I add a schema validation as a followup PR after https://github.com/ArkEcosystem/core/pull/2032:
https://github.com/ArkEcosystem/core/pull/2032#event-2135994744
https://github.com/ArkEcosystem/core/pull/2032#discussion_r256279130

Now, should I relax the validation so that it allows "legacy P2P API" replies, or completely ditch this PR
@supaiku0, @kristjank?

vasild:
By the way I get the following with this patch:
```
[ERROR]: Got unexpected reply from http://137.74.237.196:4002/peer/list: {"peers":[{...},{"ip":"2130706433",...}: child "peers" fails because ["peers" at position 1 fails because [child "ip" fails because ["ip" must be a valid ip address with a forbidden CIDR]]]
```
and the node keeps working by fetching the peer list from another peer. This looks like a good way to get rid of the IP addresses like "2130706433" with which devnet is poisoned and can't get rid of, even if removed locally.

faustbrian:
If you change it so that the responses don't change we can keep it open, otherwise I would wait until 2.6 with it and talk to @air1one as he currently is working on the websockets.

vasild:
> If you change it so that the responses don't change we can keep it open, otherwise I would wait until 2.6 with it and talk to @air1one as he currently is working on the websockets.

Yes, sure. The idea was never to change the responses. The idea is to check that the responses are the expected ones. So, I guess, the rules should be relaxed. @air1one how would this fit with websockets? I guess with websockets we still send JSONs back and forth, right? So the validation is still relevant, no matter whether we received the reply via HTTP or websockets.

faustbrian:
@vasild try to finish this until the end of the week so we can put it on devnet and freeze the develop branch for the 2.2 release.

packages/core-p2p/src/peer.ts
```diff
@@ -2,12 +2,92 @@ import { app } from "@arkecosystem/core-container";
...
 import axios from "axios";
 import dayjs from "dayjs-ext";
+import Joi from "joi";
 import util from "util";
 import { config as localConfig } from "./config";
 import { PeerPingTimeoutError, PeerStatusResponseError, PeerVerificationFailedError } from "./errors";
 import { PeerVerificationResult, PeerVerifier } from "./peer-verifier";
 
 export class Peer implements P2P.IPeer {
+    private static replySchemas: any = {
```
supaiku0: Please move the schemas into a separate file.

vasild: Moved in abb024254.

PR #2107, opened 2019-02-17
Consolidate core-api's repos into core-database & core-database-postgres

paroxysm:
## Proposed changes
Partially addresses #1784

More hierarchical changes, the repos in core-api have been split up into two logical parts.
A business repository and data-layer repository(which is just the pre-existing postgres repo implementations)

What core-api repos were doing(dynamic querying) has been abstracted in way that's understandable by data-layer implementations. I've introduced a few interfaces that abstract the intent to dynamically query for models.

`SearchParameter`, `SearchOperator`, `SearchOrderBy`, and `SearchPaginate`. The business repo layer is responsible for taking a generic `params` object and converting it into a `SearchParameters` object which is composed of the aforementioned classes. This allows us to keep the logic of how we convert our dynamic search input constant, while allowing any custom data-layer implementation to handle how it builds internal queries based on the aforementioned structures. 
The principle is simple, keep database specific details in the database specific plugin, and don't leak those details to other modules. 

Therefore, other plugins(graphql and core-api) will invoke the business repos of core-database which in turn invoke the data-layer repos(performing any shared logic in-between method calls before delegating)

I have yet to migrate core-graphql to use these business repos because I didn't want to deal w/ a lot of merging issues in case other contributors start messing with the database stuff.
I also haven't improved on the tests yet(remains the same but still passing), but now that we have better separation of concerns, it should be much easier to write non-brittle tests.

The highest impact would be query methods in core-api. I've done some high-level testing, and it mostly seems fine, but haven't done very detailed inspection, I could use some help with that.

The next step is to start writing tests for these new components, followed by consolidating core-graphql repos(they're very similar to core-api, looks like a copy-paste)

## Types of changes

## Checklist


faustbrian:
Thanks again, will take a look on tuesday.

packages/core-interfaces/src/core-database/database-repository/transactions-repository.ts
```diff
@@ -37,9 +38,20 @@ export interface ITransactionsRepository extends IRepository {
         totalAmount: Bignum
     }>;
 
+    /* TODO: Maybe consolidate this with 'statistics()'. How are they used, and why separate implementations? */
```
faustbrian: The `statistics` method is used to verify the integrity of the database. It's only used during the boot process of the node because the query is very expensive.

packages/core-database-postgres/src/models/transaction.ts
```diff
@@ -1,64 +1,112 @@
...
+     */
+    public getTable() {
+        return "transactions";
+    }
+
+    /**
+     * The read-only structure with query-formatting columns.
+     * @return {Object}
+     */
+    public getColumnSet() {
```
faustbrian: This method is exactly the same in all models, could be moved into the abstract model.

packages/core-database-postgres/src/models/transaction.ts
```diff
@@ -1,64 +1,112 @@
...
+            }))
+        }
+        return this.columnSet;
+    }
+
+    public getName(): string {
+        return "Transaction";
+    }
+
+    public getSearchableFields(): Database.SearchableField[] {
```
faustbrian: This method is exactly the same in all models, could be moved into the abstract model.

packages/core-database-postgres/src/models/transaction.ts
```diff
@@ -1,64 +1,112 @@
...
+                        colDef[prop] = col[prop];
+                    }
+                });
+                return colDef;
+            }))
+        }
+        return this.columnSet;
+    }
+
+    public getName(): string {
```
faustbrian: Could maybe be moved to the abstract model and use `this.constructor.name`?

packages/core-database-postgres/src/repositories/repository.ts
```diff
@@ -3,6 +3,7 @@ import { Model } from "../models";
 
 export abstract class Repository implements Database.IRepository {
     protected model: Model;
+    protected q;
```
faustbrian: Please rename that to `query` to avoid unnecessary abbreviations.

packages/core-database/src/database-service-factory.ts
```diff
@@ -1,12 +1,22 @@
...
+import { BlocksBusinessRepository } from "./repositories/blocks-business-repository";
+import { DelegatesBusinessRepository } from "./repositories/delegates-business-repository";
+import { TransactionsBusinessRepository } from "./repositories/transactions-business-repository";
+import { WalletsBusinessRepository } from "./repositories/wallets-business-repository";
 
 // Allow extenders of core-database to provide, optionally, a IWalletManager concrete in addition to a IDatabaseConnection, but keep the business repos common
 export const databaseServiceFactory = async (opts: any, walletManager: Database.IWalletManager, connection: Database.IDatabaseConnection): Promise<Database.IDatabaseService> => {
     let databaseService: DatabaseService;
-    databaseService = new DatabaseService(opts, connection, walletManager, new WalletsRepository(() => databaseService), new DelegatesRepository(() => databaseService));
+    databaseService = new DatabaseService(
```
faustbrian: This should be simplified rather then passing in more and more arguments for every repository that gets added.

_Any specific reason other then easier mocking that you pass it to the constructor as an argument rather then initialising it directly in the constructor itself?_

packages/core-database/src/repositories/blocks-business-repository.ts
```diff
@@ -0,0 +1,50 @@
+import { Database } from "@arkecosystem/core-interfaces";
+import { SearchParameterConverter } from "./utils/search-parameter-converter";
+
+export class BlocksBusinessRepository implements Database.IBlocksBusinessRepository {
+
+    constructor(private databaseServiceProvider: () => Database.IDatabaseService) {
+    }
+
+    public async findAll(params: any) {
+        /* In core-api/blocks repo. 'findAll' differs from search in that, it assumes 'equal' op for all passed in parameters
```
faustbrian: The `eq` behaviour is how v1 behaved, that is why the `search` and `findAll` methods currently differ as `search` is exposed through the v2 API so we have to keep the behaviour of `findAll` as is until we drop the v1 API.

packages/core-database/src/repositories/blocks-business-repository.ts
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    public async findLastByPublicKey(generatorPublicKey: string) {
+        // we order by height,desc by default
+        return await this.findAll({ generatorPublicKey});
+    }
+
+    public async findByHeight(height: number) {
+        const params = { ...{ height } };
+        return await this.findAll(params);
```
faustbrian:
```suggestion
        return await this.findAll({ height });
```

packages/core-interfaces/src/core-database/database-repository/transactions-repository.ts
paroxysm: ah ok.

packages/core-database-postgres/src/models/transaction.ts
paroxysm: I agree, I should be avoiding duplication hehe

packages/core-database/src/database-service-factory.ts
paroxysm: This is just a DI(dependency injection) convention. The factory method is just an opinionated way of constructing the database-service instance, not a decree.
We want to retain the ability to pass in different business repo implementations. 
This is in-line with "code is open to extension, closed to modification" methodology.
If the dependencies get too many(don't think it has yet), it's a sign the class is doing too much and needs to be broken up

packages/core-database/src/repositories/blocks-business-repository.ts
paroxysm: OK, are we still using v1 pathways still? I wasn't too sure if the v2 migration meant deprecation of v1 code...

faustbrian: All v1 or legacy code will be removed in 2.5 which will be roughly around summer this year, sadly have to keep it until then.

PR #2123, opened 2019-02-20
refactor(core-p2p): Remove unnecessary ping() call

vasild:
- getPeers() is only interested in /peer/list, remove unnecessary ping()
  invocation from getPeers()

- increase ping timeout when accepting new peer

- re-enable peer-verification

## Types of changes

## Checklist


vasild:
3 separate changes go in this PR. They are related, but can be chopped off in separate PRs.

faustbrian:
@vasild tests need adjustment

vasild:
> @vasild tests need adjustment

Yes, I am looking into this. Looks like the test relied on the side effect of getPeers() calling ping(). I am considering one of the following:

1. Adjust the test to call ping() explicitly, however the test's usage looks sensible: `monitor.discoverPeers(); monitor.getNetworkHeight();`

2. Adjust `discoverPeers()` to do the ping, or even more elegant: use `acceptNewPeer()` (which does the ping) instead of `addPeer()`. In this case `addPeer()` will become unused and can be removed.

3. Adjust `getNetworkHeight()` to do the pings.

faustbrian:
Would go with option 2

vasild:
> Would go with option 2

Yes. Another goodie popped out of this: `acceptNewPeer()` was returning early if it detected a test environment (`CORE_ENV === "test"`). In other words, it was practically disabled in test environments. However _all_ tests that call `acceptNewPeer()` fool that check by setting CORE_ENV to something else than "test" so that the function proceeds un-disabled. So, I have removed the "return early" code from `acceptNewPeer()` and also the hackish tweaking of environment from tests.

PR #2126, opened 2019-02-21
chore(core): remove accidentally committed env variables

faustbrian:
## Proposed changes

The database username variable was accidentally committed when the CLI was merged.

## Types of changes

- [x] Other

## Checklist


vasild:
I also accidentally committed some mods to the env files. It would be more flexible to be able to override the variables from the environment without having to modify files that are under git control.

So, I wonder, when this possibility was disabled recently in https://github.com/ArkEcosystem/core/pull/2100, was it on purpose?

https://github.com/ArkEcosystem/core/pull/2100#issuecomment-465062922

PR #2129, opened 2019-02-22
refactor(core-p2p):  reduce logging noise

supaiku0:
## Proposed changes

The recent change to channel each new peer through `acceptNewPeer` made it quite verbose. This PR also silences most of the peer verifier messages, if anyone is really interested in them they can be enabled via the undocumented env variable `CORE_P2P_PEER_VERIFIER_DEBUG_EXTRA`.

## Types of changes

## Checklist

### DIFF
--- a/packages/core-p2p/src/peer-verifier.ts
+++ b/packages/core-p2p/src/peer-verifier.ts
@@ -132,7 +132,7 @@ export class PeerVerifier {
             return false;
         }
 
-        this.log(Severity.INFO, `peer's claimed state is invalid: ${this.anyToString(claimedState)}`);
+        this.log(Severity.DEBUG_EXTRA, `peer's claimed state is invalid: ${this.anyToString(claimedState)}`);
 
         return true;
     }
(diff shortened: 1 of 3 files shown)


packages/core-p2p/src/peer-verifier.ts
```diff
@@ -505,7 +505,7 @@ export class PeerVerifier {
         }
 
         this.log(
-            Severity.INFO,
+            Severity.DEBUG_EXTRA,
```
vasild: This message, and the ones above, designate a likely attack from that peer. Should it really be never displayed, not even in DEBUG mode?

DEBUG_EXTRA messages will be printed only if CORE_P2P_PEER_VERIFIER_DEBUG_EXTRA is defined in the environment, which I gather is "never" for normal node operators.

OTOH we print useless messages like:

[2019-02-21 20:30:54][INFO]: Downloaded 1 new block accounting for a total of 0 transactions
(it does not say which block was downloaded (height, block id), nor from which peer it did download it)

[2019-02-21 20:28:52][INFO]: Block download finished

[2019-02-21 20:28:52][DEBUG]: event 'NOTSYNCED': {"syncWithNetwork":"processFinished"} -> {"syncWithNetwork":"downloadBlocks"} -> actions: [downloadBlocks]

With this patch, we give higher severity to the above `event 'NOTSYNCED'` message, than to a possible attack, or network misconfiguration or fork (all of the above messages which were changed from INFO to DEBUG_EXTRA).

PR #2145, opened 2019-02-24
feat(core): configuration and channel support for the CLI

faustbrian:
## Proposed changes

- Adds support for npm channels to allow switching from alpha to beta, beta to rc and so on.
- Adds support for CLI configuration to allow other configuration values like setting a token.

## Types of changes

## Checklist

### DIFF
--- a/packages/core/src/commands/config/cli.ts
+++ b/packages/core/src/commands/config/cli.ts
@@ -1,5 +1,8 @@
 import { flags } from "@oclif/command";
+import cli from "cli-ux";
+import { removeSync } from "fs-extra";
 import { configManager } from "../../helpers/config";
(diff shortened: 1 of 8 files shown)


packages/core/src/commands/config/cli.ts
```diff
@@ -0,0 +1,66 @@
...
+$ ark config:cli --channel=mine
+`,
+    ];
+
+    public static flags: Record<string, any> = {
+        token: flags.string({
+            description: "the name of the token that should be used",
+        }),
+        channel: flags.string({
+            description: "the name of the token that should be used",
```
dated:
```suggestion
            description: "the name of the channel that should be used",
```

Issue #1570, opened 2018-12-04
Increase vendor field length to 255 bytes


faustbrian:

ignatalexeyenko:
Hello,

I've did some research on the subject to better understand how the field is being used.
Since I understood the use cases - I come with a suggestion to grow field beyond the 255 limit.

The use cases, as I understood them are following:

## 1. Using the field as a note

Typical use cases:
- Human readable piece of information
- An ID to link transaction with an external system. For instance, could be Order ID or a Token in e-commerce system to match payment confirmation with a transaction (while using ARK Pay)

## 2. Argument to be used in ARK Contract Execution Services (ACES)

While this is an extension of a case 1, it's important to review it separately as ACES might use specific standard for serialisation parameters in call.


## Concerns
I personally concerned with a 255 limit. 
- It leaves too little room for passing multiple arguments for ACES / other systems
- While it is being said (Slack #development channel) _255 is the current limit because it allows storing of most hashes produced by hashing schemes along with some extra detail about the protocol being used_ it still leaves little room for using multiple hashes / addresses in a single call.
- It is suggested to use IPFS for larger bits of data, yet using external system increases architectural complexity greatly, which does not play along with a Push Button Blockchain concepts.

To address these - I'm suggesting to grow the field beyond 255 bytes, e.g. to 1024 bytes limit, could be even more. It's important limit is defined, the limit does not greatly impact amount of transactions in the block when used at projected capacity. As bigger limits -> fatter transactions -> less tranasaction fit into a fixed size block.

Arguments supporting this:
- With dynamic fees, the field would not likely be abused - you have to pay to slow down the network
- It easier to keep the data on-chain without need of building third-party integrations.

## Final thoughts
- Increasing the size causes blockchain size to grow faster. It can be an issue in a long run.
- Growing this to 255 now may still allow this to grow the field further, when needed. Currently it could be a premature optimisation.
- Shrinking the field later would be harder to implement - just in case big data blocks might already be stored while the limits wre larger.

PR #2159, opened 2019-02-26
feat(crypto): increase vendor field length to 255 bytes

supaiku0:
## Proposed changes

Resolves https://github.com/ArkEcosystem/core/issues/1570

Allows for 255 bytes long vendor fields at/after the specified milestones.

## Types of changes

## Checklist


packages/crypto/src/networks/devnet/milestones.json
```diff
@@ -46,5 +46,9 @@
     {
         "height": 950000,
         "ignoreInvalidSecondSignatureField": false
+    },
+    {
+        "height": 1750000,
+        "vendorField255": true
```
faustbrian: I would instead introduce a `vendorFieldLength` constant that defaults to `64/65` so we can increase the length at any time with a check for a single constant.

j-a-m-l:
@supaiku0 would it be possible to create a constant on the network config that indicates the maximum vendor field length to make it configurable and to allow clients use it instead of harcoding it?

PR #2184, opened 2019-02-28
feat(core-api): more params for delegates search endpoint

dated:
## Proposed changes

Adds support for more options to the delegates search endpoint.

## Types of changes

## Checklist


packages/core-database/src/repositories/utils/filter-rows.ts
```diff
@@ -40,13 +42,22 @@ export = <T>(rows: T[], params, filters) =>
...
                     let isLessThan = true;
+                    let value = item[elem];
+
+                    if (elem === "approval") {
+                        value = delegateCalculator.calculateApproval(item);
+                    } else if (elem === "productivity") {
+                        value = delegateCalculator.calculateProductivity(item);
+                    } else if (elem === "forgedTotal") {
+                        value = delegateCalculator.calculateForgedTotal(item);
+                    }
```
dated: I don't like this very much in here. Another option would be to manipulate the delegates array beforehand by adding the respective properties (approval / productivity / forgedTotal). What do you say @faustbrian?

faustbrian: Would need to be benchmarked what has a bigger performance hit as the manipulation would always happen while the call to this API endpoint and access to the data happens occasionally.

dated: It could be added into [`getLocalDelegates()`](https://github.com/ArkEcosystem/core/blob/e04596c2d8b40a8c1385c35e1ebfd21d9d5c1f74/packages/core-database/src/repositories/delegates-business-repository.ts#L18) by passing the requested params and manipulating on demand only.

faustbrian: Will test how much it impacts performance if you add it there.

dated:
[`objectHasSomeOwnProperty`](https://github.com/ArkEcosystem/core/blob/91af4c2f74f09b01c926edc42b3d7c1a67cb2f33/packages/core-database/src/repositories/delegates-business-repository.ts#L199) _could_ be moved to the utils as that kind of check is performed elsewhere also.

I'll add the respective API tests tonight.

dated:
```
  ● Guard › __determineOffence › should return a 1 year suspension for "Blacklisted"

    expect(received).toBe(expected) // Object.is equality

    Expected: 525600
    Received: 527040
```

is this fixed on `develop`?

faustbrian:
It's an odd issue caused by the upcoming leap year it seems, you can copy https://github.com/ArkEcosystem/core/commit/884c8946f7b7fa766ccd39c86e14582023ef445b to develop.

faustbrian:
@dated SQLite build failed on CircleCI. Could you also update the docs with the new parameters?

dated:
All good now! PR for the docs has been submitted.

PR #2205, opened 2019-03-06
feat(core-api): add active delegates endpoint

dated:
## Proposed changes

Adds an additional endpoint to the delegates API to retrieve the active delegates at a certain height (use case: display rank changes in the explorer).

The `getActiveAtHeight` method was not being used anywhere, hence it took the liberty to modify it to the needs of this PR.

## Types of changes

## Checklist

### DIFF
--- a/packages/core-api/src/versions/2/delegates/methods.ts
+++ b/packages/core-api/src/versions/2/delegates/methods.ts
@@ -4,7 +4,7 @@ import { orderBy } from "@arkecosystem/utils";
 import Boom from "boom";
 import { blocksRepository } from "../../../repositories";
 import { ServerCache } from "../../../services";
-import { paginate, respondWithResource, toPagination } from "../utils";
+import { paginate, respondWithResource, respondWithCollection, toPagination } from "../utils";
 
 const databaseService = app.resolvePlugin<Database.IDatabaseService>("database");
 
@@ -15,6 +15,16 @@ const index = async request => {
     });
 
     return toPagination(request, delegates, "delegate");
+};
+
+const active = async request => {
+    const delegates = await databaseService.delegates.getActiveAtHeight(request.query.height);
+
+    if (!delegates.length) {
+        return Boom.notFound("Delegates not found");
+    }
+
+    return respondWithCollection(request, delegates, "delegate");
 };
 
 const show = async request => {
@@ -86,6 +96,9 @@ export function registerMethods(server) {
         .method("v2.delegates.index", index, 8, request => ({
             ...request.query,
             ...paginate(request),
+        }))
+        .method("v2.delegates.active", active, 30, request => ({
+            ...request.query
         }))
         .method("v2.delegates.show", show, 8, request => ({ id: request.params.id }))
         .method("v2.delegates.search", search, 30, request => ({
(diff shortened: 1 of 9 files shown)


packages/core-p2p/src/peer-verifier.ts
```diff
@@ -324,7 +324,7 @@ export class PeerVerifier {
 
     /**
      * Get the delegates for the given round.
-     * @param {Object} round round to get delegates for
+     * @param {Object} round to get delegates for
      * @return {Object} a map of { publicKey: delegate, ... } of all delegates for the given round
      */
     private async getDelegates(round: any): Promise<any> {
```
faustbrian:
```suggestion
    private async getDelegatesByRound(round: any): Promise<any> {
```

packages/core-api/src/versions/2/delegates/methods.ts
```diff
@@ -87,6 +97,9 @@ export function registerMethods(server) {
             ...request.query,
             ...paginate(request),
         }))
+        .method("v2.delegates.active", active, 30, request => ({
```
faustbrian: The 30 seconds could probably be done as `blockTime * delegateCount` so it updates once a round.

packages/core-api/src/versions/2/delegates/methods.ts
```diff
@@ -82,11 +93,17 @@ const voterBalances = async request => {
 };
 
 export function registerMethods(server) {
+    const blockTime = config.getMilestone().blocktime;
```
faustbrian:
```suggestion
    const { activeDelegates, blockTime } = config.getMilestone();
```

:trollface:

dated: i'm not falling for your traps :danstop:

packages/core-api/src/versions/2/delegates/methods.ts
```diff
@@ -17,6 +18,16 @@ const index = async request => {
     return toPagination(request, delegates, "delegate");
 };
 
+const active = async request => {
+    const delegates = await databaseService.delegates.getActiveAtHeight(request.query.height);
```
faustbrian: I think `request.query.height` should default to the height of the last block.

dated: Yea, I'll add that. Sounds like the right thing to do.

packages/core-api/src/versions/2/delegates/methods.ts
```diff
@@ -17,6 +18,18 @@ const index = async request => {
     return toPagination(request, delegates, "delegate");
 };
 
+const active = async request => {
+    const height = this.blockchain.getLastHeight();
+
+    const delegates = await databaseService.delegates.getActiveAtHeight(request.query.height || height);
```
faustbrian:
```suggestion
    const delegates = await databaseService.delegates.getActiveAtHeight(request.query.height || this.blockchain.getLastHeight());
```

Otherwise the `this.blockchain.getLastHeight()` always get called even if a height was provided.

dated:
An unrelated test is currently failing @faustbrian.

```
FAIL  __tests__/integration/core-transaction-pool/guard.test.ts

● Transaction Guard › validate › Sign a transaction then change some fields shouldn't pass validation › should not validate when changing fields after signing - delegate registration
```

PR #2208, opened 2019-03-07
refactor(core-forger): Increase timeout for getting the network state

vasild:
The 2 seconds timeout is too short and can unnecessary blow up the
forging process in cases where it is possible to forge successfully.

Set it to an arbitrary 7 seconds which is block time minus 1 second for
other activities.

Elaborate log messages by explicitly saying "Will not forge" + an
explanation "why" when the code would decide not to forge.

## Types of changes

## Checklist

### DIFF
--- a/packages/core-forger/src/client.ts
+++ b/packages/core-forger/src/client.ts
@@ -87,11 +87,12 @@ export class Client {
      */
     public async getNetworkState(): Promise<NetworkState> {
         try {
-            const response = await this.__get(`${this.host}/internal/network/state`);
+            const response = await this.__get(`${this.host}/internal/network/state`, 7000);
             const { data } = response.data;
 
             return NetworkState.parse(data);
         } catch (e) {
+            this.logger.error(`Could not retrieve network state: ${this.host}/internal/network/state: ${e.message}`);
             return new NetworkState(NetworkStateStatus.Unknown);
         }
     }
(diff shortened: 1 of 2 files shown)


faustbrian:
This is only a band-aid that hides the issue that the peer verifier is unable to complete all requests in a reasonable amount of time.

Timeouts in the forger should be kept as low as possible to avoid missing a block simply because the request to your relay took too long.

vasild:
> This is only a band-aid that hides the issue that the peer verifier is unable to complete all requests in a reasonable amount of time.

No. The peer verifier is engaged from the `ping()` method and this patch does not touch that timeout - it is still 1.5 seconds.

Notice also that if a peer is at the same height+blockid then the peer verifier will do no extra network calls - that is, timing is the same with and without peer verifier (assuming network calls is where most of the time is spent). See "Case 3" and "Case 5" in https://github.com/ArkEcosystem/core/blob/develop/packages/core-p2p/src/peer-verifier.ts#L69.

> Timeouts in the forger should be kept as low as possible to avoid missing a block simply because the request to your relay took too long.

But if the timeout is too low, and thus the request to `/internal/network/state` times out, then the forger misses a block too.

supaiku0:
Can you rebase this PR onto 2.2?

vasild:
> Can you rebase this PR onto 2.2?

Done.

PR #2218, opened 2019-03-10
feat(core): add restart flags to update command

dated:
## Proposed changes

Adds the option to pass flags to the update command to specify the restart behaviour without prompting for user input:

- `--restart`: restarts all (running) processes
- `--no-restart`: no process will be restarted
- `--restart-core`: restarts the core process
- `--restart-relay`: restarts the relay process
- `--restart-forger`: restarts the forger process

## Types of changes

## Checklist

### DIFF
--- a/packages/core/src/commands/command.ts
+++ b/packages/core/src/commands/command.ts
@@ -284,19 +284,26 @@ export abstract class BaseCommand extends Command {
         return this.getNetworks().map(network => ({ title: network, value: network }));
     }
 
-    protected async restartProcess(processName: string) {
+    protected async restartProcess(processName: string, showPrompt: boolean = true) {
         if (processManager.isRunning(processName)) {
-            await confirm(`Would you like to restart the ${processName} process?`, () => {
-                try {
-                    cli.action.start(`Restarting ${processName}`);
+            if (showPrompt) {
+                await confirm(`Would you like to restart the ${processName} process?`, () => {
+                    this.restartRunningProcess(processName);
+                });
+            } else {
+                this.restartRunningProcess(processName);    
+            }
+        }
+    }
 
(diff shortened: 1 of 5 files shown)


packages/core/src/commands/command.ts
```diff
@@ -284,19 +284,26 @@ export abstract class BaseCommand extends Command {
...
+                await confirm(`Would you like to restart the ${processName} process?`, () => {
+                    this.restartRunningProcess(processName);
+                });
+            } else {
+                this.restartRunningProcess(processName);    
+            }
+        }
+    }
+
+    protected restartRunningProcess(processName: string) {
```
faustbrian:
```suggestion
    protected restartRunningProcess(processName: string): void {
```

packages/core/src/commands/command.ts
```diff
@@ -284,19 +284,26 @@ export abstract class BaseCommand extends Command {
...
-                } finally {
-                    cli.action.stop();
-                }
-            });
+            if (showPrompt) {
+                await confirm(`Would you like to restart the ${processName} process?`, () => {
+                    this.restartRunningProcess(processName);
+                });
+            } else {
+                this.restartRunningProcess(processName);
```
faustbrian: trailing space

packages/core/src/commands/command.ts
faustbrian: The method name implies that the `if (processManager.isRunning(processName)) {` check is performed in it.

dated: Shall I switch out the names of the `restartProcess` and `restartRunningProcess` methods?

dated:
Does the PR need approval before I can switch from draft to normal? 🤔

PR #2221, opened 2019-03-11
feat(core-tester-cli): add make:block command

faustbrian:
## Proposed changes

Sometimes I need to generate some fake blocks to test something, been doing that so far with a script but it can now be used via `tester make:block`.

**Will switch the tests from axios to got/nock after https://github.com/ArkEcosystem/core/pull/2203 has been merged.**

## Types of changes

## Checklist


vasild:
Could end up with:
```
        "verification": {
            "verified": false,
            "errors": [
                "Encountered duplicate transaction: 8309ffb072b2333284e0729bb78ea1168d61e329c690d6c7a4e7b88b6eacb3a3"
            ]
        }
```
if more than one transaction is generated and the timestamps of the transactions are equal.

packages/core-tester-cli/src/commands/make/block.ts
```diff
@@ -0,0 +1,94 @@
...
+                        ...flags,
+                        ...{
+                            amount: flags.transactionAmount + 10,
+                            transferFee: flags.transactionFee,
+                            recipient: Address.fromPassphrase(flags.passphrase),
+                        },
+                    }),
+                );
+
+                await delay(100);
```
vasild: Why this?

vasild:
```
commit 638394f3 (HEAD -> feat/blocks-gen)
Parent: 925fcbab
Author:     Vasil Dimov <vd@FreeBSD.org>
AuthorDate: Mon Mar 11 14:28:58 2019 +0100
Commit:     Vasil Dimov <vd@FreeBSD.org>
CommitDate: Mon Mar 11 14:28:58 2019 +0100
gpg: Signature made Mon Mar 11 14:31:04 2019 CET
gpg:                using RSA key E64D8D45614DB07545D9CCC154DF06F64B55CBBF
gpg: Good signature from "Vasil Dimov <vd@myforest.net>" [ultimate]
gpg:                 aka "Vasil Dimov <vd@ark.io>" [ultimate]
gpg:                 aka "Vasil Dimov <vd@FreeBSD.org>" [ultimate]
gpg:                 aka "Vasil Dimov <vasild@gmail.com>" [ultimate]


    feat(core-tester-cli): Make it possible to override the base block
    
    Add the ability to manually set the block on which newly generated
    blocks will be based.

diff --git a/packages/core-tester-cli/src/commands/make/block.ts b/packages/core-tester-cli/src/commands/make/block.ts
index 2c3330a4..a27c07a5 100644
--- a/packages/core-tester-cli/src/commands/make/block.ts
+++ b/packages/core-tester-cli/src/commands/make/block.ts
@@ -29,24 +29,32 @@ export class BlockCommand extends BaseCommand {
             default: 0.1,
         }),
         passphrase: flags.string({
             description: "passphrase of the forger",
             default: "clay harbor enemy utility margin pretty hub comic piece aerobic umbrella acquire",
         }),
+        previousBlock: flags.string({
+            description: `Previous block to base the generated block(s) on. For example: '{ "height": 50, "id": "123", "idHex": "7b" }'`,
+            default: "The genesis block"
+        }),
         write: flags.boolean({
             description: "write the blocks to the disk",
         }),
     };
 
     public async run(): Promise<models.IBlock[]> {
         const { flags } = await this.make(BlockCommand);
 
         configManager.setFromPreset(flags.network as NetworkName);
 
-        // We always start with the genesis block!
-        let previousBlock = configManager.get("genesisBlock");
+        let previousBlock
+        if (flags.previousBlock) {
+            previousBlock = JSON.parse(flags.previousBlock);
+        } else {
+            previousBlock = configManager.get("genesisBlock");
+        }
 
         const blocks: models.IBlock[] = [];
 
         for (let i = 0; i < flags.number; i++) {
             const milestone = configManager.getMilestone(previousBlock.height);
             const delegate = new models.Delegate(flags.passphrase, configManager.get("pubKeyHash"));
```
ok to push to `feat/blocks-gen`?

packages/core-tester-cli/src/commands/make/block.ts
faustbrian: To avoid duplicate transactions caused by timestamps, anything below 100 causes duplicates.

faustbrian:
@vasild apply it 👍

packages/core-tester-cli/src/commands/make/block.ts
vasild: Then this is not enough as I still got duplicate txs - anything below 1000ms (1s) has a chance of ending up in the same second. But adding 1s delay here would make it too slow - if many transactions are to be generated.

What about using different amount or different sender or different recipient in order to avoid duplicates?

faustbrian: Changing the recipient would probably be already enough, ideally we only use `genesisBlock.transactions.map(t => Address.fromPublicKey(t.senderPublicKey))` or `genesisBlock.transactions.map(t => t.recipientId)` so we can be sure the wallets exist.

PR #2225, opened 2019-03-11
Ts config

XPH-Dev:
Proposed changes
Resolves /milestone/3 typescript
Fixes a bug in core-container and resolveOptions that could result in null resolutions or providing options for plugins that are not even registered in the container
Fixed some return and value assignment issues where for example a BigNumber object was expected but a Number was assigned (JS didn't catch this and we had some tests that were passing even though they should have been failing)
Moved the peers.json config into the defaults of core-p2p as they are only used there and not by anything else (dc582fe). This means peers.json is now solely used to store the peer lists, more changes will follow in an upcoming PR.
Resolved various minor bugs that showed up during the migration
Added the yarn setup command to install, bootstrap and build all packages
Merged core-transaction-pool and core-transaction-pool-mem
Have been forging for some time with this branch on devnet now and had no errors occur so far.

Types of changes
  Refactoring (improve a current implementation without adding a new feature or fixing a bug)
  Build (changes that affect the build system)


XPH-Dev:
wrong remote sorry

PR #2229, opened 2019-03-12
fix(core-api): properly sort semver versions

luciorubeens:
## Proposed changes

Sort the peers by version.

## Types of changes

## Checklist


.circleci/config.yml
```diff
@@ -41,19 +41,25 @@ jobs:
             - ./packages/core-container/node_modules
             - ./packages/core-database/node_modules
             - ./packages/core-database-postgres/node_modules
+            - ./packages/core-debugger-cli/node_modules
```
faustbrian: You seem to be on an old clone as all those plugins don't exist anymore.

luciorubeens: Fixed

PR #2231, opened 2019-03-12
refactor(core-forger): always use latest available height

supaiku0:
## Proposed changes

Always use the height as reported by the network state and simplify quorum calculation.

## Types of changes

## Checklist

### DIFF
--- a/packages/core-p2p/src/network-state.ts
+++ b/packages/core-p2p/src/network-state.ts
@@ -39,7 +39,7 @@ class QuorumDetails {
      */
 
     /**
-     * Number of peers with a different last block id.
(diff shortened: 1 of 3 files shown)


packages/core-p2p/src/network-state.ts
```diff
@@ -53,7 +41,7 @@ class QuorumDetails {
     /**
      * Number of peers with a different last block id.
      */
-    public peersDifferentBlockId = 0;
+    public peersForked = 0;
```
vasild: The comment "Number of peers with a different last block id" needs an update. Something among the lines of "number of peers that are on a different chain (forked)"

PR #2245, opened 2019-03-14
fix(core-database): return array instead of joined string from applyOrder

dated:
## Proposed changes

The current logic of `applyOrder` breaks when the manipulated iteratee is a function.

## Types of changes

## Checklist

## Further comments

There are some other things that don't appear to work correctly with the delegate (maybe also wallets?) API endpoints, I don't know yet if they are related to this, for instance the missed blocks count is not set correctly, which in turn results in a skewed productivity.

```
curl 'http://xxx:4003/api/v2/delegates?page=1&limit=1&orderBy=productivity:desc' -H 'api-version: 2' -H "Content-Type: application/json"
```

```
{
  "meta": {
    "count": 1,
    "pageCount": 340,
    "totalCount": 340,
    "next": "/api/v2/delegates?page=2&limit=1&orderBy=productivity%3Adesc",
    "previous": null,
    "self": "/api/v2/delegates?page=1&limit=1&orderBy=productivity%3Adesc",
    "first": "/api/v2/delegates?page=1&limit=1&orderBy=productivity%3Adesc",
    "last": "/api/v2/delegates?page=340&limit=1&orderBy=productivity%3Adesc"
  },
  "data": [
    {
      "username": "genesis_34",
      "address": "DA4RAKRngGDfgEA7L6nQjYVEKcZPgi8EdP",
      "publicKey": "03b906102928cf97c6ddeb59cefb0e1e02105a22ab1acc3b4906214a16d494db0a",
      "votes": 7678989488674,
      "rank": 12,
      "blocks": {
        "produced": 37626,
        "missed": 0,
        "last": {
          "id": "10085216575111962936",
          "height": 1778534,
          "timestamp": {
            "epoch": 62482002,
            "unix": 1552583202,
            "human": "2019-03-14T17:06:42.000Z"
          }
        }
      },
      "production": {
        "approval": 0.06,
        "productivity": 100
      },
      "forged": {
        "fees": 182329488674,
        "rewards": 7482200000000,
        "total": 7664529488674
      }
    }
  ]
}
```


faustbrian:
On `develop` the missing blocks are no longer stored in the database as we dropped wallets from the database so they are only persisted in-memory.

Currently thinking about a solution for that but it's not pressing as productivity is not used for anything but visualisation.

dated:
Understood! Well, this fixes sorting by `approval` at least (the desktop wallet uses sorting by approval and productivity, where it's currently broken).

PR #2246, opened 2019-03-15
fix(tests): Reference the home directory via HOME, not ~

vasild:
"~" is expanded by the shell, if we use it in JavaScript it will end up
creating directories like "./packages/core-p2p/~/.core/"

Use HOME from the environment instead. If it is not set, then trying
"/.core/..." is fine.

### DIFF
--- a/__tests__/utils/helpers/container.ts
+++ b/__tests__/utils/helpers/container.ts
@@ -8,7 +8,7 @@ import * as path from "path";
 export async function setUpContainer(options: any): Promise<Container.IContainer> {
     options.network = options.network || "testnet";
 
-    process.env.CORE_PATH_DATA = options.data || "~/.core";
+    process.env.CORE_PATH_DATA = options.data || `${process.env.HOME}/.core`;
     process.env.CORE_PATH_CONFIG = options.config
         ? options.config
         : path.resolve(__dirname, `../config/${options.network}`);


vasild:
GitHub markup kicked in. Here is an unformatted commit message:
```
commit 4caf03c6e (HEAD -> home-dir-fixup, ArkEcosystem/core/home-dir-fixup, serialized-p2p)
Parent: bda844d9b
Author:     Vasil Dimov <vd@FreeBSD.org>
AuthorDate: Fri Mar 15 09:31:55 2019 +0100
Commit:     Vasil Dimov <vd@FreeBSD.org>
CommitDate: Fri Mar 15 09:31:55 2019 +0100
gpg: Signature made Fri Mar 15 09:37:26 2019 CET
gpg:                using RSA key E64D8D45614DB07545D9CCC154DF06F64B55CBBF
gpg: Good signature from "Vasil Dimov <vd@myforest.net>" [ultimate]
gpg:                 aka "Vasil Dimov <vd@ark.io>" [ultimate]
gpg:                 aka "Vasil Dimov <vd@FreeBSD.org>" [ultimate]
gpg:                 aka "Vasil Dimov <vasild@gmail.com>" [ultimate]


    fix(tests): Reference the home directory via HOME, not ~

    "~" is expanded by the shell, if we use it in JavaScript it will end up
    creating directories like "./packages/core-p2p/~/.core/"

    Use HOME from the environment instead. If it is not set, then trying
    "/.core/..." is fine.
```

PR #2273, opened 2019-03-21
feat: websockets (socketcluster) for p2p communication

### LINKED ISSUE #1589: Migrate P2P API to Websockets
Will resolve https://github.com/ArkEcosystem/core/issues/1151

http://socketcluster.io

air1one:
## Proposed changes

Use websockets instead of current http api for p2p communication.

Resolves #1589.

## Types of changes

## Checklist


packages/core-forger/src/defaults.ts
```diff
@@ -1,3 +1,8 @@
 export const defaults = {
-    hosts: [`http://127.0.0.1:${process.env.CORE_P2P_PORT || 4002}`],
+    hosts: [
```
faustbrian: The `packages/core/bin/config/**/plugins.js` files also need to be updated. We can simply remove the defaults from the `plugins.js` files as the defaults in this file should rarely need to be changed.

PR #2300, opened 2019-03-25
feat(core-snapshots): rollback by a given amount of blocks

faustbrian:
## Proposed changes

Allows to run `ark snapshot:rollback --number=51` instead of `ark snapshot:rollback --height=123456789` to perform a rollback.

## Types of changes

## Checklist

### DIFF
--- a/packages/core-snapshots/src/manager.ts
+++ b/packages/core-snapshots/src/manager.ts
@@ -79,7 +79,7 @@ export class SnapshotManager {
         this.database.close();
     }
 
-    public async rollbackChain(height) {
+    public async rollbackByHeight(height) {
         const lastBlock = await this.database.getLastBlock();
         const config = app.getConfig();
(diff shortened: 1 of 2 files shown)


packages/core-snapshots/src/manager.ts
```diff
@@ -79,38 +79,45 @@ export class SnapshotManager {
         this.database.close();
     }
 
-    public async rollbackChain(height) {
-        const lastBlock = await this.database.getLastBlock();
-        const config = app.getConfig();
-        const maxDelegates = config.getMilestone(lastBlock.height).activeDelegates;
+    public async rollbackByHeight(height) {
+        if (!height) {
```
dated: What about negatives?

```suggestion
        if (!height || height < 0) {
```

faustbrian: 👍

PR #2302, opened 2019-03-25
fix(core-snapshots): always load core-database-postgres

supaiku0:
## Proposed changes

`core-snapshots` implemented migrations itself which corrupted the database. Now instead
always load the `core-database-postgres` plugin which already performs the migrations and properly checks wether they were already applied or not.

## Types of changes

## Checklist

### DIFF
--- a/packages/core-snapshots/src/plugin.ts
+++ b/packages/core-snapshots/src/plugin.ts
@@ -16,6 +16,6 @@ export const plugin: Container.PluginDescriptor = {
 
         const databaseService = container.resolvePlugin<Database.IDatabaseService>("database");
         const connection = databaseService.connection as PostgresConnection;
-        return await manager.make(connection);
+        return manager.make(connection);
     },
 };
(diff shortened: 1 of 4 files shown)


packages/core-snapshots/src/plugin.ts
```diff
@@ -15,10 +15,7 @@ export const plugin: Container.PluginDescriptor = {
         const manager = new SnapshotManager(options);
 
         const databaseService = container.resolvePlugin<Database.IDatabaseService>("database");
-        if (!!databaseService) {
-            const connection = databaseService.connection as any;
-            return await manager.make(connection as PostgresConnection);
-        }
-        return await manager.make(null);
+        const connection = databaseService.connection as PostgresConnection;
+        return await manager.make(connection);
```
faustbrian:
```suggestion
        return manager.make(databaseService.connection as PostgresConnection);
```

PR #2306, opened 2019-03-26
fix(core-api): adjust schema validation of blocks show

dated:
## Proposed changes

The blocks/show endpoint does currently not accept block ids of the new sha256 format. There are no fixtures for blocks with the new id format i believe?

## Types of changes

## Checklist


faustbrian:
I will update the fixture generator soon and add new block fixtures.

PR #2307, opened 2019-03-26
perf(crypto): skip transaction verification overhead

supaiku0:
## Proposed changes

Fetching transactions from the database is super slow, because we always perform crypto verify,
id calculation and other overhead. All of this shouldn't be necessary, since all the steps are already performed before a transaction is even allowed to enter the database in the first place.

This PR introduces a new function called `fromBytesUnsafe` which deserializes a transaction
without all the overhead.

Non-scientific benchmark:
 (400 blocks, 15000+ transactions)
```
GET localhost:4002/peer/blocks?lastBlockHeight=269572

GET BLOCKS: 12493.502ms	  | no change
GET BLOCKS: 3695.236ms 	  | skip verify
GET BLOCKS: 1985.264ms 	  | skip verify, skip id
```
This explains many block download timeouts and nodes that get stuck around height 270000.

## Types of changes

## Checklist

### DIFF
--- a/__tests__/unit/crypto/transactions/transaction.test.ts
+++ b/__tests__/unit/crypto/transactions/transaction.test.ts
@@ -146,6 +146,17 @@ describe("Models - Transaction", () => {
         });
     });
 
+    describe("fromBytesUnsafe", () => {
+        it("should be ok", () => {
+            const bytes = Transaction.toBytes(transactionData);
+            const id = transactionData.id;
+
+            const transaction = Transaction.fromBytesUnsafe(bytes, id);
+            expect(transaction).toBeInstanceOf(Transaction);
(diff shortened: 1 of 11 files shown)


n4ru:
Can this be made optional via a config flag? Seems like a very critical step to bypass, especially when it comes to sharing snapshots (which some people - regardless of whether or not it should be done - still do).

supaiku0:
@n4ru 
Compared to before, even though crypto verify and id calculation were performed, the results weren't used. Meaning there's no difference compared to now except that we no longer waste time performing "useless" computations.

Additionally, we perform integrity checks during node startup which ensures there are no balance inconsistencies otherwise it blows up.

Yes, the integration checks can be improved and snapshots should always be verified, but this is a different topic since ultimately nothing changed in this regard.

PR #2313, opened 2019-03-26
fix(core-database-postgres): add EQ operator to block height

dated:
## Proposed changes

The block model does currently not allow the `equals` operator, passing a blockheight to the `blocks/{id}` endpoint will therefore result in a 404 error.

## Types of changes

## Checklist

### DIFF
--- a/packages/core-database-postgres/src/models/block.ts
+++ b/packages/core-database-postgres/src/models/block.ts
@@ -27,7 +27,11 @@ export class Block extends Model {
             },
             {
                 name: "height",
-                supportedOperators: [Database.SearchOperator.OP_LTE, Database.SearchOperator.OP_GTE],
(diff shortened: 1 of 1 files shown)


dated:
Looks like this didn't have it's intended result, as it still matches against the the id only, see https://dexplorer.ark.io/api/v2/blocks/1 for instance.

[core-database-postgres/src/repositories/blocks.ts#L14](https://github.com/ArkEcosystem/core/blob/develop/packages/core-database-postgres/src/repositories/blocks.ts#L14) could be adjusted so that the query is built like in the [core-api blocks repository](https://github.com/ArkEcosystem/core/blob/develop/packages/core-api/src/repositories/blocks.ts#L53) instead of looking up only by id.

Any recommendation / preference @faustbrian?

PR #2316, opened 2019-03-26
fix(core-database): map block to transaction in findById

dated:
## Proposed changes

The `findById` method called by the `transactions/{id}` API endpoint does not map the block to the transaction.

[untested]

## Types of changes

## Checklist

### DIFF
--- a/packages/core-database/src/repositories/transactions-business-repository.ts
+++ b/packages/core-database/src/repositories/transactions-business-repository.ts
@@ -59,7 +59,8 @@ export class TransactionsBusinessRepository implements Database.ITransactionsBus
     }
 
     public async findById(id: string) {
-        return this.databaseServiceProvider().connection.transactionsRepository.findById(id);
+        const rows = return this.databaseServiceProvider().connection.transactionsRepository.findById(id);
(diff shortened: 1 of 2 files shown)


packages/core-database/src/repositories/transactions-business-repository.ts
```diff
@@ -59,7 +59,8 @@ export class TransactionsBusinessRepository implements Database.ITransactionsBus
     }
 
     public async findById(id: string) {
-        return this.databaseServiceProvider().connection.transactionsRepository.findById(id);
+        const rows = return this.databaseServiceProvider().connection.transactionsRepository.findById(id);
```
faustbrian:
```suggestion
        const rows = await this.databaseServiceProvider().connection.transactionsRepository.findById(id);
```

dated: Yea, thanks :D

dated:
```
  ● Transactions Business Repository › findById › should invoke findById on db repository

    TypeError: Cannot read property 'get' of undefined
```

Will have a look once home.

dated:
All good now @faustbrian

PR #2321, opened 2019-03-27
feat: add more block and transaction pool events

faustbrian:
## Proposed changes

Adds 4 new events to make some relatively hard to grab data more accessible.

- **block.received** - Emitted when a new block is incoming
- **block.disregarded** - Emitted when a new block is disregarded
- **transaction.pool.added** - Emitted when transactions are added to the pool, always emits an array of transactions
- **transaction.pool.not-added** - Emitted when transactions are _not_ added to the pool, always emits an array of transactions
- **transaction.pool.removed** - Emitted when a transaction is removed by its id, only emits the ID of the transaction

@supaiku0 when testing you should attach a debugger/profiler on a long-run running process that keeps triggering those events as the node.js event emitter is notorious for causing memory leaks.

## Types of changes

## Checklist


zillionn:
Why `transaction.pool.not-added` but not `transaction.pool.rejected` as the API response?

PR #2322, opened 2019-03-27
fix(core): preserve existing logs and attach the last line when the log is modified

faustbrian:
## Proposed changes

@supaiku0 when testing this check if any lines are missing compared to `pm2 logs` and that it now behaves as closely as possible like `pm2 logs`.

## Types of changes

## Checklist

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -81,7 +81,7 @@
         "latest-version": "^4.0.0",
         "listr": "^0.14.3",
         "lodash.minby": "^4.6.0",
-        "nsfw": "^1.2.2",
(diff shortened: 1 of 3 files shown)


dated:
```
debounceMS: 250,
```

I assume this is the time `nsfw` waits until the callback is fired after an event happens? This probably needs to be lowered then, as multiple `actions.MODIFIED` events in the same array passed to the callback will result in multiple outputs of the same line and some lines missing?

faustbrian:
That is what the comment about missing lines was related to as it needs some testing to get the right value as logging happens quite fast at times.

PR #2338, opened 2019-03-29
refactor(core-database-postgres): add findByHeight to avoid confusion with the misleading findByHeights

faustbrian:
## Proposed changes

The `findByHeight` method in `core-database-postgres` at the moment has a very misleading name as the singular `height` in the same indicates that you will received a single block but it executes `manyOrNone` which means the result will be an array.

https://github.com/ArkEcosystem/core/blob/develop/packages/core-database-postgres/src/repositories/blocks.ts#L32

This PR provides a `findByHeight` and `findByHeights` method which return a single and multiple blocks. This will avoid confusion and unexpected behaviour like there currently is with the `findByHeight` method based on its name.

## Types of changes

## Checklist

### DIFF
--- a/packages/core-database-postgres/src/repositories/blocks.ts
+++ b/packages/core-database-postgres/src/repositories/blocks.ts
@@ -15,13 +15,27 @@ export class BlocksRepository extends Repository implements Database.IBlocksRepo
         return this.db.oneOrNone(sql.findById, { id });
     }
 
+    /**
+     * Find many blocks by their IDs.
+     * @param {String[]} ids
+     */
     public async findByIds(ids: string[]) {
-        const query = this.query
-            .select()
-            .from(this.query)
-            .where(this.query.id.in(ids))
-            .group(this.query.id);
-        return await this.findMany(query);
+        return this.findMany(
+            this.query
+                .select()
+                .from(this.query)
+                .where(this.query.id.in(ids))
+                .group(this.query.id),
+        );
+    }
+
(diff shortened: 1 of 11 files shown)


packages/core-database-postgres/src/repositories/blocks.ts
```diff
@@ -15,21 +15,35 @@ export class BlocksRepository extends Repository implements Database.IBlocksRepo
...
     }
 
     /**
      * Get all of the blocks at the given heights.
      * @param  {Array} heights the heights of the blocks to retrieve
      * @return {Promise}
      */
-    public async findByHeight(heights) {
+    public async findByHeights(heights: number[]) {
         return this.db.manyOrNone(sql.findByHeight, { heights });
```
vasild: s/sql.findByHeight/sql.findByHeights/

packages/core-database-postgres/src/repositories/blocks.ts
```diff
@@ -15,21 +15,35 @@ export class BlocksRepository extends Repository implements Database.IBlocksRepo
...
+                .group(this.query.id),
+        );
+    }
+
+    /**
+     * Get a block at the given height.
+     * @param  {Number} height the height of the blocks to retrieve
+     * @return {Promise}
+     */
+    public async findByHeight(height: number) {
```
vasild: I think this method is unused after the change - you may as well drop it.

PR #2341, opened 2019-03-30
refactor(core-forger): memory leak, log spam, cleanup

### LINKED ISSUE #2326: Memory leak:- ark-forger process.
**Describe the bug**
Multiple users have reported that the ark-forger process shows signs of a memory leak. A delegate's ark-forger was using 1.4GB of memory upon checking as shown below...

ark@gnia:~$ pm2 status
┌────────────┬────┬──────┬────────┬───┬───────┬────────────┐
│ Name       │ id │ mode │ status │ ↺ │ cpu   │ memory     │
├────────────┼────┼──────┼────────┼───┼───────┼────────────┤
│ ark-forger │ 2  │ fork │ online │ 0 │ 3%    │ 1.4 GB     │
│ ark-relay  │ 1  │ fork │ online │ 0 │ 14.9% │ 374.1 MB   │
└────────────┴────┴──────┴────────┴───┴───────┴────────────

**To Reproduce**
1. Start ark-forger process
2. Let the process run for weeks/months without restart

**Expected behavior**
For the ark-forger process to be using significantly less than 1.4GB of memory.

**Screenshots**

**Server (please complete the following information):**
 - OS: Ubuntu
 - Version: 18:04
However, this issue has been reported by multiple users with various versions of operating system& system specs.

**Additional context**

Delegate **alessio // fun** has created a patch file for this issue which he will send to me. Once received, I shall forward it to the dev team for review.

supaiku0:
## Proposed changes

Refactors the forger and addresses the following points:
- during startup, don't block when no host is responsive
- host selection now tries all configured hosts before throwing an error
- host selection no longer uses recursion
- no longer spams the log each possible event cycle when no host is available
- uses a timeout to schedule the next `__monitor()` call instead of promise recursion
- removes obsolete code

Resolves #2326 

The memory leak was caused by two things:
(https://github.com/nodejs/node/issues/6673)
1. promise recursion in `__monitor()`
2. promise recursion in `__chooseHost()`

The former is the lesser evil and only creeps up slowly over time, but once the forger relay becomes unresponsive the latter makes the memory leak really fast.

(It may be noted that the memory leak as described in #2326 seems to be almost non-present with Node.js 11 in the case of Core despite the provided script leaking in both versions)

## Types of changes

## Checklist

### DIFF
--- a/packages/core-forger/src/client.ts
+++ b/packages/core-forger/src/client.ts
@@ -3,7 +3,6 @@ import { Logger } from "@arkecosystem/core-interfaces";
 import { NetworkState } from "@arkecosystem/core-p2p";
 import { httpie } from "@arkecosystem/core-utils";
 import { models } from "@arkecosystem/crypto";
-import sample from "lodash.sample";
 import { URL } from "url";
 import { HostNoResponseError, RelayCommunicationError } from "./errors";
 
@@ -115,12 +114,20 @@ export class Client {
      * Chose a responsive host.
      */
     public async selectHost(): Promise<void> {
-        const host = sample(this.hosts);
-        try {
-            await this.get(`${host}/peer/status`);
-            this.host = host;
-        } catch (error) {
-            throw new HostNoResponseError(host);
+        let queriedHosts = 0;
+        for (const host of this.hosts) {
+            try {
+                await this.get(`${host}/peer/status`);
+                this.host = host;
+            } catch (error) {
+                if (queriedHosts === this.hosts.length - 1) {
+                    throw new HostNoResponseError(host);
+                } else {
+                    this.logger.warn(`Failed to get response from ${host}. Trying another host.`);
+                }
+            } finally {
+                queriedHosts++;
+            }
         }
     }
 
--- a/packages/core-forger/src/client.ts
+++ b/packages/core-forger/src/client.ts
@@ -114,7 +114,7 @@ export class Client {
     /**
      * Chose a responsive host.
      */
-    private async selectHost(): Promise<void> {
+    public async selectHost(): Promise<void> {
         const host = sample(this.hosts);
         try {
             await this.get(`${host}/peer/status`);
--- a/packages/core-forger/src/client.ts
+++ b/packages/core-forger/src/client.ts
@@ -45,14 +45,14 @@ export class Client {
     /**
      * Send the given block to the relay.
      */
-    public async broadcast(block: models.IBlockData): Promise<void> {
+    public async broadcast(block: models.IBlockData): Promise<IHttpieResponse<null>> {
         this.logger.debug(
             `Broadcasting forged block id:${block.id} at height:${block.height.toLocaleString()} with ${
                 block.numberOfTransactions
             } transactions to ${this.host}`,
         );
 
-        await this.post(`${this.host}/internal/blocks`, { block });
+        return this.post(`${this.host}/internal/blocks`, { block });
     }
 
     /**
--- a/packages/core-forger/src/client.ts
+++ b/packages/core-forger/src/client.ts
@@ -1,8 +1,8 @@
 import { app } from "@arkecosystem/core-container";
 import { Logger } from "@arkecosystem/core-interfaces";
-import { NetworkState } from "@arkecosystem/core-p2p";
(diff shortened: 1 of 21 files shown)


packages/core-forger/src/manager.ts
```diff
@@ -1,25 +1,28 @@
...
-    private usernames: any;
-    private isStopped: any;
+
+    private secrets: string[];
+    private network: networks.INetwork;
+    private client: Client;
+    private delegates: models.Delegate[];
+    private usernames: { [key: string]: string };
+    private isStopped: boolean;
+    private round: any;
```
faustbrian: Should add a struct/interface for the round data.

packages/core-forger/src/manager.ts
```diff
@@ -292,10 +255,31 @@ export class ForgerManager {
...
+            const delegates = this.delegates.map(
+                delegate => `${this.usernames[delegate.publicKey]} (${delegate.publicKey})`,
+            );
+
+            this.logger.debug(`Loaded ${pluralize("delegate", delegates.length, true)}: ${delegates.join(", ")}`);
+            this.initialized = true;
+        }
+    }
+
+    private async checkLater(timeout: number) {
```
faustbrian:
```suggestion
    private checkLater(timeout: number): void {
```

packages/core-forger/src/client.ts
```diff
@@ -37,93 +44,48 @@ export class Client {
 
     /**
      * Send the given block to the relay.
-     * @param  {(Block|Object)} block
-     * @return {Object}
      */
-    public async broadcast(block) {
+    public async broadcast(block: models.IBlockData): Promise<any> {
```
faustbrian: Should add a struct/interface for the expected data we want to return instead of using `any`.

packages/core-forger/src/client.ts
```diff
@@ -37,93 +44,48 @@ export class Client {
...
-        try {
-            await this.__chooseHost();
-
-            const response = await this.__get(`${this.host}/internal/rounds/current`);
-
-            return response.body.data;
-        } catch (e) {
-            return {};
-        }
+    public async getRound(): Promise<any> {
```
faustbrian: Should add a struct/interface for the expected data we want to return instead of using `any`.

packages/core-forger/src/client.ts
```diff
@@ -145,40 +107,43 @@ export class Client {
...
+                queriedHosts++;
             }
-
-            await this.__chooseHost(wait);
         }
     }
 
-    public async __get(url, timeout: number = 2000) {
-        return httpie.get(url, { headers: this.headers, timeout });
+    private async get(url, timeout: number = 2000) {
```
faustbrian: Should add a struct/interface for the expected return value, maybe add a `HttpieResponse` to `@arkecosystem/core-utils`.

packages/core-forger/src/client.ts
```diff
@@ -145,40 +107,43 @@ export class Client {
...
+        try {
+            return httpie.get(url, { headers: this.headers, timeout });
+        } catch (error) {
+            throw new RelayCommunicationError(url, error.message);
+        }
     }
 
-    public async __post(url, body) {
-        return httpie.post(url, { body, headers: this.headers, timeout: 2000 });
+    private async post(url, body) {
```
faustbrian: Should add a struct/interface for the expected return value, maybe add a `HttpieResponse` to `@arkecosystem/core-utils`.

alessiodf:
@supaiku0 

>(It may be noted that the memory leak as described in #2326 seems to be almost non-present with Node.js 11 despite the provided script saying otherwise)

I suggest you take a snapshot it with `heapdump` and check it because promise recursion leaks are definitely still present in Node.js v11:

Screenshot 2019-03-30 at 19 01 11
Screenshot 2019-03-30 at 19 01 15

supaiku0:
> I suggest you take a snapshot with heapdump and check it because promise recursion leaks are definitely still present in Node.js v11 and the script is accurate.

Yeah, you are right. The promise leak is still present in Node 11. But in the case of Core I can only reproduce the leak on Node 10. My wording was a bit misleading.

kristjank:
Referencing this: https://github.com/ArkEcosystem/core/issues/2326

PR #2428, opened 2019-04-16
fix(core-database): don't swallow errors in buildWallets

faustbrian:
## Proposed changes

Errors in `buildWallets` were swallowed which meant the state machine would never be able to catch any issues with it.

## Types of changes

## Checklist

### DIFF
--- a/packages/core-database/src/database-service.ts
+++ b/packages/core-database/src/database-service.ts
@@ -122,15 +122,10 @@ export class DatabaseService implements Database.IDatabaseService {
         }
     }
 
-    // @TODO: make this throw an error
     public async buildWallets(): Promise<void> {
         this.walletManager.reset();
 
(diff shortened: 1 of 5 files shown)


packages/core-database/src/database-service.ts
```diff
@@ -122,15 +122,10 @@ export class DatabaseService implements Database.IDatabaseService {
...
-    // @TODO: make this throw an error
     public async buildWallets(): Promise<void> {
         this.walletManager.reset();
 
-        try {
-            await this.connection.buildWallets();
-        } catch (err) {
-            this.logger.error(err.stack);
-        }
+        await this.connection.buildWallets();
```
supaiku0:
```suggestion
        return this.connection.buildWallets();
```

PR #2430, opened 2019-04-16
refactor: block broadcast when block comes from forger

air1one:
## Proposed changes

When block is received from forger, always broadcast to all peers.

Also refactored the block broadcast general logic, to use "peer quotas" based on block pings, instead of using randomization to select peers to broadcast to.

## Types of changes

## Checklist

### DIFF
--- a/packages/core-p2p/src/network-monitor.ts
+++ b/packages/core-p2p/src/network-monitor.ts
@@ -310,9 +310,9 @@ export class NetworkMonitor implements P2P.INetworkMonitor {
             // wait a bit before broadcasting if a bit early
             const diff = blockPing.last - blockPing.first;
             const maxHop = 4;
-            let proba = (maxHop - blockPing.count) / maxHop;
+            let broadcastQuota = (maxHop - blockPing.count) / maxHop;
 
-            if (diff < 500 && proba > 0) {
+            if (diff < 500 && broadcastQuota > 0) {
                 await delay(500 - diff);
 
                 blockPing = blockchain.getBlockPing();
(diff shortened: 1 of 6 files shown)


packages/core-p2p/src/socket-server/versions/peer.ts
```diff
@@ -63,8 +63,9 @@ export async function postBlock({ req }): Promise<void> {
     const blockchain: Blockchain.IBlockchain = app.resolvePlugin<Blockchain.IBlockchain>("blockchain");
 
     const block: Interfaces.IBlockData = req.data.block;
+    const fromForger = isWhitelisted(app.resolveOptions("p2p").remoteAccess, req.headers.remoteAddress);
```
faustbrian:
```suggestion
    const fromForger: boolean = isWhitelisted(app.resolveOptions("p2p").remoteAccess, req.headers.remoteAddress);
```

packages/core-p2p/src/network-monitor.ts
```diff
@@ -310,9 +310,9 @@ export class NetworkMonitor implements P2P.INetworkMonitor {
             // wait a bit before broadcasting if a bit early
             const diff = blockPing.last - blockPing.first;
             const maxHop = 4;
-            let proba = (maxHop - blockPing.count) / maxHop;
+            let broadcastQuota = (maxHop - blockPing.count) / maxHop;
```
faustbrian:
```suggestion
            let broadcastQuota: number = (maxHop - blockPing.count) / maxHop;
```

PR #2431, opened 2019-04-16
fix(core-database): wrong delegate order after start up

### LINKED ISSUE #2350: Delegate ranks in wallet manager are all over the place until the first block is received
**Describe the bug**
When a node is restarted the delegate ranks are all over the place until the first block is received which triggers the recalculation of ranks.

**Note that this does not affect the forging list as the rank property is just returned via API and not used internally.**

**To Reproduce**
The easiest way to visualize this is to run dexplorer locally, use your devnet node as host and keep the delegate monitor open while you restart your node. Shortly after the public API is mounted and new data is polled you will see a delegate list that looks very odd.

This changes a few seconds later as soon as the first block comes in.

**Expected behavior**
Return the correct delegate list right from the start.

supaiku0:
## Proposed changes

Resolves #2350 for real this time.

## Types of changes

## Checklist

### DIFF
--- a/__tests__/unit/core-database/repositories/delegates-business-repository.test.ts
+++ b/__tests__/unit/core-database/repositories/delegates-business-repository.test.ts
@@ -51,6 +51,8 @@ describe("Delegate Repository", () => {
         it("should return the local wallets of the connection that are delegates", () => {
             // @ts-ignore
             jest.spyOn(walletManager, "allByAddress").mockReturnValue(wallets);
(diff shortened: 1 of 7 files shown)


alessiodf:
@supaiku0 This isn't fixed as of 2.4.0-next.5. The order is totally wrong until a new round starts.

supaiku0:
Yeah, I'm able to reproduce it. Thanks!

Issue #1060, opened 2018-10-04
Transaction expiration calculation -> change times to only block time


kristjank:
**Is your feature request related to a problem? Please describe.**
Now it is possible to set transaction expiration in two ways (block-time and also normal time-to-live before expiration).

**Describe the solution you'd like**
The only true and measurable time is block height and so in this case also transaction expiration should be set, in explicit value of block height when it should expires.transaction: {

 expiration: 5980000
}

For usability the client applications would still calculate this in a time-spam, so we can show more friendly information to the user sending the transaction. The clients can also suggest like 10 blocks, 2 rounds and then the explicit block expiration height is calculated in the background.

kristjank:
@vasild did you manage to check into this

vasild:
No, but I have a rough idea:
1. Rename `transaction.expiration` -> `transaction.expirationSecondsSinceGenesis`
2. Introduce new `transaction.expirationHeight`
If 2. is set then override/ignore 1. and expire based on blocks, if 2. is not set then do as the current code does (expire by seconds).
3. Get the user-facing apps to set 2.
4. Delete 1. from the code as it will be unused after 3.

IMO the user-facing apps should present the user time-based expiration dialogs, like "expire after N hours _approximately_" or "expire at 14:56 on Nov 1 2018 _approximately_" and convert into block number under the hood. Then maybe have some advanced, geeks-only option to set/view expiration block number directly.

vasild:
Implemented in https://github.com/ArkEcosystem/core/pull/2461

PR #2450, opened 2019-04-19
feat(core-transaction-pool): Switch expiration to chain height

vasild:
Switch from calculating transactions' expiration in "number of seconds
since genesis block" to "chain height".

Notice: this change alters the semantic of the maxTransactionAge beyond
just the units (seconds vs height). Before the change we would use the
user-defined transaction.timestamp and add maxTransactionAge (seconds)
to it. After the change we use the current chain height when we have
received the transaction (which is not user-defined and not necessary
the same as when the transaction was created by the user) and add
maxTransactionAge (height) to it.

Resolves https://github.com/ArkEcosystem/core/issues/1060

## Types of changes

## Checklist


faustbrian:
Point this towards the `2.4` branch.

packages/core-transaction-pool/src/memory.ts
```diff
@@ -138,7 +145,13 @@ export class Memory {
             this.byType[type].add(MemoryTransaction);
         }
 
-        if (MemoryTransaction.expiresAt(maxTransactionAge) !== null) {
+        if (type !== Enums.TransactionTypes.TimelockTransfer) {
+            const maxHeight: number = Crypto.slots.getHeight() + maxTransactionAge;
```
supaiku0: I don't think this works, because we never update the slots height so it will always return 1. We have two options, either update the slots height when we receive a new block (see https://github.com/ArkEcosystem/core/blob/2.4/packages/core-blockchain/src/state-storage.ts#L99) or query the blockchain state (`app.resolve("state").getLastBlock().data.height`).

PR #2461, opened 2019-04-23
feat(core-transaction-pool): Switch expiration to chain height

### LINKED ISSUE #1060: Transaction expiration calculation -> change times to only block time

vasild:
Switch from calculating transactions' expiration in "number of seconds
since genesis block" to "chain height".

Notice: this change alters the semantic of the maxTransactionAge beyond
just the units (seconds vs height). Before the change we would use the
user-defined transaction.timestamp and add maxTransactionAge (seconds)
to it. After the change we use the current chain height when we have
received the transaction (which is not user-defined and not necessary
the same as when the transaction was created by the user) and add
maxTransactionAge (height) to it.

Resolves #1060

- [ ] Bugfix
- [x] New feature
- [ ] Build-related changes
- [ ] Documentation
- [ ] Other, please describe:

- [x] Yes
- [ ] No

## Checklist


packages/core-transaction-pool/src/defaults.ts
```diff
@@ -11,7 +11,10 @@ export const defaults = {
     allowedSenders: [],
     maxTransactionsPerRequest: process.env.CORE_TRANSACTION_POOL_MAX_PER_REQUEST || 40,
     maxTransactionBytes: process.env.CORE_TRANSACTION_POOL_MAX_TRANSACTIONS_SIZE || 1047876,
-    maxTransactionAge: 21600,
+    // Max transaction age in number of blocks produced since receiving a transaction.
+    // If a transaction stays that long in the pool without being included in any block,
+    // then it will be removed.
+    maxTransactionAge: 2700,
```
faustbrian: Rename this to something better so the comment is not needed to understand what it does.

packages/core-transaction-pool/src/memory-transaction.ts
```diff
@@ -1,8 +1,13 @@
...
-import { Enums, Interfaces } from "@arkecosystem/crypto";
+import { Interfaces } from "@arkecosystem/crypto";
 import assert from "assert";
 
+/**
+ * A mem pool transaction.
+ * A normal transaction
+ * + a sequence number used to order by insertion time
+ */
 export class MemoryTransaction {
```
faustbrian: Rename this to something like `SequentialTransaction` so the comment is not needed to understand what it does, the comment is already superfluous with how simple the code is.

packages/core-transaction-pool/src/memory.ts
```diff
@@ -18,6 +18,13 @@ export class Memory {
...
     private bySender: { [key: string]: Set<MemoryTransaction> } = {};
     private byType: { [key: number]: Set<MemoryTransaction> } = {};
+    /**
+     * An array of transactions, sorted by expiration (lower height comes first).
+     * This array may not contain all transactions that are in the pool,
+     * transactions that are without expiration are not included. Used to:
+     * - find all transactions that have expired (have an expiration height
+     *   lower than the current height) - they are at the beginning of the array.
+     */
     private byExpiration: MemoryTransaction[] = [];
```
faustbrian: Why do you need that long of a comment to understand what an array with the name `byExpiration` holds? The name is pretty self-explanatory and the code that uses it is also very simple.

If you feel the need to write that long comments, or even comments at all for such trivial things, you should think about why you really need those comments and remove the need for them and figure out how to make the code easier to understand.

__tests__/unit/core-transaction-pool/connection.test.ts
```diff
@@ -241,9 +241,15 @@ describe("Connection", () => {
             expect(notAdded).toBeEmpty();
 
             expect(connection.getPoolSize()).toBe(4);
-            await delay((expireAfterSeconds + 1) * 1000);
+
+            const origHeight = slots.getHeight();
```
faustbrian:
```suggestion
            const origHeight: number = slots.getHeight();
```

__tests__/unit/core-transaction-pool/connection.test.ts
```diff
@@ -217,8 +217,8 @@ describe("Connection", () => {
         it("should add the transactions to the pool and they should expire", async () => {
             expect(connection.getPoolSize()).toBe(0);
 
-            const expireAfterSeconds = 3;
-            const expiration = slots.getTime() + expireAfterSeconds;
+            const expireAfterBlocks = 3;
```
faustbrian:
```suggestion
            const expireAfterBlocks: number = 3;
```

__tests__/unit/core-transaction-pool/connection.test.ts
```diff
@@ -217,8 +217,8 @@ describe("Connection", () => {
...
+            const expiration = slots.getHeight() + expireAfterBlocks;
```
faustbrian:
```suggestion
            const expiration: number = slots.getHeight() + expireAfterBlocks;
```

packages/core-transaction-pool/src/memory.ts
```diff
@@ -48,17 +56,17 @@ export class Memory {
...
-    public getExpired(maxTransactionAge: number): Interfaces.ITransaction[] {
+    public getExpired(): Interfaces.ITransaction[] {
         if (!this.byExpirationIsSorted) {
-            this.byExpiration.sort((a, b) => a.expiresAt(maxTransactionAge) - b.expiresAt(maxTransactionAge));
+            this.byExpiration.sort((a, b) => a.transaction.data.expiration - b.transaction.data.expiration);
             this.byExpirationIsSorted = true;
         }
 
-        const now: number = Crypto.slots.getTime();
+        const currentHeight: number = app.resolve("state").getLastBlock().data.height;
```
faustbrian:
```suggestion
        const currentHeight: number = app.resolve("blockchain").getLastHeight();
```

packages/core-transaction-pool/src/defaults.ts
vasild: I think the variable name is fine, any suggestions are welcome though.

It is not possible to condense the information from 3 lines into a variable name. In this case the important information the comment is conveying is:
* The units of measurement (it is blocks, not seconds)
* When does the counter start ticking (it is when the transaction has been received, not when it was created like it was before this change)
* What happens when the time expires.

Please, do not remove the comment.

It is normal to have one or two sentences describing each variable/config parameter, for example:
```
  -debuglogfile=<file>
       Specify location of debug log file. Relative paths will be prefixed by a
       net-specific datadir location. (-nodebuglogfile to disable;
       default: debug.log)

  -includeconf=<file>
       Specify additional configuration file, relative to the -datadir path
       (only useable from configuration file, not command line)

  -loadblock=<file>
       Imports blocks from external blk000??.dat file on startup

  -maxmempool=<n>
       Keep the transaction memory pool below <n> megabytes (default: 300)

  -maxorphantx=<n>
       Keep at most <n> unconnectable transactions in memory (default: 100)

  -mempoolexpiry=<n>
       Do not keep transactions in the mempool longer than <n> hours (default:
       336)

  -par=<n>
       Set the number of script verification threads (-8 to 16, 0 = auto, <0 =
       leave that many cores free, default: 0)

  -persistmempool
       Whether to save the mempool on shutdown and load on restart (default: 1)

  -prune=<n>
       Reduce storage requirements by enabling pruning (deleting) of old
       blocks. This allows the pruneblockchain RPC to be called to
       delete specific blocks, and enables automatic pruning of old
       blocks if a target size in MiB is provided. This mode is
       incompatible with -txindex and -rescan. Warning: Reverting this
       setting requires re-downloading the entire blockchain. (default:
       0 = disable pruning blocks, 1 = allow manual pruning via RPC,
       >=550 = automatically prune block files to stay under the
       specified target size in MiB)
```

packages/core-transaction-pool/src/memory-transaction.ts
vasild: Renamed to `SequentialTransaction`, however the comment is still useful because it says what is it being used for, which can't be deduced from `SequentialTransaction`.

packages/core-transaction-pool/src/memory.ts
vasild: The comment is needed to convey the following information, which cannot be deduced from the name of the variable:
* The array does not contain all transactions
* Is it sorted in ascending or descending order

packages/core-transaction-pool/src/memory.ts
```diff
@@ -62,7 +62,7 @@ export class Memory {
             this.byExpirationIsSorted = true;
         }
 
-        const currentHeight: number = app.resolve("state").getLastBlock().data.height;
+        const currentHeight: number = app.resolve("blockchain").getLastHeight();
```
vasild: Why did you suggest this change? Was there something wrong with `app.resolve("state").getLastBlock().data.height;`?

Your suggestion fails with:
`TypeError: core_container_1.app.resolve(...).getLastHeight is not a function`, same if `app.resolvePlugin<Blockchain.IBlockchain>("blockchain").getLastHeight()` is used instead.

faustbrian: Just a shortcut for the same call but by using `getLastHeight` you won't be affected by changes to the underlying `state` storage as the return value will remain the same.

There is a small issue here with the startup of a node, you rely on bindings from `core-blockchain` but those won't be available until `core-blockchain` is mounted which happens a few plugins later. You would need a fallback to use the database on boot while `core-blockchain` bindings are not available.

packages/core-transaction-pool/src/memory.ts
```diff
@@ -48,21 +56,21 @@ export class Memory {
...
-    public getExpired(maxTransactionAge: number): Interfaces.ITransaction[] {
+    public getExpired(): Interfaces.ITransaction[] {
         if (!this.byExpirationIsSorted) {
-            this.byExpiration.sort((a, b) => a.expiresAt(maxTransactionAge) - b.expiresAt(maxTransactionAge));
+            this.byExpiration.sort((a, b) => a.transaction.data.expiration - b.transaction.data.expiration);
             this.byExpirationIsSorted = true;
         }
 
-        const now: number = Crypto.slots.getTime();
+        const currentHeight: number = app.resolve("blockchain").getLastHeight();
```
faustbrian:
```suggestion
        let currentHeight: number;
        if (app.has("blockchain")) {
            currentHeight = app.resolvePlugin<Blockchain.IBlockchain>("blockchain").getLastHeight();
        } else {
            const { data }: Interfaces.IBlock = await app
                .resolvePlugin<Database.IDatabaseService>("database")
                .getLastBlock();

            currentHeight = data.height;
        }
```

Keep in mind that you will need to import the interfaces via `import { Blockchain, Database } from "@arkecosystem/core-interfaces";` and this change also requires the method to become `async` as the database call requires `await`.

packages/core-transaction-pool/src/defaults.ts
faustbrian: How is it not possible to convey that minuscule bit of information in a variable name? All of the following make it pretty clear that it is about lifetime/expiration of the transaction and the unit of measurement is a number of blocks.

- `transactionTtlInBlocks`
- `transactionLifetimeInBlocks`
- `transactionValidForNumberOfBlocks`
- `transactionMaxAgeInBlocks`
- `transactionExpiresAfterNumberOfBlocks`

packages/core-transaction-pool/src/memory.ts
vasild: This causes an avalanche effect of turning all methods to `async`. That is not acceptable.

Instead I changed the code to try to query the current height from the state or blockchain in (an synchronous way), if they are initialized and if not - then handle "unknown current height".

packages/core-transaction-pool/src/memory.ts
```diff
@@ -256,4 +273,16 @@ export class Memory {
 
         return removed;
     }
+
+    private currentHeight(): number {
+        if (app.has("state")) {
+            return app.resolve("state").getLastBlock().data.height;
+        }
+
+        if (app.has("blockchain")) {
```
faustbrian: This will never be reached once `core-blockchain` is mounted because `if (app.has("state"))` should always be `true` at that point. You have to chose between `getLastHeight` or `getLastBlock`.

vasild: Even better! Removed.

packages/core-transaction-pool/src/memory.ts
```diff
@@ -274,7 +274,7 @@ export class Memory {
     }
 
     private currentHeight(): number {
-        if (app.has("state")) {
+        if (app && typeof app.has === "function" && app.has("state")) {
```
faustbrian: What's the purpose of this? If you are getting an error here during tests it means that a mock is missing and not that you should add more checks to get a test passing.

__tests__/unit/core-transaction-pool/connection.test.ts
```diff
@@ -217,8 +215,8 @@ describe("Connection", () => {
         it("should add the transactions to the pool and they should expire", async () => {
             expect(connection.getPoolSize()).toBe(0);
 
-            const expireAfterSeconds = 3;
-            const expiration = Slots.getTime() + expireAfterSeconds;
+            const expireAfterBlocks: number = 3;
+            const expiration: number = /* XXX how to get current height? */ 123 + expireAfterBlocks;
```
faustbrian: This is a unit test so you can just use a fixed value, no need to get a real height.

__tests__/unit/core-transaction-pool/connection.test.ts
```diff
@@ -241,16 +239,19 @@ describe("Connection", () => {
             expect(notAdded).toBeEmpty();
 
             expect(connection.getPoolSize()).toBe(4);
-            await delay((expireAfterSeconds + 1) * 1000);
+
+            /* XXX how to wait until the current height becomes `expiration` or fake it?
```
faustbrian: See https://github.com/ArkEcosystem/core/commit/f8c2acc04a8e9146286ded9c0bb0153d4b79d148 to get an idea how to apply the necessary mocks.

vasild: That worked with a few tweaks, thanks!

packages/core-transaction-pool/src/memory.ts
vasild: Removed

packages/core-transaction-pool/src/memory.ts
```diff
@@ -242,4 +273,12 @@ export class Memory {
 
         return removed;
     }
+
+    private currentHeight(): number {
```
faustbrian: After the introduction of core-state in https://github.com/ArkEcosystem/core/commit/a4567cd2696f2c9c6061c47dc0172dc31d994417 the last block is always present before the transaction pool mounts because `core-database-postgres` mounts before `core-transaction-pool` which will set the last block.

Instead of calling this method you can now do `app.resolvePlugin("state").getLastHeight()` to get the last height.

packages/core-transaction-pool/src/memory.ts
```diff
@@ -49,20 +58,31 @@ export class Memory {
     }
 
     public getExpired(maxTransactionAge: number): Interfaces.ITransaction[] {
+        const currentHeight: number = this.currentHeight();
+
+        if (!currentHeight) {
```
supaiku0: This check can be removed, there's now always a block and therefore height always truthy.

packages/core-transaction-pool/src/memory.ts
```diff
@@ -76,70 +96,82 @@ export class Memory {
...
             // Append to existing transaction ids for this type.
-            this.byType[type].add(MemoryTransaction);
+            this.byType[type].add(SequentialTransaction);
         }
 
-        if (MemoryTransaction.expiresAt(maxTransactionAge) !== null) {
-            this.byExpiration.push(MemoryTransaction);
+        if (type !== Enums.TransactionTypes.TimelockTransfer) {
+            const currentHeight: number = this.currentHeight();
+            if (currentHeight) {
```
supaiku0: This check can be removed, there's now always a block and therefore height always truthy.

packages/core-transaction-pool/src/memory.ts
vasild: Y

packages/core-transaction-pool/src/memory.ts
vasild: Y

packages/core-transaction-pool/src/memory.ts
vasild: Y

packages/core-transaction-pool/src/memory.ts
supaiku0: Imo the information the comment tries to convey is a bit redundant since it can be quickly deduced by checking the relevant code. I personally would remove the comment, too.

packages/core-transaction-pool/src/memory.ts
```diff
@@ -76,70 +91,77 @@ export class Memory {
...
+        if (SequentialTransactions !== undefined) {
+            return SequentialTransactions;
         }
 
         return new Set();
     }
 
-    public remember(MemoryTransaction: MemoryTransaction, maxTransactionAge: number, databaseReady?: boolean): void {
-        const transaction: Interfaces.ITransaction = MemoryTransaction.transaction;
+    public remember(SequentialTransaction: SequentialTransaction, maxTransactionAge: number, databaseReady?: boolean): void {
```
faustbrian:
```suggestion
    public remember(sequentialTransaction: SequentialTransaction, maxTransactionAge: number, databaseReady?: boolean): void {
```

Variable names should always be in `camelCase`.

vasild: You named it `MemoryTransaction` all over the place, I did not wanted to argue and assumed you wanted it to be with capital letter for some reason.

packages/core-transaction-pool/src/memory.ts
vasild: I changed this in a followup PR, see if you like it: https://github.com/ArkEcosystem/core/pull/2494/files#diff-ea1f8e8a5fdc70f328cde94d66d99dfeR20

packages/core-transaction-pool/src/memory.ts
vasild: I completely removed the `[sS]equentialTransaction` from the code in a followup PR: https://github.com/ArkEcosystem/core/pull/2494

PR #2471, opened 2019-04-24
fix(crypto): add exceptions for invalid recipients

### LINKED ISSUE #2470: Missing exceptions on Mainnet
**Describe the bug**
Cant setup a node from scratch as it finds bad blocks and transactions that need an exception. Mainly regarding second signature, problem persist after importing a snapshot via cli from other server.

`1|ark-relay  | [2019-04-24 12:16:16.035] ERROR: [PoolWalletManager] Can't apply transaction daa14884aa508f396484be52779e07d5fa1a80c69fd63253915c579fe9a2b8e5 from AKBUegoaAoSq3bepmXmzUL7XeEt6ScVM98 due to "Failed to apply transaction, because wallet does not allow second signatures."`

**To Reproduce**
setup new mainnet relay

dated:
## Summary

#2470 is about a different kind of error. This PR adds exceptions for transactions to the following addresses (which also stop a node from syncing from zero):

```
TobY8GsEuAkpDQFM9rdz6QadmMfsw3wnoS
14owCmVDn8SaAFZcLbZfCVu5jvc4Lq7Tm1
151CWqrw1mWNDQGeKCznF6ZSjKvivfwApp
1Mi5BKbXQzhEvdcCQXdTRthkPW7ryS1p8g
```

- [x] Bugfix

- [x] No

## Checklist

## Further Comments

The removed ID `4f83bc708288044a6f5f2773b90e1456af2ddc50ed8f3b63e960f5f16cac7d73` was listed twice.


supaiku0:
Thanks! Devnet also needs around 11 exceptions, care to add them? 
Also, I don't think that it resolves the linked issue, since the tx in question is not failing because of an invalid recipient.

dated:
I'm on mobile now so that's a bit difficult at the moment ;)

> Also, I don't think that it resolves the linked issue, since the tx in question is not failing because of an invalid recipient.

Yea, i adjusted the description since.

dated:
A quick db lookup revealed 5 transactions on devnet which appear to be handled already, I can have a deeper look later.

supaiku0:
Yep, nevermind. Devnet is already handled.

PR #2528, opened 2019-05-03
refactor(crypto): invalidate blocks with expired transactions

supaiku0:
## Summary

The expiration is now height based, but it's still possible to forge expired transactions when bypassing the pool. This PR prevents expired transactions from entering the pool and blocks which contain expired transactions are now invalid.

- [X] Bugfix
- [ ] New feature
- [ ] Build-related changes
- [ ] Documentation
- [ ] Other, please describe:

- [X] Yes
- [ ] No

- [ ] Yes
  - [ ] All tests are passing
- [ ] No

## Checklist


vasild:
Actually that should be part of block validation - if a block at height H contains a transaction with expiration <= H, then that block is invalid.

supaiku0:
> Actually that should be part of block validation - if a block at height H contains a transaction with expiration <= H, then that block is invalid.

Yeah, updated the PR. I went with the schema approach first, because then I wouldnt need to touch the tx pool/block validation at all, but it turns out that there are too many edge cases which require more complex changes to handle expired transactions correctly.

PR #2538, opened 2019-05-07
feat: Implement Delegate Resignation (AIP11)

vasild:


packages/core-jest-matchers/src/transactions/types/delegate.ts
```diff
@@ -17,7 +17,8 @@ expect.extend({
     toBeDelegateType: received => {
         return {
             message: () => "Expected value to be a valid DELEGATE transaction.",
-            pass: received.type === DelegateRegistration,
+            pass: received.type === TransactionTypes.DelegateRegistration ||
+                  received.type === TransactionTypes.DelegateResignation,
```
faustbrian: This matcher is supposed to check for a delegate registration, this will now cause false results for tests if you don't pay attention. All changes in this file need to be reverted.

There is already a matcher for a resignation at https://github.com/ArkEcosystem/core/blob/b58f7dcfca73a34158f7d76486bcf3e11dcfa596/packages/core-jest-matchers/src/transactions/types/delegate-resignation.ts.

packages/crypto/src/transactions/types/schemas.ts
```diff
@@ -215,9 +215,23 @@ export const multiPayment = extend(transactionBaseSchema, {
 
 export const delegateResignation = extend(transactionBaseSchema, {
     $id: "delegateResignation",
+    required: ["asset"],
     properties: {
         type: { transactionType: TransactionTypes.DelegateResignation },
         amount: { bignumber: { minimum: 0, maximum: 0 } },
+        asset: {
```
faustbrian: Why do you need an asset? The transaction will be applied to the sender which is required to be a wallet with a username so you can grab the username from it directly.

packages/core-state/src/wallets/wallet-manager.ts
```diff
@@ -129,19 +129,16 @@ export class WalletManager implements State.IWalletManager {
     }
 
     public loadActiveDelegateList(roundInfo: Shared.IRoundInfo): State.IDelegateWallet[] {
+        const delegates: State.IWallet[] = this.buildDelegateRanking(undefined, roundInfo);
```
supaiku0: Why this change? Line 325 should be enough.

vasild:
Alright, I added a `resigned` flag to mark delegates that have resigned instead of un-delegating them.

> The sender must be a delegate

The check is here: https://github.com/ArkEcosystem/core/pull/2538/files#diff-304db985d999842253be956ceec47a2aR20 `!wallet.username`

> Once forged the delegate is irreversibly deactivated

Will be handled by existing code: https://github.com/ArkEcosystem/core/blob/afb69f2141383/packages/core-transactions/src/handlers/delegate-registration.ts#L60

> no longer possible to vote for this delegate

Added a check here: https://github.com/ArkEcosystem/core/pull/2538/files#diff-ced27755786a5f642370021ba08a2853R66

> no longer treated as an active delegates, regardless of vote balance

Removed resigned delegates: https://github.com/ArkEcosystem/core/pull/2538/files#diff-097e9ec56428de146b54cea1f67d0448R325

packages/core-state/src/wallets/wallet.ts
```diff
@@ -164,6 +166,10 @@ export class Wallet implements State.IWallet {
             audit.push({ "New username": username });
         }
 
+        if (transaction.type === Enums.TransactionTypes.DelegateResignation) {
+            audit.push({ "Resigned delegate": this.username });
```
supaiku0: Don't need an asset.

packages/core-transactions/src/errors.ts
```diff
@@ -90,6 +90,15 @@ export class WalletUsernameAlreadyRegisteredError extends TransactionError {
     }
 }
 
+export class WalletUsernameMismatchError extends TransactionError {
```
supaiku0: The resignation can only be made by delegates. Therefore we don't need a username.

packages/core-transactions/src/handlers/delegate-resignation.ts
```diff
@@ -8,21 +9,59 @@ export class DelegateResignationTransactionHandler extends TransactionHandler {
     }
 
     public async bootstrap(connection: Database.IConnection, walletManager: State.IWalletManager): Promise<void> {
-        return;
+        // XXX should we do something here?
```
supaiku0: Yes, here we have to update the wallets and apply the `resignated`  status.
See the other handlers, but basically:
1) Get all resignation assets
2) Iterate over them and apply resignated flag to sender wallet

packages/core-transactions/src/handlers/delegate-resignation.ts
```diff
@@ -8,21 +9,59 @@ export class DelegateResignationTransactionHandler extends TransactionHandler {
...
     }
 
     public canBeApplied(
         transaction: Interfaces.ITransaction,
         wallet: State.IWallet,
         databaseWalletManager: State.IWalletManager,
     ): boolean {
+        if (!wallet.username || !transaction.data.asset.delegate.username) {
```
supaiku0: The transaction doesn't need an asset.

packages/core-transactions/src/handlers/delegate-resignation.ts
```diff
@@ -8,21 +9,59 @@ export class DelegateResignationTransactionHandler extends TransactionHandler {
...
+            throw new WalletUsernameEmptyError();
+        }
+
+        if (wallet.username !== transaction.data.asset.delegate.username) {
```
supaiku0: Can be removed.

packages/crypto/src/transactions/builders/transactions/delegate-resignation.ts
```diff
@@ -11,12 +11,20 @@ export class DelegateResignationBuilder extends TransactionBuilder<DelegateResig
         this.data.type = TransactionTypes.DelegateResignation;
         this.data.fee = feeManager.get(TransactionTypes.DelegateResignation);
         this.data.amount = BigNumber.ZERO;
-        this.data.asset = {};
+        this.data.recipientId = undefined;
+        this.data.senderPublicKey = undefined;
+        this.data.asset = { delegate: {} } as ITransactionAsset;
+    }
+
+    public usernameAsset(username: string): DelegateResignationBuilder {
```
supaiku0: Remove.

packages/crypto/src/transactions/builders/transactions/delegate-resignation.ts
```diff
@@ -11,12 +11,20 @@ export class DelegateResignationBuilder extends TransactionBuilder<DelegateResig
...
+        this.data.asset.delegate.username = username;
+        return this;
     }
 
     public getStruct(): ITransactionData {
         const struct: ITransactionData = super.getStruct();
         struct.amount = this.data.amount;
+        struct.recipientId = this.data.recipientId;
```
supaiku0: Remove.

packages/crypto/src/transactions/serializer.ts
```diff
@@ -69,7 +69,8 @@ export class Serializer {
                 break;
             }
 
-            case TransactionTypes.DelegateRegistration: {
+            case TransactionTypes.DelegateRegistration:
+            case TransactionTypes.DelegateResignation: {
```
supaiku0: Revert. This is legacy V1 serialization.

packages/crypto/src/transactions/types/delegate-resignation.ts
```diff
@@ -12,10 +12,22 @@ export class DelegateResignationTransaction extends Transaction {
     }
 
     public serialize(options?: ISerializeOptions): ByteBuffer {
-        return new ByteBuffer(0);
```
supaiku0: Revert.

packages/crypto/src/transactions/types/delegate-resignation.ts
```diff
@@ -12,10 +12,22 @@ export class DelegateResignationTransaction extends Transaction {
...
+        const buffer: ByteBuffer = new ByteBuffer(delegateBytes.length, true);
+
+        buffer.writeByte(delegateBytes.length);
+        buffer.append(delegateBytes, "hex");
+
+        return buffer;
     }
 
     public deserialize(buf: ByteBuffer): void {
-        return;
```
supaiku0: Revert.

packages/core-state/src/wallets/wallet-manager.ts
vasild: This change is because we want to check against `maxDelegates` with resigned delegates removed. Thus need to call `buildDelegateRanking()` before the check because `allByUsername()` would not filter anything.

Also this change makes the code simpler because it eliminates the call to `allByUsername()` and removes the variable `delegatesWallets`, so one call and one variable less.

Btw what a wretched naming for `maxDelegates`! We actually want the number of delegates to be bigger than that and emit an error if it is *less*. `minDelegates` would be more appropriate, but out of the scope of this PR.

packages/core-transactions/src/handlers/delegate-resignation.ts
```diff
@@ -8,21 +9,58 @@ export class DelegateResignationTransactionHandler extends TransactionHandler {
...
+            walletManager.findByPublicKey(transaction.senderPublicKey).resigned = true;
+        }
     }
 
     public canBeApplied(
         transaction: Interfaces.ITransaction,
         wallet: State.IWallet,
         databaseWalletManager: State.IWalletManager,
     ): boolean {
+        if (!wallet.username) {
```
supaiku0: Also throw if the sender already `resigned`.

packages/core-transactions/src/handlers/delegate-resignation.ts
```diff
@@ -8,21 +9,58 @@ export class DelegateResignationTransactionHandler extends TransactionHandler {
...
+    public canEnterTransactionPool(
+        data: Interfaces.ITransactionData,
+        pool: TransactionPool.IConnection,
+        processor: TransactionPool.IProcessor,
+    ): boolean {
+        if (this.typeFromSenderAlreadyInPool(data, pool, processor)) {
+            processor.pushError(
+                data,
+                "ERR_PENDING",
+                `Delegate resignation for "${data.asset.delegate.username}" already in the pool`,
```
supaiku0: There's no asset. You need to get the username from the sender wallet.

__tests__/functional/transaction-forging/delegate-resignation.test.ts
```diff
@@ -0,0 +1,238 @@
...
+            await expect(initialFunds).toBeAccepted();
+            await support.snoozeForBlock(1);
+            await expect(initialFunds.id).toBeForged();
+
+            // Resign a delegate
+            const transactionsResign = TransactionFactory.delegateResignation()
+                .withPassphrase(passphrase)
+                .createOne();
+
+            await expect(transactionsResign.id).toBeRejected();
```
faustbrian:
```suggestion
            await expect(transactionsResign).toBeRejected();
```

__tests__/functional/transaction-forging/delegate-resignation.test.ts
```diff
@@ -0,0 +1,238 @@
...
+            await expect(transactionsResign1).toBeAccepted();
+            await support.snoozeForBlock(1);
+            await expect(transactionsResign1.id).toBeForged();
+
+            // Resign a delegate
+            const transactionsResign2 = TransactionFactory.delegateResignation()
+                .withPassphrase(passphrase)
+                .createOne();
+
+            await expect(transactionsResign2.id).toBeRejected();
```
faustbrian:
```suggestion
            await expect(transactionsResign2).toBeRejected();
```

__tests__/functional/transaction-forging/delegate-resignation.test.ts
```diff
@@ -0,0 +1,238 @@
...
+            const passphrase = generateMnemonic();
+
+            // Initial Funds
+            const initialFunds = TransactionFactory.transfer(Identities.Address.fromPassphrase(passphrase), 100 * 1e8)
+                .withPassphrase(genesisPassphrase)
+                .createOne();
+
+            await expect(initialFunds).toBeAccepted();
+            await support.snoozeForBlock(1);
+            await expect(initialFunds.id).toBeAccepted();
```
faustbrian:
```suggestion
            await expect(initialFunds.id).toBeForged();
```

PR #2557, opened 2019-05-09
feat(core): add db configuration command

dated:
## Summary

Adds a new config command to the cli so that the database can be easily configured.

```
ark config:database --host=localhost --port=5432 --username=ark --database=ark_mainnet --password=password
```

Configuring the database credentials is possible already through the cli by setting the respective environment variables. This however requires knowledge over these variable names.

- [ ] Bugfix
- [x] New feature
- [ ] Build-related changes
- [ ] Documentation
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [ ] Yes
  - [ ] All tests are passing
- [x] No

## Checklist

## Other information

The CLI does not interact with the database in any way; changing the database name for instance will not rename the postgres database itself.


packages/core/src/commands/config/db.ts
```diff
@@ -0,0 +1,71 @@
...
+`,
+        `Set the database name
+$ ark config:db --database=ark_mainnet
+`,
+        `Set the database password
+$ ark config:db --password=password
+`,
+    ];
+
+    private static readonly validFlags: string[] = ["host", "username", "database", "password"];
```
faustbrian: Any reason you didn't include a `port` flag?

dated: I went by the vars laid out in https://docs.ark.io/guidebook/core/configuration.html#environment-configuration, i'll check up on the port envvar.

dated: I'll add it here and to the docs asap.

packages/core/src/commands/config/db.ts
```diff
@@ -0,0 +1,77 @@
...
+`,
+        `Set the database name
+$ ark config:db --database=ark_mainnet
+`,
+        `Set the database password
+$ ark config:db --password=password
+`,
+    ];
+
+    private static readonly validFlags: string[] = ["host", "port", "username", "database", "password"];
```
supaiku0: `database` should be `name` or vice versa?

packages/core/src/commands/config/db.ts
```diff
@@ -0,0 +1,77 @@
...
+        host: flags.string({
+            description: "the host of the database"
+        }),
+        port: flags.integer({
+            description: "the port of the database"
+        }),
+        username: flags.string({
+            description: "the name of the database user",
+        }),
+        name: flags.string({
```
supaiku0:
```suggestion
        database: flags.string({
```

Issue #2086, opened 2019-02-12
feat: nonce Introduction


kristjank:
Introduce NONCE into the transaction payload, to make replay attack harder.

Requirements:

- [X] NONCE size 8-bytes
- [X] Tx validation = Wallet.nonce + 1
- [X] Increase transaction version to 2 (hard fork)
- [X] Remove timestamp from tx-version 2
- [x] Order block content by sequence (as before), but make sure that this order does not violate nonce order for transactions from the same sender
- [x] Add to block validation order by nonce on protocol level
- [X] Add nonce column to `transactions` table + migration
- [X] Keep a transition period to move from tx-version-1 to 2
- [X] Update transaction pool logic to use nonce, instead of own sequence and timestamp
- [X] Updating of clients (to support both)
- [X] Light Client should maintain wallets.nonce state

PR #2573, opened 2019-05-13
feat: initial nonce implementation

supaiku0:
## Summary

Initial implementation of #2086. 

- [ ] Bugfix
- [X] New feature
- [X] Refactoring / Performance Improvements
- [ ] Build-related changes
- [ ] Documentation
- [ ] Other, please describe:

- [ ] Yes
- [ ] No

- [ ] Yes
  - [ ] All tests are passing
- [ ] No

## Checklist


packages/core-state/src/wallets/wallet.ts
```diff
@@ -82,6 +78,14 @@ export class Wallet implements State.IWallet {
         return false;
     }
 
+    public incrementNonce(): void {
+        this.nonce = this.nonce.plus(1);
+    }
+
+    public decrementNonce(): void {
+        this.nonce = this.nonce.minus(1);
```
vasild: Consider an assert that it does not go negative here.

packages/core-transaction-pool/src/processor.ts
```diff
@@ -224,7 +227,15 @@ export class Processor implements TransactionPool.IProcessor {
     }
 
     private addTransactionsToPool(): void {
-        const { notAdded }: ITransactionsProcessed = this.pool.addTransactions(Array.from(this.accept.values()));
+        const acceptedTransactions = Array.from(this.accept.values());
+        // We need to decrement the nonce of the sender, since canBeApplied
+        // is called on all transactions before they are actually applied
+        // and the nonce lags behind.
+        for (const accepted of acceptedTransactions) {
+            this.walletManager.decrementNonce(accepted.data.senderPublicKey);
```
vasild: This looks very bad - a hack that makes the code fragile in such a way that some future change could messup the nonce increment/decrements.

Is it not possible to have nonce increment in just one place when applying a transaction and decrement just in one place - when/if reverting a transaction?

packages/core-transaction-pool/src/wallet-manager.ts
```diff
@@ -53,6 +61,10 @@ export class WalletManager extends Wallets.WalletManager {
                     sender,
                     this.databaseService.walletManager,
                 );
+                // We need to decrement the nonce of the sender, since canBeApplied
+                // is called on all transactions before they are actually applied
+                // and the nonce lags behind.
+                sender.incrementNonce();
```
vasild: ditto

packages/core-transactions/src/handlers/transaction.ts
```diff
@@ -147,19 +154,25 @@ export abstract class TransactionHandler implements ITransactionHandler {
         return false;
     }
 
-    public applyToSenderInPool(transaction: any, poolWalletManager: State.IWalletManager): void {
+    public applyToSenderInPool(transaction: Interfaces.ITransaction, poolWalletManager: State.IWalletManager): void {
+        // TOOD: this is working around the fact that the tx pool needs to increment the nonce
+        // before applying it, but it's not exactly the best solution.
+        poolWalletManager.findByPublicKey(transaction.data.senderPublicKey).decrementNonce();
```
vasild: ditto

packages/core-transaction-pool/src/processor.ts
supaiku0: We already do increment/decrement when applying/reverting a transaction.

But without reworking a big part of the transaction pool processor, we need to increment/decrement nonces at this stage, too.

It's necessary because of how the pool works. We receive a payload of transactions and first perform some checks and call canBeApplied on each of them. At this stage the nonce does not change, but this violates the condition inside canBeApplied (that is if we receive multiple tx of the same sender in the same payload ).

Here's a super simplified version of how it works atm:
```
const acceptedTransactions = [];
for (const transaction of incomingTransactions) {
        ...
        a lot of checks
        canBeApplied(transaction)
        some more checks
        acceptedTransactions.push(transaction)
}
...
and more checks
...
addTransactionsToPool(acceptedTransactions) // wallet.apply is finally called here
```

How it needs to be to not need the hack:
```
for (const transaction of incomingTransactions) {
        ...
        some checks
        canBeApplied(transaction)
        more checks
        addTransactionToPool(transaction) // wallet.apply with each iteration
}
```

This way the nonce increases (if accepted) properly between each iteration and multiple transactions of the same sender don't cause issues anymore.

It's a bit out of scope for this PR.

vasild: Ok, I agree - this is out of scope.

Issue #2579, opened 2019-05-15
(core-transaction-pool): refactor validation of incoming transactions


supaiku0:
Currently, the transaction pool validates all transactions first before adding them. With the introduction of nonces this becomes a problem when the payload contains multiple transactions from the same sender. 

This is because `wallet.canBeApplied` checks if the nonce of the transaction matches, but it breaks as soon there's one more transaction from the same sender since the nonce of the wallet is only increased, when a transaction is actually applied.

Mega simplified view of how it is:
```javascript
const accepted = []
for (const transaction of transactionsInPayload) {
... // checks
    if (wallet.canBeApplied(transaction)) {
        accepted.push(transaction; // wallet nonce is not incremented between transactions
    }
... // checks
}
... // checks
addTransactionsToPool(accepted); // wallet.apply is called
```

Mega simplified view of a possible solution:
```javascript
for (const transaction of transactionsInPayload) {
... // checks
    if (wallet.canBeApplied(transaction)) {
        addTransactionToPool(transaction); // wallet nonce gets incremented between transactions
    }
...
}
```

This allows us to get rid of the workaround which manually deals with incrementing and decrementing the nonces to make it somehow work.

PR #2587, opened 2019-05-17
refactor: do not temporary inc/dec nonce when checking tx validity

vasild:
Instead relax the check in canBeApplied() to just requiring that the tx
nonce be larger than the wallet nonce. Do the strict check when applying
the transaction: tx nonce == wallet nonce + 1.

* Change the return type of canBeApplied() from boolean to void. It would
  either throw an exception on error or return true. No caller is
  checking the return value.

* Call canBeApplied() when applying a transaction in applyToSender(),
  this way the code is more robust because it eliminates the possibility
  for somebody to forget canBeApplied() before calling applyToSender().
  Also callers get simplified.

* Remove now unneeded methods incrementNonce() and decrementNonce().

* Simplify and rename throwIfApplyingFails().

* Remove the method applyToRecipientInPool() because it only calls
  applyToRecipient(). Call applyToRecipient() directly from the callers
  instead.

* Remove the applyToRecipientInPool() for the same reasons as
  applyToRecipientInPool().


vasild:
Resolves https://github.com/ArkEcosystem/core/issues/2579

packages/core-database/src/database-service.ts
```diff
@@ -568,7 +568,8 @@ export class DatabaseService implements Database.IDatabaseService {
         const dbTransaction = await this.getTransaction(transaction.data.id);
 
         try {
-            return transactionHandler.canBeApplied(transaction, sender, this.walletManager) && !dbTransaction;
+            transactionHandler.canBeApplied(transaction, sender, this.walletManager);
```
supaiku0: Please revert, this just looks off.

packages/core-transaction-pool/src/connection.ts
```diff
@@ -274,29 +274,26 @@ export class Connection implements TransactionPool.IConnection {
                 : undefined;
 
             if (recipientWallet) {
-                transactionHandler.applyToRecipientInPool(transaction, this.walletManager);
+                transactionHandler.applyToRecipient(transaction, this.walletManager);
```
supaiku0: Revert. They are purposely named that way because they are only intended to be called by the pool (they get passed the pool wallet manager). Now it's way more ambiguous.

packages/core-transaction-pool/src/connection.ts
```diff
@@ -274,29 +274,26 @@ export class Connection implements TransactionPool.IConnection {
...
             }
 
             if (exists) {
                 this.removeTransaction(transaction);
             } else if (senderWallet) {
-                // TODO: rework error handling
                 try {
                     transactionHandler.canBeApplied(transaction, senderWallet, this.databaseService.walletManager);
+                    transactionHandler.applyToSender(transaction, this.walletManager);
```
supaiku0: Same.

packages/core-transaction-pool/src/connection.ts
```diff
@@ -342,7 +339,7 @@ export class Connection implements TransactionPool.IConnection {
             try {
                 const transactionHandler: Handlers.TransactionHandler = Handlers.Registry.get(transaction.type);
                 transactionHandler.canBeApplied(transaction, senderWallet, this.databaseService.walletManager);
-                transactionHandler.applyToSenderInPool(transaction, this.walletManager);
+                transactionHandler.applyToSender(transaction, this.walletManager);
```
supaiku0: Same.

packages/core-transaction-pool/src/connection.ts
```diff
@@ -438,8 +435,8 @@ export class Connection implements TransactionPool.IConnection {
         this.memory.remember(transaction, this.options.maxTransactionAge);
 
         try {
-            this.walletManager.throwIfApplyingFails(transaction);
-            Handlers.Registry.get(transaction.type).applyToSenderInPool(transaction, this.walletManager);
+            this.walletManager.senderIsKnownAndTrxCanBeApplied(transaction);
+            Handlers.Registry.get(transaction.type).applyToSender(transaction, this.walletManager);
```
supaiku0: Same.

packages/core-transaction-pool/src/wallet-manager.ts
```diff
@@ -24,60 +24,28 @@ export class WalletManager extends Wallets.WalletManager {
...
-    public decrementNonce(publicKey: string): void {
-        this.findByPublicKey(publicKey).decrementNonce();
-    }
-
-    public throwIfApplyingFails(transaction: Interfaces.ITransaction): void {
+    public senderIsKnownAndTrxCanBeApplied(transaction: Interfaces.ITransaction): void {
         // Edge case if sender is unknown and has no balance.
         // NOTE: Check is performed against the database wallet manager.
-        const { senderPublicKey } = transaction.data;
+        const senderPublicKey: string = transaction.data.senderPublicKey;
```
supaiku0: Why remove the destructure?

packages/core-transaction-pool/src/wallet-manager.ts
```diff
@@ -24,60 +24,28 @@ export class WalletManager extends Wallets.WalletManager {
...
 
-                throw new Error(JSON.stringify([error.message]));
-            }
-        }
+        Handlers.Registry.get(transaction.type).canBeApplied(transaction, sender, this.databaseService.walletManager);
     }
 
     public revertTransactionForSender(transaction: Interfaces.ITransaction): void {
-        Handlers.Registry.get(transaction.type).revertForSenderInPool(transaction, this);
+        Handlers.Registry.get(transaction.type).revertForSender(transaction, this);
```
supaiku0: Same.

packages/core-transactions/src/handlers/delegate-registration.ts
```diff
@@ -45,7 +45,7 @@ export class DelegateRegistrationTransactionHandler extends TransactionHandler {
         transaction: Interfaces.ITransaction,
         wallet: State.IWallet,
         databaseWalletManager: State.IWalletManager,
-    ): boolean {
+    ): void {
```
supaiku0: Please keep the return type. You have to add a explicit return either way, else tslint is giving you a hard time.

packages/core-transactions/src/handlers/delegate-registration.ts
```diff
@@ -110,7 +110,7 @@ export class DelegateRegistrationTransactionHandler extends TransactionHandler {
         return true;
     }
 
-    protected applyToSender(transaction: Interfaces.ITransaction, walletManager: State.IWalletManager): void {
+    public applyToSender(transaction: Interfaces.ITransaction, walletManager: State.IWalletManager): void {
```
supaiku0: Don't make those public. Just keep the ones you removed.

packages/core-transactions/src/handlers/transaction.ts
```diff
@@ -37,39 +37,37 @@ export abstract class TransactionHandler implements ITransactionHandler {
...
     public canBeApplied(
         transaction: Interfaces.ITransaction,
-        wallet: State.IWallet,
-        databaseWalletManager: State.IWalletManager,
-    ): boolean {
-        const { data }: Interfaces.ITransaction = transaction;
+        sender: State.IWallet,
+        databaseWalletManager?: State.IWalletManager,
+    ): void {
+        const data: Interfaces.ITransactionData = transaction.data;
```
supaiku0: Keep the destructure.

packages/core-transactions/src/handlers/transaction.ts
```diff
@@ -37,39 +37,37 @@ export abstract class TransactionHandler implements ITransactionHandler {
 
     public canBeApplied(
         transaction: Interfaces.ITransaction,
-        wallet: State.IWallet,
-        databaseWalletManager: State.IWalletManager,
-    ): boolean {
-        const { data }: Interfaces.ITransaction = transaction;
+        sender: State.IWallet,
+        databaseWalletManager?: State.IWalletManager,
```
supaiku0: Why did you make it optional?

packages/core-transactions/src/handlers/transaction.ts
```diff
@@ -37,39 +37,37 @@ export abstract class TransactionHandler implements ITransactionHandler {
...
-        if (wallet.secondPublicKey) {
+        if (sender.secondPublicKey) {
             // Ensure the database wallet already has a 2nd signature, in case we checked a pool wallet.
-            const databaseWallet: State.IWallet = databaseWalletManager.findByPublicKey(
-                transaction.data.senderPublicKey,
-            );
-
-            if (!databaseWallet.secondPublicKey) {
-                throw new UnexpectedSecondSignatureError();
+            if (databaseWalletManager) {
```
supaiku0: There should always be a wallet manager.

packages/core-transactions/src/handlers/transaction.ts
```diff
@@ -81,23 +79,21 @@ export abstract class TransactionHandler implements ITransactionHandler {
...
-            if (!wallet.verifySignatures(data, wallet.multisignature)) {
+
+            if (!sender.verifySignatures(data, sender.multisignature)) {
                 throw new InvalidMultiSignatureError();
             }
         } else if (transaction.type !== Enums.TransactionTypes.MultiSignature && transaction.data.signatures) {
             throw new UnexpectedMultiSignatureError();
         }
-
-        return true;
```
supaiku0: Keep the return.

packages/core-transactions/src/handlers/transaction.ts
```diff
@@ -110,26 +106,50 @@ export abstract class TransactionHandler implements ITransactionHandler {
         this.revertForRecipient(transaction, walletManager);
     }
 
-    protected applyToSender(transaction: Interfaces.ITransaction, walletManager: State.IWalletManager): void {
+    public applyToSender(transaction: Interfaces.ITransaction, walletManager: State.IWalletManager): void {
```
supaiku0: Protected.

packages/core-transactions/src/handlers/transaction.ts
```diff
@@ -110,26 +106,50 @@ export abstract class TransactionHandler implements ITransactionHandler {
...
         const sender: State.IWallet = walletManager.findByPublicKey(transaction.data.senderPublicKey);
-        sender.balance = sender.balance.minus(transaction.data.amount).minus(transaction.data.fee);
-        sender.incrementNonce();
+        const d: Interfaces.ITransactionData = transaction.data;
```
supaiku0:
```suggestion
        const { data } = transaction;
```

packages/core-transactions/src/handlers/transaction.ts
```diff
@@ -110,26 +106,50 @@ export abstract class TransactionHandler implements ITransactionHandler {
...
+
+        sender.balance = newBalance;
     }
 
-    protected revertForSender(transaction: Interfaces.ITransaction, walletManager: State.IWalletManager): void {
+    public revertForSender(transaction: Interfaces.ITransaction, walletManager: State.IWalletManager): void {
         const sender: State.IWallet = walletManager.findByPublicKey(transaction.data.senderPublicKey);
-        sender.balance = sender.balance.plus(transaction.data.amount).plus(transaction.data.fee);
-        sender.decrementNonce();
+        const d: Interfaces.ITransactionData = transaction.data;
```
supaiku0: Please use better names or just destructure properly then you don't get this kind of variable names.

packages/core-transactions/src/handlers/transaction.ts
```diff
@@ -110,26 +106,50 @@ export abstract class TransactionHandler implements ITransactionHandler {
...
 
-    protected abstract applyToRecipient(
-        transaction: Interfaces.ITransaction,
-        walletManager: State.IWalletManager,
-    ): void;
-    protected abstract revertForRecipient(
-        transaction: Interfaces.ITransaction,
-        walletManager: State.IWalletManager,
-    ): void;
+    public abstract applyToRecipient(transaction: Interfaces.ITransaction, walletManager: State.IWalletManager): void;
```
supaiku0: Revert.

packages/core-transactions/src/handlers/transaction.ts
```diff
@@ -154,28 +174,6 @@ export abstract class TransactionHandler implements ITransactionHandler {
         return false;
     }
 
-    public applyToSenderInPool(transaction: Interfaces.ITransaction, poolWalletManager: State.IWalletManager): void {
```
supaiku0: Keep them, even if they may not be necessary.

packages/core-database/src/database-service.ts
vasild: The interface of `canBeApplied()` was flawed. In general there are two patterns:

A. Indicate the result via the return value. For example - `true` for success and `false` for failure. In this case the usage involves checking the returned value:
```
if (foo()) {
  code for success
} else {
  code for failure
}
```

B. Indicate the result by throwing an exception on failure and not throwing on success. In this case the return value is irrelevant, and so the function does not return anything (ie `void`). In this case the usage involves try/catch:
```
try {
  foo()
  code for success
} catch () {
  code for failure
}
```

It is arguable whether it is good to throw an exception to indicate an error in a user supplied input and that A should be preferred over B, but that is another topic.

Before this patch we had the interface of `canBeApplied()` be a mixture of A and B - it would throw on failure and return `true` on success. Which is flawed because the return value is redundant - the caller knows that if it returned, then it succeeded and the return value is *always* `true`.

So, please pick one of A or B, but not a mix of both. With this patch I have made it B.

What is this with tslint? Would it insist that every function returns something!? Frowning upon `void` functions in general?

packages/core-transaction-pool/src/connection.ts
vasild: After the changes in this patch the `fooInPool()` methods become identical to the `foo()` methods (where `foo` is one of `applyToSender`, `applyToRecipient`, `revertForSender`, `revertForRecipient`):.
```
fooInPool(x, y) {
  foo(x, y);
}
```
which is the *middle man code smell*, so I have fixed it by removing the `fooInPool()` methods had have their callers call just `foo()` instead.

It does not make sense to introduce middle man methods just to indicate where it will be called from. Imagine this:
```
paintItBlack(car) { ... }
paintItBlack(ford)
paintItBlack(audi)
paintItBlack(mazda)

paintItBlackFerrari(car) { paintItBlack(car) }
paintItBlackFerrari(ferrari)
```

packages/core-transaction-pool/src/wallet-manager.ts
vasild: Because this way the reader can be hinted about the type of the variable. You know:
```
const x: string = someMethod();
```
is to be preferred over
```
const x = someMethod();
```

packages/core-transactions/src/handlers/delegate-registration.ts
vasild: It makes no sense to have:
```
class A {
  public method1(x, y) {
    this.method2(x, y);
  }
  private method2(x, y) {
    ...
  }
}
```

packages/core-transactions/src/handlers/transaction.ts
vasild: The reason for my change is to make it clear to the reader what is the type of the `data` variable.
The previous code `const { data }: Interfaces.ITransaction = transaction;` not only does not show the type of the `data` variable, but it misleadingly contains another type `Interfaces.ITransaction` which is NOT the type of `data`.

packages/core-transactions/src/handlers/transaction.ts
vasild: That is a leftover, fixed, thanks!

packages/core-transactions/src/handlers/transaction.ts
vasild: Yes, indeed, good catch. Fixed in 5f3c02be1.

packages/core-transactions/src/handlers/transaction.ts
vasild: With destructuring it should actually be `const { data }: Interfaces.ITransaction = transaction;` but that would defeat one of the purposes of having types:
```
const variable: TypeOfVariable = expression;
```
VS
```
const { variable }: BlahSomeOtherType = expression;
```

packages/core-transactions/src/handlers/transaction.ts
vasild: Renamed `d` to `data`.

packages/core-transactions/src/handlers/transaction.ts
supaiku0: How is it misleading unless you don't understand how destructuring works. It doesn't `contain another type` at all. This just duplicates the field name for no reason.

packages/core-transactions/src/handlers/delegate-registration.ts
supaiku0: It makes sense when taking into account that it is not supposed to be called from the outside except by the pool. Now it gives the impression it's okay to call it from anywhere. Whatever, I don't insist on it.

packages/core-transaction-pool/src/connection.ts
supaiku0: See other comment.

packages/core-database/src/database-service.ts
supaiku0: Then `canBeApplied` needs at least a different name that doesn't imply it returns a boolean.

packages/core-transactions/src/handlers/transaction.ts
vasild: `const { variable }: Type = expression;` is misleading because `Type` is not the type of `variable`.

packages/core-database/src/database-service.ts
vasild: Right, `canBeApplied` gives the impression that it would return a boolean. What about `checkApplicability` or `checkIfCanBeApplied`? Btw, I would have coded this using variant A - returning boolean true/false and no exceptions to indicate wrong user input, but I kept it with the exceptions because I thought that the author likes the exceptions. Maybe I should change it to A. and keep the name `canBeApplied`?

supaiku0: Without exceptions you lose the information about why it failed. Meaning you need to refactor a lot more than just the return value. So I wouldn't change it for the sake of changing it. There are more important issues to work on.

`throwIfCannotBeApplied()`?

packages/core-transactions/src/handlers/transaction.ts
```diff
@@ -105,24 +103,48 @@ export abstract class TransactionHandler implements ITransactionHandler {
...
+        if (data.version > 1) {
+            if (!sender.nonce.plus(1).isEqualTo(data.nonce)) {
+                throw new UnexpectedNonceError(data.nonce, sender.nonce, false);
+            }
+
+            sender.nonce = data.nonce;
+        }
+
+        const newBalance: Utils.BigNumber = sender.balance.minus(data.amount).minus(data.fee);
+        assert(!newBalance.isNegative());
```
faustbrian:
```suggestion
        assert(newBalance.isPositive());
```

vasild: Positive is not the same as not negative.

packages/core-database/src/database-service.ts
vasild: Renamed.

packages/core-transactions/src/handlers/transaction.ts
vasild: Marking this as resolved without addressing it. Assuming wrong suggestion.

vasild:
> Merging #2587 into 2.6 will increase coverage by 42.87%

:-O I think codecov has gone really insane this time

PR #2611, opened 2019-05-23
fix(core-database): assign calculated delegate ranks from temp wallets

### LINKED ISSUE #2350: Delegate ranks in wallet manager are all over the place until the first block is received

supaiku0:
## Summary

Fixes #2350 this time for real.

- [X] Bugfix
- [ ] New feature
- [ ] Build-related changes
- [ ] Documentation
- [ ] Other, please describe:

- [ ] Yes
- [ ] No

- [ ] Yes
  - [ ] All tests are passing
- [ ] No

## Checklist

### DIFF
--- a/packages/core-database/src/database-service.ts
+++ b/packages/core-database/src/database-service.ts
@@ -632,7 +632,7 @@ export class DatabaseService implements Database.IDatabaseService {
         }
 
         // Now retrieve the active delegate list from the temporary wallet manager.
-        const delegates = tempWalletManager.loadActiveDelegateList(roundInfo);
+        const delegates: State.IDelegateWallet[] = tempWalletManager.loadActiveDelegateList(roundInfo);
 
         for (const delegate of tempWalletManager.allByUsername()) {
             this.walletManager.findByUsername(delegate.username).rate = delegate.rate;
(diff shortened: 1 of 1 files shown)


packages/core-database/src/database-service.ts
```diff
@@ -632,7 +632,13 @@ export class DatabaseService implements Database.IDatabaseService {
         }
 
         // Now retrieve the active delegate list from the temporary wallet manager.
-        return tempWalletManager.loadActiveDelegateList(roundInfo);
+        const delegates = tempWalletManager.loadActiveDelegateList(roundInfo);
```
faustbrian:
```suggestion
        const delegates: State.IDelegateWallet[] = tempWalletManager.loadActiveDelegateList(roundInfo);
```

PR #2622, opened 2019-05-27
refactor(core-transaction-pool): deserialize transactions before they leave the pool

faustbrian:
## Summary

Currently the assumption is that all data that is in the transaction pool is valid as it otherwise could not enter, which is obviously not true. In order to resolve that issue a few things have been changed.

- When the node loads all transactions from the the pool on boot every transaction that fails to deserialised or results in an unexpected id gets removed.
- Deserialise transactions before they leave the pool and apply the same rules as during boot
- The forger now calls `fromBytesUnsafe` instead of `fromBytes` because the relay will ensure that all data it sends out is valid

## What kind of change does this PR introduce? (check at least one)

- [ ] Bugfix
- [ ] Feature
- [x] Refactor
- [ ] Performance
- [ ] Tests
- [ ] Build-related change
- [ ] Documentation
- [ ] Code style update
- [ ] Continuous Integration
- [ ] Other, please describe:

## Does this PR introduce a breaking change? (check one)

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

- [x] All tests are passing

### DIFF
--- a/__tests__/integration/core-blockchain/blockchain.test.ts
+++ b/__tests__/integration/core-blockchain/blockchain.test.ts
@@ -17,6 +17,44 @@ let configManager;
 let container;
 let blockchain: Blockchain;
 
+const resetBlocksInCurrentRound = async () => {
+    await blockchain.database.loadBlocksFromCurrentRound();
+};
+
+const resetToHeight1 = async () => {
+    const lastBlock = await blockchain.database.getLastBlock();
+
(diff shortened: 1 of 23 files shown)


vasild:
> all data that is in the transaction pool is valid as it otherwise could not enter

That assumption was true at some point in the past. Why not restore it?

PR #2651, opened 2019-06-03
fix(core-snapshots): avoid trying to INSERT duplicates in rounds

vasild:
Follow the logic of INSERTing into blocks and transactions also for
rounds: only INSERT a row during snapshot restore if its identifier
is higher than the highest in the database (ie a sloppy check whether
a row is in the database).

- [ ] Bugfix
- [ ] Feature
- [ ] Refactor
- [ ] Performance
- [ ] Tests
- [ ] Build
- [ ] Documentation
- [ ] Code style update
- [ ] Continuous Integration
- [ ] Other, please describe:

- [ ] Yes
- [ ] No

- [ ] Yes
- [ ] No

- [ ] All tests are passing

### DIFF
--- a/packages/core-snapshots/src/manager.ts
+++ b/packages/core-snapshots/src/manager.ts
@@ -139,6 +139,7 @@ export class SnapshotManager {
 
         const lastBlock = await this.database.getLastBlock();
         params.lastBlock = lastBlock;
+        params.lastRound = await this.database.getLastRound();
         params.chunkSize = this.options.chunkSize || 50000;
 
         if (exportAction) {
--- a/packages/core-snapshots/src/manager.ts
+++ b/packages/core-snapshots/src/manager.ts
@@ -49,7 +49,7 @@ export class SnapshotManager {
 
         if (params.truncate) {
             await this.database.truncate();
-            params.lastBlock = undefined;
+            params.lastBlock = null;
         }
 
         await importTable("blocks", params);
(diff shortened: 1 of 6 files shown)


vasild:
> I run into an error when I do the following:
> 
> ```
> $ snapshot:dump
> (forge some more blocks)
> $ snapshot:dump --blocks previous-dump
> ...
> $ snapshot:truncate
> ...
> $ snapshot:restore
> 0sTypeError: Cannot read property 'id' of null
>     at Object.exports.canImportRecord (~/git/ark-core/packages/core-snapshots/dist/transport/verification.js:63:44)
>     at Object.exports.importTable (~/git/ark-core/packages/core-snapshots/dist/transport/index.js:84:28)
> ```

Fixed in a6e5b0c.

packages/core-snapshots/src/manager.ts
```diff
@@ -49,7 +49,7 @@ export class SnapshotManager {
 
         if (params.truncate) {
             await this.database.truncate();
-            params.lastBlock = undefined;
+            params.lastBlock = null;
```
faustbrian: - https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines#null-and-undefined
- https://basarat.gitbooks.io/typescript/docs/javascript/null-undefined.html

vasild: Changed back to `undefined` in d096d66f1 to silence the tslint warning.

Notice however that the second link says

> Something hasn't been initialized : undefined
> Something is currently unavailable: null

and in this case the last block is not available because there are no blocks, so `null` should be used. This is also in line with the `getLastBlock()` which returns `null` if there are no blocks.

PR #2672, opened 2019-06-07
refactor(core-transaction-pool): expire transactions that don't have …

vasild:
…expiration

If a transaction does not have an expiration value set by the user, then
we would keep it for a long time in the pool and also it could possibly
be re-sent by somebody (not the author) long time after it was created,
if it has not been forged yet.

Change this to expire after 6h even transactions that don't have an
user-set expiration. 6h considered since the transaction creation time.

- [ ] Bugfix
- [ ] Feature
- [x] Refactor
- [ ] Performance
- [ ] Tests
- [ ] Build
- [ ] Documentation
- [ ] Code style update
- [ ] Continuous Integration
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

- [ ] All tests are passing

### DIFF
--- a/packages/core-transaction-pool/src/memory.ts
+++ b/packages/core-transaction-pool/src/memory.ts
@@ -49,8 +49,8 @@ export class Memory {
                     return 1;
                 }
 
-                const expirationA: number = this.calcTxExpiration(a, expirationContext);
-                const expirationB: number = this.calcTxExpiration(b, expirationContext);
+                const expirationA: number = this.calculateTransactionExpiration(a, expirationContext);
+                const expirationB: number = this.calculateTransactionExpiration(b, expirationContext);
 
                 if (expirationA !== null && expirationB !== null) {
                     return expirationA - expirationB;
@@ -75,7 +75,9 @@ export class Memory {
 
         if (!this.byExpirationIsSorted) {
             this.byExpiration.sort(
-                (a, b) => this.calcTxExpiration(a, expirationContext) - this.calcTxExpiration(b, expirationContext)
+                (a, b) =>
+                    this.calculateTransactionExpiration(a, expirationContext) -
+                    this.calculateTransactionExpiration(b, expirationContext)
             );
             this.byExpirationIsSorted = true;
         }
@@ -83,7 +85,7 @@ export class Memory {
         const transactions: Interfaces.ITransaction[] = [];
 
         for (const transaction of this.byExpiration) {
-            if (this.calcTxExpiration(transaction, expirationContext) > currentHeight) {
+            if (this.calculateTransactionExpiration(transaction, expirationContext) > currentHeight) {
                 break;
             }
 
@@ -164,7 +166,7 @@ export class Memory {
             currentHeight: currentHeight,
             now: Crypto.Slots.getTime()
         };
-        const expiration: number = this.calcTxExpiration(transaction, expirationContext);
+        const expiration: number = this.calculateTransactionExpiration(transaction, expirationContext);
         if (expiration !== null) {
             this.byExpiration.push(transaction);
             this.byExpirationIsSorted = false;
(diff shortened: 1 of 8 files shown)


packages/core-transaction-pool/src/memory.ts
```diff
@@ -250,4 +275,43 @@ export class Memory {
...
+    /**
+     * Calculate the expiration height of a transaction.
+     * An expiration height H means that the transaction cannot be included in block at height
+     * H or any higher block.
+     * If the user did not specify an expiration height when creating the transaction then
+     * we calculate one from the timestamp of the transaction creation and the configured
+     * maximum transaction age.
+     * @return number expiration height or null if the transaction does not expire
+     */
+    private calcTxExpiration(
```
supaiku0: `calculateTransactionExpiration`

packages/core-transaction-pool/src/memory.ts
```diff
@@ -250,4 +275,43 @@ export class Memory {
...
+        if (transaction.data.expiration > 0) {
+            return transaction.data.expiration;
+        }
+
+        // Since the user did not specify an expiration we set one by calculating
+        // approximately the height of the chain as of the time the transaction was
+        // created and adding maxTransactionAge to that.
+
+        // Both now and transaction.data.timestamp use [number of seconds since the genesis block].
+        const txCreatedSecondsAgo: number = context.now - transaction.data.timestamp;
```
supaiku0: Please remove the `tx` prefix.

packages/core-transaction-pool/src/memory.ts
```diff
@@ -250,4 +275,43 @@ export class Memory {
...
+
+        const txCreatedBlocksAgo: number = Math.floor(txCreatedSecondsAgo / context.blockTime);
```
supaiku0: Same.

packages/core-transaction-pool/src/memory.ts
```diff
@@ -250,4 +275,43 @@ export class Memory {
...
+
+        const txCreatedAtHeight: number = context.currentHeight - txCreatedBlocksAgo;
```
supaiku0: Same.

alessiodf:
>... it could possibly be re-sent by somebody (not the author) long time after it was created, if it has not been forged yet.

I don't see how this PR mitigates that risk. Somebody who is not the author who wants to do that could easily add an arbitrary `expiration: someHeightInTheFuture` property to the transaction and broadcast it since the `expiration` value is not signed so can be manipulated and still pass validation and verification.

packages/core-transaction-pool/src/memory.ts
vasild: Done in 7a9b97211.

packages/core-transaction-pool/src/memory.ts
vasild: Done in 7a9b97211.

packages/core-transaction-pool/src/memory.ts
vasild: Done in 7a9b97211.

packages/core-transaction-pool/src/memory.ts
vasild: Done in 7a9b97211.

vasild:
> > ... it could possibly be re-sent by somebody (not the author) long time after it was created, if it has not been forged yet.
> 
> I don't see how this PR mitigates that risk. Somebody who is not the author who wants to do that could easily add an arbitrary `expiration: someHeightInTheFuture` property to the transaction and broadcast it since the `expiration` value is not signed so can be manipulated and still pass validation and verification.

Right, good catch! I assumed that the `expiration` field is signed, but that is the case only with v2 transactions. I changed the behavior to ignore the expiration field for v1 transactions (ie proceed as if it is 0) in 78d658499.

Thanks!

vasild:
> Please also add a test for it.

See the updated test at:

https://github.com/ArkEcosystem/core/pull/2672/files#diff-b0a172ced81203aed79dde98f43fc8e0R233

It creates a few transactions with version=1 and version=2 with different expiration values.

PR #2678, opened 2019-06-10
fix: do not sort transactions in forger / update purgeByBlock logic for handling nonces

air1one:
A few fixes related to nonces :
- (core-forger) do not sort transactions when forging (we already get sorted transactions from relay)
- (core-transaction-pool) update purgeByBlock to manage reverting transactions with higher nonces

- [x] Bugfix
- [ ] Feature
- [ ] Refactor
- [ ] Performance
- [ ] Tests
- [ ] Build
- [ ] Documentation
- [ ] Code style update
- [ ] Continuous Integration
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

- [x] All tests are passing
- [x] New/updated tests are included

### DIFF
--- a/packages/core-transaction-pool/src/connection.ts
+++ b/packages/core-transaction-pool/src/connection.ts
@@ -328,15 +328,7 @@ export class Connection implements TransactionPool.IConnection {
     }
 
     public purgeByBlock(block: Interfaces.IBlock): void {
-        // Revert in reverse order so that we don't violate nonce rules.
-        for (let i = block.transactions.length - 1; i >= 0; i--) {
-            const transaction = block.transactions[i];
-            if (this.has(transaction.id)) {
-                this.removeTransaction(transaction);
-
-                this.walletManager.revertTransactionForSender(transaction);
-            }
-        }
+        this.purgeTransactions(ApplicationEvents.TransactionPoolRemoved, block.transactions);
     }
 
     public purgeByPublicKey(senderPublicKey: string): void {
--- a/__tests__/unit/core-forger/delegate.test.ts
+++ b/__tests__/unit/core-forger/delegate.test.ts
@@ -2,9 +2,7 @@ import "jest-extended";
 
 import { Types, Utils } from "@arkecosystem/crypto";
 import { Delegate } from "../../../packages/core-forger/src/delegate";
-import { ITransactionData } from "../../../packages/crypto/src/interfaces";
 import { testnet } from "../../../packages/crypto/src/networks";
-import { sortTransactions } from "../../../packages/crypto/src/utils";
 import { TransactionFactory } from "../../helpers/transaction-factory";
 
 const dummy = {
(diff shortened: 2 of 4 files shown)


packages/core-transaction-pool/src/connection.ts
```diff
@@ -328,15 +328,7 @@ export class Connection implements TransactionPool.IConnection {
...
-        // Revert in reverse order so that we don't violate nonce rules.
-        for (let i = block.transactions.length - 1; i >= 0; i--) {
-            const transaction = block.transactions[i];
-            if (this.has(transaction.id)) {
-                this.removeTransaction(transaction);
-
-                this.walletManager.revertTransactionForSender(transaction);
-            }
-        }
+        this.purgeTransactions(ApplicationEvents.TransactionPoolRemoved, block.transactions);
```
vasild: `this.purgeTransactions()` will remove `block.transactions` in forward order (starting from 0). How come this does not break? If a block has the following transactions:
```
tx1
tx2
tx3 from sender A, nonce=5
tx4
tx5 from sender A, nonce=6
tx6
```
then tx5 should be reverted first and then tx3.

air1one: From what I see, `this.purgeTransactions()` will remove transactions from highest to lowest nonce :thinking: And that is what the comment says too :
```
// Revert all transactions that have bigger or equal nonces than the ones in
// lowestNonceBySender in order from bigger nonce to smaller nonce.
```
Can you confirm ?

vasild: I confirm that it is necessary to revert a given sender's transactions starting from the highest nonce - e.g. first tx5 (nonce=6) and then tx3 (nonce=5) from the above example. Otherwise, if we revert tx3 (nonce=5) first there will be a situation where we have a gap (there will be transactions with nonces 1,2,3,4,6, so nonce=5 is missing) which must not be allowed by the code to ever happen.

BUT:
> From what I see, this.purgeTransactions() will remove transactions from highest to lowest nonce

I don't think so:
```
    private purgeTransactions(event: string, transactions: Interfaces.ITransaction[]): void {
        for (const transaction of transactions) {
            this.emitter.emit(event, transaction.data);

            this.walletManager.revertTransactionForSender(transaction);

            this.memory.forget(transaction.id, transaction.data.senderPublicKey);

            this.syncToPersistentStorageIfNecessary();
        }
    }
```
`this.purgeTransactions()` will remove `block.transactions` in forward order (starting from 0). The transactions in a block are ordered from low to high nonce (for a given sender).

It is a bit worrying how this change did not break any test...

air1one: Are you looking at the right branch ? :smile: This is based on 2.6.

vasild: I was looking at `purgeTransactions()` in the `develop` branch, but it is much more involved in the `2.6` branch:

https://github.com/ArkEcosystem/core/blob/2.6/packages/core-transaction-pool/src/connection.ts#L580

So, all is good and no need to worry why no tests broke :)

PR #2792, opened 2019-07-10
fix(core-forger): also check if still in same slot

supaiku0:

- [X] Bugfix
- [ ] Feature
- [ ] Refactor
- [ ] Performance
- [ ] Tests
- [ ] Build
- [ ] Documentation
- [ ] Code style update
- [ ] Continuous Integration
- [ ] Other, please describe:

- [ ] Yes
- [ ] No

- [ ] Yes
- [ ] No

- [ ] All tests are passing

### DIFF
--- a/__tests__/unit/core-forger/manager.test.ts
+++ b/__tests__/unit/core-forger/manager.test.ts
@@ -48,7 +48,7 @@ describe("Forger Manager", () => {
             const del = new Delegate("a secret", testnet.network);
             const round = {
                 lastBlock: { id: sampleBlock.data.id, height: sampleBlock.data.height },
-                timestamp: 1,
+                timestamp: Crypto.Slots.getTime(),
                 reward: 2 * 1e8,
             };
 
@@ -92,7 +92,7 @@ describe("Forger Manager", () => {
             const del = new Delegate("a secret", testnet.network);
             const round = {
                 lastBlock: { id: sampleBlock.data.id, height: sampleBlock.data.height },
-                timestamp: 1,
+                timestamp: Crypto.Slots.getTime(),
                 reward: 2 * 1e8,
             };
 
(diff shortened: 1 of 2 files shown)


__tests__/unit/core-forger/manager.test.ts
```diff
@@ -115,6 +115,20 @@ describe("Forger Manager", () => {
             });
 
             expect(forgeManager.client.broadcastBlock).not.toHaveBeenCalled();
+
+            jest.resetAllMocks();
+
+            jest.spyOn(Crypto.Slots, "getTimeInMsUntilNextSlot").mockReturnValueOnce(2000);
+            jest.spyOn(Crypto.Slots, "getSlotNumber").mockImplementation(epoch => {
+                return epoch ? 1 : 2;
```
vasild: What about `s/epoch/timestamp/`?
And one or both of `(timestamp?) =>...` or `return timestamp !== undefined ? 1 : 2;` to hint that we are testing for `undefined`, not for `0`

packages/core-forger/src/manager.ts
```diff
@@ -170,11 +170,13 @@ export class ForgerManager {
...
-            this.logger.warn(
-                `Failed to forge new block by delegate ${this.usernames[delegate.publicKey]} (${
-                    delegate.publicKey
-                }, because there were less than ${minimumMs}ms left`,
-            );
+            if (currentSlot !== roundSlot) {
+                this.logger.warn(`Failed to forge new block by delegate ${prettyName}, because already in next slot.`);
+            } else {
+                this.logger.warn(
+                    `Failed to forge new block by delegate ${prettyName}, because there were less than ${minimumMs}ms left`,
```
vasild: May also be useful to print the time left and also print time left from what:
```
..., because there were ${timeLeftInMs}ms left in the current slot (less than ${minimumMs}ms)
```

__tests__/unit/core-forger/manager.test.ts
```diff
@@ -119,8 +119,8 @@ describe("Forger Manager", () => {
             jest.resetAllMocks();
 
             jest.spyOn(Crypto.Slots, "getTimeInMsUntilNextSlot").mockReturnValueOnce(2000);
-            jest.spyOn(Crypto.Slots, "getSlotNumber").mockImplementation(epoch => {
-                return epoch ? 1 : 2;
+            jest.spyOn(Crypto.Slots, "getSlotNumber").mockImplementation(timestamp => {
+                return timestamp === undefined ? 1 : 2;
```
vasild: The logic got reverted, previously we would return 1 if it was defined and now we would return 1 if it is not defined?

supaiku0: Erm, it did. But that doesn't change the outcome. The slots just have to be different.

vasild: ok

PR #2814, opened 2019-07-16
fix(core-api): return count of filtered peers

dated:
The peer api currently returns the total count of all peers, instead of the count of the filtered peers, thus messing with the pagination.

It's worth noting that the schema for index currently accepts `ip`, `status`, and `port` as query params, yet filtering based on those values is not implemented.

- [x] Bugfix
- [ ] Feature
- [ ] Refactor
- [ ] Performance
- [ ] Tests
- [ ] Build
- [ ] Documentation
- [ ] Code style update
- [ ] Continuous Integration
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

- [x] All tests are passing

### DIFF
--- a/packages/core-api/src/handlers/peers/controller.ts
+++ b/packages/core-api/src/handlers/peers/controller.ts
@@ -13,7 +13,7 @@ export class PeersController extends Controller {
             ? result.filter(peer => peer.version === (request.query as any).version)
             : result;
 
-        const count = result.length;
+        const count: number = result.length;
 
         result = result.slice(0, +request.query.limit || 100);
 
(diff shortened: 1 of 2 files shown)


faustbrian:
> It's worth noting that the schema for index currently accepts `ip`, `status`, and `port` as query params, yet filtering based on those values is not implemented.

1. `ip` filtering would have to be something like pattern matching but could be implemented.
2. `status` can be removed from the schema as all peers that don't have a good status are disconnected.
3. `port` can be removed from the schema as a fixed port is enforced so filtering it would do nothing.

packages/core-api/src/handlers/peers/controller.ts
```diff
@@ -12,6 +12,9 @@ export class PeersController extends Controller {
         result = request.query.version
             ? result.filter(peer => peer.version === (request.query as any).version)
             : result;
+
+        const count = result.length;
```
faustbrian:
```suggestion
        const count: number = result.length;
```

dated:
I've left `ip` in the schema to be implemented in a future PR, `status` and `port` have been removed.

PR #2851, opened 2019-07-30
fix(core-p2p): ensure payload database exists

faustbrian:

- [ ] Bugfix
- [ ] Feature
- [ ] Refactor
- [ ] Performance
- [ ] Tests
- [ ] Build
- [ ] Documentation
- [ ] Code style update
- [ ] Continuous Integration
- [ ] Other, please describe:

- [ ] Yes
- [ ] No

- [ ] Yes
- [ ] No

- [ ] All tests are passing

### DIFF
--- a/packages/core-p2p/package.json
+++ b/packages/core-p2p/package.json
@@ -36,6 +36,7 @@
         "better-sqlite3": "^5.4.0",
         "dayjs": "^1.8.15",
         "delay": "^4.3.0",
+        "fs-extra": "^8.1.0",
         "hapi-rate-limit": "^4.0.0",
         "ip": "^1.1.5",
         "ipaddr.js": "^1.9.0",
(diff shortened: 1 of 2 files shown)


alessiodf:
This is a bit dangerous now. For performance reasons, and the fact this is essentially a throwaway database, it runs with no journaling. This can lead to a corrupt sqlite database in the event Core crashes, which was previously fine because the database file, if it existed, was deleted and recreated afresh on every startup. Now that won't happen so if the actual database file is corrupted or locked or whatever, it may not be able to function on restart with unexpected behaviour.

faustbrian:
The bug was that on multiple machines the file was not created by `better-sqlite3` which is the reason the transaction pool uses the exact same code to ensure the file actually exists because this has been an issue in the past.

```
TypeError Cannot open database because the directory does not exist
    at new Database (core/node_modules/better-sqlite3/lib/database.js:31:9)
    at new PayloadProcessor (core/packages/core-p2p/dist/socket-server/payload-processor.js:23:32)
    at Object.<anonymous> (core/packages/core-p2p/dist/socket-server/payload-processor.js:120:28)
```

There are some other minor fixes coming so will include a combination of the two in the next release as that covers everything then.

PR #2873, opened 2019-08-12
feat(core-blockchain): enforce transactions' nonce order from blockPr…

vasild:
…ocessor

- [ ] Bugfix
- [ ] Feature
- [ ] Refactor
- [ ] Performance
- [ ] Tests
- [ ] Build
- [ ] Documentation
- [ ] Code style update
- [ ] Continuous Integration
- [ ] Other, please describe:

- [ ] Yes
- [ ] No

- [ ] Yes
- [ ] No

- [ ] All tests are passing

### DIFF
--- a/__tests__/unit/core-blockchain/processor/block-processor.test.ts
+++ b/__tests__/unit/core-blockchain/processor/block-processor.test.ts
@@ -248,6 +248,24 @@ describe("Block processor", () => {
                 database.getActiveDelegates = getActiveDelegatesBackup;
             });
 
+            it("should reject a block with invalid nonce order", async () => {
+                const getActiveDelegatesBackup = database.getActiveDelegates;
+                database.getActiveDelegates = jest.fn(() => [delegates[0]]);
+
+                const transactions = TransactionFactory.transfer(delegates[1].address)
+                    .withNetwork("unitnet")
+                    .withPassphrase(delegates[0].passphrase)
+                    .create(2);
+
+                const block = BlockFactory.fromData(getBlock([transactions[1], transactions[0]]));
+                block.verification.verified = true;
+
+                const processResult = await blockProcessor.process(block);
+                expect(processResult).toBe(BlockProcessorResult.Rejected);
+
+                database.getActiveDelegates = getActiveDelegatesBackup;
+            });
+
             it("should 'discard but broadcast' a block higher than current height + 1", async () => {
                 const blockVerified = BlockFactory.fromData(getBlock([]));
                 blockVerified.verification.verified = true;
--- a/__tests__/unit/core-blockchain/processor/block-processor.test.ts
+++ b/__tests__/unit/core-blockchain/processor/block-processor.test.ts
@@ -248,24 +248,6 @@ describe("Block processor", () => {
                 database.getActiveDelegates = getActiveDelegatesBackup;
             });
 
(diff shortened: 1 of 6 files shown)


packages/core-blockchain/src/processor/block-processor.ts
```diff
@@ -112,4 +117,44 @@ export class BlockProcessor {
...
+     * In addition, for a given sender, v2 transactions must have strictly
+     * increasing nonce without gaps.
+     */
+    private blockContainsOutOfOrderNonce(block: Interfaces.IBlock): boolean {
+        const nonceBySender = {};
+
+        for (const transaction of block.transactions) {
+            const data = transaction.data;
+
+            if (data.version < 2) {
```
supaiku0: I think if a block mixes V1 and V2 transactions it should be rejected.

packages/core-blockchain/src/processor/block-processor.ts
```diff
@@ -112,4 +117,44 @@ export class BlockProcessor {
...
+                continue;
+            }
+
+            const sender: string = data.senderPublicKey;
+
+            if (nonceBySender[sender] === undefined) {
+                nonceBySender[sender] = this.blockchain.database.walletManager.getNonce(sender);
+            }
+
+            const currentTransactionNonce = Utils.BigNumber.make(data.nonce);
```
supaiku0: `data.nonce` is already a bignumber so no need to make a new one.

__tests__/unit/core-blockchain/processor/block-processor.test.ts
```diff
@@ -248,6 +248,24 @@ describe("Block processor", () => {
                 database.getActiveDelegates = getActiveDelegatesBackup;
             });
 
+            it("should reject a block with invalid nonce order", async () => {
```
supaiku0: Can you instead create a new handler test like how it's done for the other handlers here: https://github.com/ArkEcosystem/core/blob/master/__tests__/unit/core-blockchain/processor/handlers/

Then it's also more clear wether it is really rejected because of the nonce order or if it's just a false positive.

packages/core-blockchain/src/processor/block-processor.ts
vasild: Done in 4b07b8b26.

packages/core-blockchain/src/processor/block-processor.ts
vasild: Simplified in 0d55b6d8b.

__tests__/unit/core-blockchain/processor/block-processor.test.ts
vasild: The handler is really simple:
```
export class NonceOutOfOrderHandler extends BlockHandler {
    public async execute(): Promise<BlockProcessorResult> {
        await super.execute();

        return BlockProcessorResult.Rejected;
    }
}
```

It does not make sense to write a test that creates a new class `NonceOutOfOrderHandler`, calls its `execute()` method and checks that it returns `BlockProcessorResult.Rejected` because that is already being done as part of the newly added test "should reject a block with invalid nonce order".

In `__tests__/unit/core-blockchain/processor/handlers/` we have tests only for the more complex handlers that contain their own logic.

As for the false positive, I modified the test a bit to verify that `NonceOutOfOrderHandler` will be used in 57e14f355

PR #2903, opened 2019-09-03
misc(core-blockchain): log the reason for discarding a block

vasild:
When processing blocks, if we find that some block is not chained then
we abandon the rest of the blocks. In this case, print a detailed
message explaining that a block is not chained and why.

Without this printout sync failures could be very obscure and difficult
to analyze.

A summary of what changes this PR introduces and why they were made.

## What kind of change does this PR introduce?

- [ ] Bugfix
- [ ] Feature
- [ ] Refactor
- [ ] Performance
- [ ] Tests
- [ ] Build
- [ ] Documentation
- [ ] Code style update
- [ ] Continuous Integration
- [ ] Other, please describe:

## Does this PR introduce a breaking change?

- [ ] Yes
- [ ] No

## Does this PR release a new version?

- [ ] Yes
- [ ] No

If yes, please describe the impact and migration path for existing applications:

**The PR fulfills these requirements:**

- [ ] It's submitted to the `develop` branch, _not_ the `master` branch
- [ ] All tests are passing
- [ ] New/updated tests are included

If adding a **new feature**, the PR's description includes:

- [ ] A convincing reason for adding this feature (to avoid wasting your time, it's best to open a suggestion issue first and wait for approval before working on it)

PR #2905, opened 2019-09-03
fix(core-blockchain): remove too early check for a chained block

vasild:
In processBlocks() we would check whether the first downloaded block is
chained and if not then we would return from processBlocks(). However,
if that happens it leads to an "infinite" loop of downloading blocks,
not applying any of them until the nodejs process runs out of memory to
store all the blocks it has downloaded. A block that would cause this is
(on devnet): height=1199883, id=5174570296595098048.

Remove this early check. Later all blocks (not only the first one) are
verified whether chained or not.

- [ ] Bugfix
- [ ] Feature
- [ ] Refactor
- [ ] Performance
- [ ] Tests
- [ ] Build
- [ ] Documentation
- [ ] Code style update
- [ ] Continuous Integration
- [ ] Other, please describe:

- [ ] Yes
- [ ] No

- [ ] Yes
- [ ] No

- [ ] All tests are passing

### DIFF
--- a/packages/core-blockchain/src/blockchain.ts
+++ b/packages/core-blockchain/src/blockchain.ts
@@ -12,7 +12,7 @@ import {
 } from "@arkecosystem/core-interfaces";
 import { Blocks, Crypto, Interfaces, Managers } from "@arkecosystem/crypto";
 
-import { isBlockChained, roundCalculator } from "@arkecosystem/core-utils";
+import { roundCalculator } from "@arkecosystem/core-utils";
 import async from "async";
 import delay from "delay";
 import pluralize from "pluralize";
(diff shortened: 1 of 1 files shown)


vasild:
This PR supersedes https://github.com/ArkEcosystem/core/pull/2903. If we merge this one, then 2903 can be discarded.

vasild:
Adding "isException()" check would only solve the problem for block "height=1199883, id=5174570296595098048" in particular because we already have it listed as an exception.

However there is a bigger problem - due to this check the nodejs process crashes whenever we receive an unchained block for which we don't have an exception.

Checking the first of 10k blocks would indeed save us processing time if the first block turns out to be unchained. However that would only work in 1 of 10k cases - if the second or 10th or 100th block is unchained we will still process the remaining ones.

So, the pro of this check is that we save processing time in 1 of 10k cases. The con is that it could cause the nodejs process to crash. IMO it hurts more than it helps.

Another thing to consider is that at the time of this check (I think) we have not verified that the block is signed by an active delegate, so a malicious relay node could send us some arbitrary unchained block to cause us to crash.

supaiku0:
Superseded by #2903

PR #2919, opened 2019-09-11
fix(core-transactions): reject delegate resignation if not enough inactive delegates

supaiku0:
Core assumes to always have at least `activeDelegates` (i.e. 51) delegates so we mustn't allow dropping below that number if there are no other delegates to fill that spot.

- [X] Bugfix
- [ ] Feature
- [ ] Refactor
- [ ] Performance
- [ ] Tests
- [ ] Build
- [ ] Documentation
- [ ] Code style update
- [ ] Continuous Integration
- [ ] Other, please describe:

- [ ] Yes
- [ ] No

- [ ] Yes
- [ ] No

- [ ] All tests are passing

### DIFF
--- a/__tests__/functional/transaction-forging/delegate-resignation.test.ts
+++ b/__tests__/functional/transaction-forging/delegate-resignation.test.ts
@@ -1,5 +1,5 @@
 import { app } from "@arkecosystem/core-container";
-import { Database, State } from "@arkecosystem/core-interfaces";
+import { Database } from "@arkecosystem/core-interfaces";
 import { Identities } from "@arkecosystem/crypto";
 import { generateMnemonic } from "bip39";
 import { TransactionFactory } from "../../helpers/transaction-factory";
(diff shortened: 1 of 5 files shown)


vasild:
It is wrong to reject a resignation. A resignation is an act of good will from a delegate to announce that he is not going to continue forging anymore. We cannot forcefully make him continue forging when he stops doing so, even if we reject his resignation.

> Core assumes to always have at least activeDelegates (i.e. 51) delegates

this is what needs to be fixed

vasild:
> > It is wrong to reject a resignation. A resignation is an act of good will from a delegate to announce that he is not going to continue forging anymore. We cannot forcefully make him continue forging when he stops doing so, even if we reject his resignation.
> 
> I agree and disagree. There should always be a minimum number of forgers to keep the network running. If all resigned the network would be dead. As long as none of the genesis forgers resigned it won't ever reject a resignation.

True, but rejecting a resignation is not successfully enforce enough delegates. If the guys shut down their hardware, then, rejected or not, there will be less number of delegates forging blocks.

If all resign then the network is dead for sure, but what if some resign and still there are e.g. 45 active and forging (from the desired 51), still dead?

That said, rejecting a resignation is a noop IMO. It will not hurt, but will not help either. So, if you still feel this PR should go, then go for it.

supaiku0:
>  It will not hurt, but will not help either.

It is just a safeguard for now, because it prevents the chain from getting completely rekt when there are less than 51 non-resigned delegates. As said above when the consensus is getting refactored we will revise it.

PR #2941, opened 2019-09-19
perf: Avoid O(m*n) when filtering pool txs and simplify the code

vasild:
validateTransactions() would take an array of transaction objects and
return a subset of it, but just the serialized property, not the entire
transaction object.

The two callers of validateTransactions() would then filter the array
they provided to validateTransactions() so that it contains only the
returned valid transactions. But, given that the returned array is a
subset of the provided array that means that the returned array is
the same as the filtered input. On top of this the unnecessary filtering
was done using an ineffective O(m*n) algorithm.

So, change validateTransactions() to return the transaction objects
instead of just the serialized property and use the returned array
directly (avoid the unnecessary filtering altogether).

- [ ] Bugfix
- [ ] Feature
- [ ] Refactor
- [ ] Performance
- [ ] Tests
- [ ] Build
- [ ] Documentation
- [ ] Code style update
- [ ] Continuous Integration
- [ ] Other, please describe:

- [ ] Yes
- [ ] No

- [ ] Yes
- [ ] No

- [ ] All tests are passing

### DIFF
--- a/packages/core-transaction-pool/src/connection.ts
+++ b/packages/core-transaction-pool/src/connection.ts
@@ -53,11 +53,8 @@ export class Connection implements TransactionPool.IConnection {
         }
 
         this.emitter.once("internal.stateBuilder.finished", async () => {
-            const validTransactions = await this.validateTransactions(transactionsFromDisk);
-            transactionsFromDisk = transactionsFromDisk.filter(transaction =>
-                validTransactions.includes(transaction.serialized.toString("hex")),
-            );
-
+            // Remove from the pool invalid entries found in `transactionsFromDisk`.
+            await this.validateTransactions(transactionsFromDisk);
             await this.purgeExpired();
             this.syncToPersistentStorage();
         });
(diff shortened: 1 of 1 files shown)


packages/core-transaction-pool/src/connection.ts
```diff
@@ -53,11 +53,8 @@ export class Connection implements TransactionPool.IConnection {
         }
 
         this.emitter.once("internal.stateBuilder.finished", async () => {
-            const validTransactions = await this.validateTransactions(transactionsFromDisk);
-            transactionsFromDisk = transactionsFromDisk.filter(transaction =>
-                validTransactions.includes(transaction.serialized.toString("hex")),
-            );
-
+            // Remove from the pool invalid entries found in `transactionsFromDisk`.
+            await this.validateTransactions(transactionsFromDisk);
```
vasild: Notice: we don't need the `transactionsFromDisk` variable to be updated here. This code is executed just once.

PR #2948, opened 2019-09-20
refactor(core-transaction-pool): don't accept expired v1 transactions

dated:
The transaction pool currently accepts transactions that are already "expired" at the time they are posted to a node, instead of triggering the expiration error. This is because the expiration check is based solely on the `expiration` property, which v1 transactions do not have.

This change would require many tests to be adapted, and might not be worth the trouble, since the transactions are being correctly purged from the pool afterwards anyway. If anything it highly increases the UX.

@vasild can i have your opinion on this?

- [ ] Bugfix
- [ ] Feature
- [x] Refactor
- [ ] Performance
- [ ] Tests
- [ ] Build
- [ ] Documentation
- [ ] Code style update
- [ ] Continuous Integration
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

- [x] All tests are passing


dated:
fyi the tests have not been touched yet, as i wanted to know if this change is deemed necessary at all.

vasild:
The idea behind this change looks sound. However the implementation can be improved. It is bad to spoil the encapsulation of the `Memory` class. Can the `calculateTransactionExpiration()` method be moved to the `Transaction` class and then be used from both `Memory` and `Processor` classes?

dated:
Agreed, this was a dirty mockup just to showcase the situation. If the `Transaction` class is fine with you I'll move the method there.

vasild:
Yes, and then instead of `(this.pool.memory as any).calculateTransactionExpiration(transaction, expirationContext);` or `this.calculateTransactionExpiration(transaction, expirationContext);` we would have `transaction.calculateExpiration(expirationContext);` - much cleaner.

dated:
Thanks for the feedback :ok_hand:

packages/crypto/src/transactions/types/transaction.ts
```diff
@@ -97,4 +98,45 @@ export abstract class Transaction implements ITransaction {
...
+        const data: ITransactionData = this.data;
+        // We ignore data.expiration in v1 transactions because it is not signed
+        // by the transaction creator.
+        // TODO: check if ok
+        if (data.version >= 2) {
+            // tslint:disable-next-line:no-null-keyword
+            return data.expiration || null;
+        }
+
+        const maxTransactionAge = app.resolveOptions("transaction-pool").maxTransactionAge;
```
faustbrian: The crypto package is decoupled from core, it has no knowledge of the container and isn't allowed to have any.

It also shouldn't really be the task of the transaction itself to calculate the expiration but that is an issue for another day.

dated:
Apparently v2 transactions without an explicit `expiration` do not expire?

https://github.com/ArkEcosystem/core/blob/07d54dd31fade4d9789a3da838259e2f0c2c8b40/packages/crypto/src/transactions/types/transaction.ts#L108

07d54dd should therefore be the only adjustment needed to make the functional tests pass again.

packages/core-transaction-pool/src/processor.ts
```diff
@@ -176,11 +172,30 @@ export class Processor implements TransactionPool.IProcessor {
...
+            .getStore()
+            .getLastHeight();
+
+        const maxTransactionAge: number = app.resolveOptions("transaction-pool").maxTransactionAge;
+
+        const expirationContext = {
+            blockTime: Managers.configManager.getMilestone(lastHeight).blocktime,
+            currentHeight: lastHeight,
+            now: Crypto.Slots.getTime(),
+            maxTransactionAge,
```
vasild: Use `app.resolveOptions("transaction-pool").maxTransactionAge` directly here and ditch the `maxTransactionAge` variable - it is used in just one place.

packages/crypto/src/transactions/types/transaction.ts
```diff
@@ -97,4 +97,44 @@ export abstract class Transaction implements ITransaction {
...
         return false;
     }
+
+    /**
+     * Calculate the expiration height of a transaction.
+     * An expiration height H means that the transaction cannot be included in block at height
+     * H or any higher block.
+     * If the user did not specify an expiration height when creating the transaction then
+     * we calculate one from the timestamp of the transaction creation and the configured
+     * maximum transaction age.
```
vasild: Append `...age (for v1 transactions). v2 transactions have a dedicated expiration property`.

vasild:
> Apparently v2 transactions without an explicit `expiration` do not expire?

Correct.

dated:
i'll try to debug the failing integration tests, at the moment i have no idea what causes the `TransactionVersionError` to be thrown.

packages/core-transaction-pool/src/processor.ts
```diff
@@ -176,11 +172,28 @@ export class Processor implements TransactionPool.IProcessor {
...
+            .getLastHeight();
+
+        const expirationContext = {
+            blockTime: Managers.configManager.getMilestone(lastHeight).blocktime,
+            currentHeight: lastHeight,
+            now: Crypto.Slots.getTime(),
+            maxTransactionAge: app.resolveOptions("transaction-pool").maxTransactionAge,
+        };
+
+        const transactionInstance: Interfaces.ITransaction = Transactions.TransactionFactory.fromData(transaction);
```
supaiku0: We cannot do this. This call is not just expensive, but also the reason why the unit tests are blowing up now. I'd just move the expiration calculation into a utility function inside `core-transaction-pool` or `core-utils` since that doesn't really fit semantically into the transaction class in the first place.

dated: :ok_hand:

vasild: > We cannot do this. This call is not just expensive, but also the reason why the unit tests are blowing up now.

The code before this patch already calls `fromData()`, next to the only invocation of `validateTransaction()`:

```
111 } else if (await this.validateTransaction(transaction)) {
112     try {
113         const receivedId: string = transaction.id;
114         const transactionInstance: Interfaces.ITransaction = Transactions.TransactionFactory.fromData(
115             transaction,
116         );
```

How could calling `fromData()` one more time break any test or be problematic in any way, other than performance wise? If that call is really expensive, then we can unite the 2 calls into a single one like this:

```
111 } else {    
112     const transactionInstance: Interfaces.ITransaction = Transactions.TransactionFactory.fromData(
113         transaction,
114     );  
115     if (await this.validateTransaction(transactionInstance)) {
```
and adjust `validateTransaction()` to take argument of type `ITransaction` instead of `ITransactionData`.

> the expiration calculation ... doesn't really fit semantically into the transaction class in the first place.

Why? Can you justify your reasoning?

I asked to get the expiration as a method of `ITransaction`. Here is my reasoning: the expiration is a property of a transaction, just like its id, amount, sender, or any of the other properties. Thus the expiration naturally belongs to the transaction class. But it cannot be a member variable because it is calculated differently for v1 and v2 transactions, so that naturally leads to having a method instead of a member variable, a method that does the calculation behind the scenes. It could even be a getter method (`get expiration()...`).

supaiku0: > How could calling fromData() one more time break any test or be problematic in any way, other than performance wise? 

Because of a missing try catch.

> Why? Can you justify your reasoning?
Following that reasoning, we should also have signature calculation, id calculation and everything else part of the transaction class. While the calculated expiration is only needed by the transaction pool. Plus, what if the context is extended and at one point we need access to the container for it. Well, then it's out of scope for the transaction class for real.

vasild: > > How could calling fromData() one more time break any test or be problematic in any way, other than performance wise?
> 
> Because of a missing try catch.

Well, then it can be added:
```
111 } else {
      try {    
112     const transactionInstance: Interfaces.ITransaction = Transactions.TransactionFactory.fromData(
113         transaction,
114     );
      } catch (...) {...}
115   if (await this.validateTransaction(transactionInstance)) {
```

> > Why? Can you justify your reasoning?
> 
> Following that reasoning, we should also have signature calculation, id calculation and everything else part of the transaction class.

Yes. You do `person.name()` not `calculatePersonName(person)`. I am not sure about "everything else". If some context is needed to get the name it is provided by the caller, e.g. `person.name(english)`, not `calculatePersonName(person, english)`. Also, this way if the calculation is expensive the class may cache it transparently and on subsequent calls to `name()` return the pre-calculated value.

> While the calculated expiration is only needed by the transaction pool.

Where it is used is irrelevant - if a property belongs to a class (e.g. name to a person, expiration to a transaction) it should be in that class. Also, where it is used from could change in the future.

> Plus, what if the context is extended and at one point we need access to the container for it. Well, then it's out of scope for the transaction class for real.

supaiku0: I don't agree with `person.name(english)` either, since the person shouldn't have to care about  the localization. But I get what you are saying and agree that there are cases where it makes sense. However, I don't think this is such a case. It pollutes the Transaction class just to handle legacy transactions and it forces you to create a transaction instance which could be as well avoided.

So to come to an end... the PR looks good to me now and I prefer it over the previous approach. :+1:

PR #2949, opened 2019-09-20
refactor: strengthen a nonce check in performGenericWalletChecks()

vasild:
refactor: strengthen a nonce check in performGenericWalletChecks()

    In TransactionHandler.performGenericWalletChecks() we would only check
    if the transaction's nonce is greater than the sender wallet nonce. The
    check should be not just greater, but exactly sender wallet nonce plus
    one.

    Also, move multiplicated code into two methods and use those methods.


supaiku0:
LGTM. Just need to sort out the failing tests.

vasild:
This change revealed a flaw in the code: if we receive more than one transaction for addition in the pool, in `Processor.validate()` / `filterAndTransformTransactions()` we would call `throwIfCannotBeApplied()` on each one, without applying any of them:
https://github.com/ArkEcosystem/core/blob/7bfba2710/packages/core-transaction-pool/src/processor.ts#L124
this, obviously, will not work wrt nonce - receiving two transactions from the same sender, we have to apply the first one so that the nonce is incremented before testing if the second can be applied.

There are a few solutions:

* Remove the `throwIfCannotBeApplied()` call from `filterAndTransformTransactions()`. It looks unnecessary because later `throwIfCannotBeApplied()` is called again from the pool's `Connection.addTransaction()`.

* If the call to `throwIfCannotBeApplied()` from `filterAndTransformTransactions()` is absolutely unnecessary, then a bigger refactor is due, changing https://github.com/ArkEcosystem/core/blob/7bfba2710/packages/core-transaction-pool/src/processor.ts#L28 to process transactions one by one: calling `filterAndTransformTransactions()`, `removeForgedTransactions()` and `addTransactionsToPool()` for one transaction before proceeding with the next.

* Workaround the flaw from the test, processing the transactions one by one:
```
             // reset processor, if not the 1st transaction will still be in this.accept and mess up
             processor = transactionPool.makeProcessor();

-            await processor.validate([votes[0].data, delegateRegs[0].data, signatures[0].data]);
+            await processor.validate([votes[0].data]);
+            processor = transactionPool.makeProcessor();
+            await processor.validate([delegateRegs[0].data]);
+            processor = transactionPool.makeProcessor();
+            await processor.validate([signatures[0].data]);

```

supaiku0:
We could also apply the transactions on a temporary wallet manager, but given that we already call apply (on the pool wallet manager) at the end of the processor, removing the check completely should also be fine.

PR #2994, opened 2019-09-29
refactor(core-api): integrate hapi-pagination to replace fork

faustbrian:
## Summary

Integrate the https://github.com/faustbrian/hapi-pagination repository directly into `core-api` and update all hapi dependencies to the `@hapi/*` namespace.

This gets rid of a bunch of dependencies and the need to keep a fork alive, also gives us the freedom to make any change we need to do. Also removes the warning about the outdated `hapi` scope when you install core.

- Removes a lot of configuration values. Only the page limit value can be configured.
- Removes a lot of logic that revolved around the configuration values which have been removed. This removes a little bit of overhead from each request.
- Route registrations are now set inside the plugin with a method and path to avoid the hacky method check logic. This is more explicit and reliable, either a route matches or it doesn't.

## Checklist

- [ ] Tests _(if necessary)_
- [x] Ready to be merged


packages/core-api/src/plugins/pagination/ext.ts
```diff
@@ -0,0 +1,154 @@
...
+        const newSource = {
+            meta: {
+                ...(source.meta || {}),
+                ...{
+                    count: results.length,
+                    pageCount: pageCount || 1,
+                    totalCount: totalCount ? totalCount : 0,
+
+                    next: getUri(currentPage + 1),
+                    previous: getUri(currentPage - 1),
```
dated:
```suggestion
                    // tslint:disable-next-line: no-null-keyword
                    previous: (totalCount && currentPage > 1 && currentPage <= pageCount + 1) ? getUri(currentPage - 1) : null,
```

packages/core-api/src/plugins/pagination/ext.ts
```diff
@@ -0,0 +1,154 @@
...
+
+        const newSource = {
+            meta: {
+                ...(source.meta || {}),
+                ...{
+                    count: results.length,
+                    pageCount: pageCount || 1,
+                    totalCount: totalCount ? totalCount : 0,
+
+                    next: getUri(currentPage + 1),
```
dated:
```suggestion
                    // tslint:disable-next-line: no-null-keyword
                    next: (totalCount && currentPage < pageCount) ? getUri(currentPage + 1) : null,
```

PR #3004, opened 2019-10-01
fix(crypto): schema convert to bignumber earlier

vasild:
Convert to bignumber even if some of the min/max checks would fail. The
caller may want to continue regardless of those failures and would
expect the property to be converted from string to bignumber.

## Checklist

- [ ] Tests _(if necessary)_
- [ ] Ready to be merged


vasild:
This patch is incomplete - if we have two properties that are supposed to be bignumbers and the first one fails the validation then the second one will not be converted.

PR #3015, opened 2019-10-03
perf(crypto): memoize base58 de/encoding

supaiku0:
## Summary

Memoize Base58 functions to speed up multipayment processing by up to 50%.
```
Before:
 ✓  Block.fromData (Multipayments 500 recipients) x 10.46 ops/sec ±1.14% (29 runs sampled)
After:
 ✓  Block.fromData (Multipayments 500 recipients) x 15.92 ops/sec ±3.32% (43 runs sampled)
```

## Checklist

- [ ] Tests _(if necessary)_
- [ ] Ready to be merged

### DIFF
--- a/packages/crypto/src/utils/base58.ts
+++ b/packages/crypto/src/utils/base58.ts
@@ -1,22 +1,26 @@
 import { base58 } from "bstring";
+import moize from "fast-memoize";
 import { HashAlgorithms } from "../crypto";
 
-export class Base58 {
-    public static encodeCheck(buffer: Buffer): string {
-        const checksum: Buffer = HashAlgorithms.hash256(buffer);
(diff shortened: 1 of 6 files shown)


packages/crypto/src/utils/base58.ts
```diff
@@ -1,22 +1,26 @@
...
+    if (checksum.readUInt32LE(0) !== buffer.slice(-4).readUInt32LE(0)) {
+        throw new Error("Invalid checksum");
     }
-}
+
+    return payload;
+};
+
+export const Base58 = {
+    encodeCheck: moize(encodeCheck),
```
vasild: How much data is this going to store in memory? There is no configuration or anything about that mentioned in https://www.npmjs.com/package/fast-memoize

PR #3019, opened 2019-10-03
perf: group multi-payment by senders and by recipients

vasild:
## Checklist

- [ ] Tests _(if necessary)_
- [ ] Ready to be merged


vasild:
This actually makes it slower, just posting a PR for the records, some ideas can be grabbed of it later.

PR #3051, opened 2019-10-10
chore(core): load core-p2p earlier

alessiodf:
## Summary

Many remote server monitoring scripts check that defined ports are open and trigger an alert if they are inaccessible. These alerts are often fired during database migration processes since `core-p2p` is loaded after `core-database-postgres`, so the p2p port is closed for a prolonged time during such periods. By loading `core-p2p` earlier, the port opens prior to the migrations, so remote port-based server monitoring scripts do not trigger false alarms.

Loading `core-p2p` sooner does not cause any problems because the `isAppReady` check returns `false` until both `core-transaction-pool` and `core-blockchain` are ready, and `core-blockchain` does not activate until `core-database-postgres` has finished.

I think this simple change is useful because it reduces false alarms generated by remote server monitoring tools and does not break anything.

I spoke to @vasild privately about this beforehand and he recommended that I make a PR to rearrange the order, along with justification for making the change to see what others think about it.

## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged


alessiodf:
I will look at the failing test later.

PR #3108, opened 2019-10-22
refactor(crypto): overwrite arrays when merging milestones

### LINKED ISSUE #3105: [2.6][T3] duplication of values in buildConstants()
It appears that while building the milestone constants there is an error in the `buildConstants()` method.

The last two milestones are defined as:

```
{
    "height": 3963000,
    "p2p": {
        "minimumVersions": ["^2.6 || ^2.6.0-next.0"]
    }
},
{
    "height": 4006000,
    "aip11": true
}
```

These get merged into:

```
{
    "p2p": {
        "minimumVersions": [
            "^2.6 || ^2.6.0-next.0",
            "^2.6 || ^2.6.0-next.0"
        ]
    },
    "aip11": true
}
```

as can be seen at https://dexplorer.ark.io/api/node/configuration

dated:
## Summary

Instead of concatenating arrays, overwrite them with the `source` array. In the particular case of `minimumVersions` we don't want to keep older values around, this PR is a solution targeted at this case and is not intended as a general solution for merging arrays - although as of now the `minimumVersions` array is the only array used in milestones.

Possibly closes #3105.

## Checklist

- [ ] Tests _(if necessary)_
- [x] Ready to be merged

### DIFF
--- a/__tests__/unit/crypto/managers/config.test.ts
+++ b/__tests__/unit/crypto/managers/config.test.ts
@@ -32,6 +32,18 @@ describe("Configuration", () => {
         expect(configManager.getMilestones()).toEqual(devnet.milestones);
     });
 
+    it('should build milestones without concatenating the "minimumVersions" array', () => {
+        configManager.setHeight(4006000);
+
+        const lastVersions = devnet.milestones
+            .sort((a, b) => a.height - b.height)
+            .find(milestone => !!milestone.p2p && !!milestone.p2p.minimumVersions);
+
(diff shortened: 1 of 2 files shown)


supaiku0:
Thanks! Can you also try to write tests for what you mentioned in #3105 and here?

The relevant tests would have to be added here:
https://github.com/ArkEcosystem/core/blob/master/__tests__/unit/crypto/managers/config.test.ts#L10

Issue #3133, opened 2019-10-25
[2.6][T4] MaximumPaymentCountExceededError shows incorrect received amount


Lemii:
When creating a multipayment with more than 500 payments, the error message will always show a received amount of '501', even though the number is greater than 501.

```
UnhandledPromiseRejectionWarning: MaximumPaymentCountExceededError: Expected a maximum of 500 payments, but got 501.
```

**Expected Behavior**
Show the correct amount of received payments or amend the error message to ommit the received amount.

**Your Environment**
Devnet 2.6

ArkEcosystemBot:
**Thanks for opening this issue!** A maintainer will review this in the next few days and explicitly select labels so you know what's going on.

_If no reviewer appears after a week, a reminder will be sent out._

faustbrian:
@Lemii Would you want to submit a PR to fix this? The error originates from https://github.com/ArkEcosystem/core/blob/0081a2fd2cdb123bbe764ddabae99f8da51b56b4/packages/crypto/src/transactions/builders/transactions/multi-payment.ts#L23.

Lemii:
Yep, traced it back to there also. I'll take the PR 👍

faustbrian:
Assigned you to the issue.

Lemii:
It appears that the `payments` array that is passed on to the error is always the length of the imposed size limit (in this case 500). 

Shall I remove the Error parameter and change the error message to omit the `, but got ${given}`?

faustbrian:
Someone could create a multi payment without that constraint and then we would have more than 500 in the array so that should be tested and than we could show the diff if possible.

Lemii:
> Someone could create a multi payment without that constraint and then we would have more than 500 in the array so that should be tested and than we could show the diff if possible.

Hmmm, just looked into this.. I'm not sure how the array that is passed in the `addPayment` method can be greater than the limit. How it is currently set up, it is impossible for the array to be greater than the limit because the error is thrown when it reaches that limit.

https://github.com/ArkEcosystem/core/blob/develop/packages/crypto/src/transactions/builders/transactions/multi-payment.ts

This `MaximumPaymentCountExceededError` is currently only used in the `addPayment` method of the `MultiPaymentBuilder`.

If someone somehow manages to create and broadcast a multipayment tx with a number of payments that exceed the limit, it would have bypassed this particular Error altogether since it didn't utilize the `addPayment` method of the `MultiPaymentBuilder`.

[...]

PR #3136, opened 2019-10-25
refactor(crypto): use multiPaymentLimit from milestone if available

### LINKED ISSUE #3135: [2.6][T4] hardcoded cap for maximum amount of transactions per multi payment
The `addPayment` method sets a hard cap for the amount of transactions that can be included in a multi payment transaction. Setting the milestone property `multiPaymentLimit` to a value higher than 500 will yield undesired results, since the hard cap will still be enforced.

https://github.com/ArkEcosystem/core/blob/61c4d5f2233099731712f0d28fe9718316c45691/packages/crypto/src/transactions/builders/transactions/multi-payment.ts#L21-L32

dated:
## Summary

Closes #3135.

## Checklist

- [ ] Tests _(if necessary)_
- [x] Ready to be merged

### DIFF
--- a/packages/crypto/src/transactions/builders/transactions/multi-payment.ts
+++ b/packages/crypto/src/transactions/builders/transactions/multi-payment.ts
@@ -1,5 +1,6 @@
 import { MaximumPaymentCountExceededError } from "../../../errors";
 import { ITransactionData } from "../../../interfaces";
+import { configManager } from "../../../managers";
 import { BigNumber } from "../../../utils";
 import { MultiPaymentTransaction } from "../../types";
 import { TransactionBuilder } from "./transaction";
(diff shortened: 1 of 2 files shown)


packages/crypto/src/transactions/builders/transactions/multi-payment.ts
```diff
@@ -19,8 +20,10 @@ export class MultiPaymentBuilder extends TransactionBuilder<MultiPaymentBuilder>
     }
 
     public addPayment(recipientId: string, amount: string): MultiPaymentBuilder {
-        if (this.data.asset.payments.length >= 500) {
-            throw new MaximumPaymentCountExceededError(this.data.asset.payments.length + 1);
+        const limit: number = configManager.getMilestone().multiPaymentLimit || 500;
+
+        if (this.data.asset.payments.length >= limit) {
+            throw new MaximumPaymentCountExceededError(limit, this.data.asset.payments.length + 1);
```
supaiku0: This should just be `payments.length > limit` and the `+1` should be removed. It was previously already wrong since `limit` is inclusive.

dated: Fixed!

dated: For that to work the transaction would need to be pushed to the array before checking the length actually 🤔 I will revert the last commit, #3133 will refactor this specific error anyway.

PR #3160, opened 2019-10-29
feat(core-api): add additional fields to businesses search schema

### LINKED ISSUE #3159: [2.6][T4] extend businesses search endpoint
The `businesses/search` endpoint only supports a `businessId` in it's payload schema. This should be extended to also support searching by `name`, `website`, `vat` and `repository`. 

@faustbrian, yes, i can submit a PR :trollface:

dated:
## Summary

Closes #3159.

## Checklist

- [ ] Tests _(if necessary)_
- [x] Ready to be merged

### DIFF
--- a/packages/core-api/src/handlers/businesses/schema.ts
+++ b/packages/core-api/src/handlers/businesses/schema.ts
@@ -47,7 +47,7 @@ export const search: object = {
             .integer()
             .min(1),
         name: Joi.string()
-            .pattern(/^[a-zA-Z0-9_-]+$/)
+            .regex(/^[a-zA-Z0-9_-]+$/)
             .max(40),
         website: Joi.string().max(50),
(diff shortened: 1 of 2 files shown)


packages/core-api/src/handlers/businesses/schema.ts
```diff
@@ -46,5 +46,13 @@ export const search: object = {
         businessId: Joi.number()
             .integer()
             .min(1),
+        name: Joi.string()
+            .pattern(/^[a-zA-Z0-9_-]+$/)
```
faustbrian: If I remember correct the `pattern` function only exists in joi 16, in joi 15 it should be `regex `.

dated: My bad, patch incoming.

PR #3162, opened 2019-10-29
refactor(core-magistrate-crypto): use URI schema for website and repository validation

### LINKED ISSUE #3106: [2.6][T3] Perform validation on the `website` field of a business registration/update
When creating a business registration transaction, it's currently possible to use any `website` value, as long as it's a string of at least 4 characters. Similarly to the `peer` validation, this field should contain additional validation to ensure that a user inputs a valid URL.

**Expected Behavior**  
An error should be returned when a user tries to create a business registration transaction with an invalid `website` value.

**Current Behavior**  
The transaction is created as long as the `website` value is a string of 4 characters.

**Possible Solution**  
Validate the `website` value against a RegEx that verifies that it is a valid URL.

There are many available expressions to do this, check out this reference for example: [https://mathiasbynens.be/demo/url-regex](https://mathiasbynens.be/demo/url-regex)

**Your Environment**  
- nodejs  
- @arkecosystem/crypto: 2.6.0-next.2

Lemii:
## Summary  

This Pull Request changes the following:

- Adds the URI schema
- Refactors `website`, `repository` and `bridgechainRepository` validation to use URI schema
- Updates all unit tests to comply with new URI schema (a lot of tests used 'ark.io' or similar, making the tests fail after this change)
- Adds new unit tests
- Adds new functional tests

Closes #3106 

## Checklist

- [X] Tests _(if necessary)_
- [X] Ready to be merged

## Comments  
This PR took a bit longer to make due to being forced of refactoring (and re-testing) all those tests.

### DIFF
--- a/packages/crypto/src/validation/schemas.ts
+++ b/packages/crypto/src/validation/schemas.ts
@@ -56,6 +56,11 @@ export const schemas = {
         allOf: [{ type: "string", pattern: "^[a-zA-Z0-9_-]+$" }, { minLength: 1, maxLength: 40 }],
     },
 
+    uri: {
+        $id: "uri",
+        allOf: [{ type: "string" }, { format: "uri" }, { minLength: 4, maxLength: 50 }],
+    },
+
     blockHeader: {
         $id: "blockHeader",
         type: "object",
--- a/packages/crypto/src/validation/schemas.ts
+++ b/packages/crypto/src/validation/schemas.ts
@@ -58,7 +58,7 @@ export const schemas = {
 
     uri: {
         $id: "uri",
-        allOf: [{ type: "string" }, { format: "uri" }, { minLength: 4, maxLength: 50 }],
+        allOf: [{ format: "uri" }, { minLength: 4, maxLength: 80 }],
     },
 
     blockHeader: {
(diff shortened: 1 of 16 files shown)


packages/crypto/src/validation/schemas.ts
```diff
@@ -56,6 +56,11 @@ export const schemas = {
         allOf: [{ type: "string", pattern: "^[a-zA-Z0-9_-]+$" }, { minLength: 1, maxLength: 40 }],
     },
 
+    uri: {
+        $id: "uri",
+        allOf: [{ type: "string" }, { format: "uri" }, { minLength: 4, maxLength: 50 }],
```
faustbrian: Is the `{ type: "string" }` part needed if we already have the URI expectation which logically would have to be a string?

Lemii: The ajv documentation could be interpreted as that it is required. 

> The following formats are supported for **string validation** with "format" keyword:
Source: https://www.npmjs.com/package/ajv#formats

However, I tested it without it and that seemed to work fine also.

Lemii: Would you like me to remove it?

faustbrian: We should probably also increase the `maxLength`. 50 is quite short as some people might have really long URLs, especially private chains or repositories hosted on custom domains.

Lemii: Good point. How about 80? 100 feels a bit excessive to me.

faustbrian: Lets go with 80 for now, we can increase it later on if it doesn't seem sufficient.

faustbrian:
@Lemii https://github.com/ArkEcosystem/core/pull/3160/files#diff-80bb865e0754fa368c6d341f7328c8f5 this has been merged so the schema needs to be updated from 50 length to 80 to match your changes.

PR #3165, opened 2019-10-29
feat(core-api): add additional fields to bridgechains search schema

### LINKED ISSUE #3164: [2.6][T4] extend bridgechains search endpoint
The `bridgechains/search` endpoint only supports a `bridgechainId` in its payload schema. This should be extended to also support searching by `businessId`, `name`, `seedNodes`, `genesisHash` and `bridgechainRepository`. 

@faustbrian, yes, i can submit a PR :trollface:

dated:
## Summary

Closes #3164.

## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged

### DIFF
--- a/packages/core-api/src/handlers/bridgechains/schema.ts
+++ b/packages/core-api/src/handlers/bridgechains/schema.ts
@@ -35,5 +35,16 @@ export const search: object = {
         bridgechainId: Joi.number()
             .integer()
             .min(1),
+        bridgechainRepository: Joi.string().max(80),
+        businessId: Joi.number()
+            .integer()
+            .min(1),
+        genesisHash: Joi.string()
+            .hex()
+            .length(64),
+        name: Joi.string()
+            .regex(/^[a-zA-Z0-9_-]+$/)
+            .max(40),
+        seedNodes: Joi.string().ip(),
     },
 };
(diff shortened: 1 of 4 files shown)


packages/core-api/src/handlers/bridgechains/schema.ts
```diff
@@ -35,5 +35,16 @@ export const search: object = {
...
+        businessId: Joi.number()
+            .integer()
+            .min(1),
+        genesisHash: Joi.string()
+            .hex()
+            .length(64),
+        name: Joi.string()
+            .regex(/^[a-zA-Z0-9_-]+$/)
+            .max(40),
+        seedNodes: Joi.string().ip(),
```
supaiku0: Shouldn't this accept an array or be named singular?

dated: Indeed, sorry about that. Will be fixed.

dated:
I've added a new filter, `every`, which returns `true` if all values of the given array are included in the checked property @supaiku0.

packages/core-database/src/repositories/utils/filter-rows.ts
```diff
@@ -68,6 +68,23 @@ export = <T = any>(rows: ReadonlyArray<T>, params: Database.IParameters, filters
             }
         }
 
+        if (filters.hasOwnProperty("every")) {
+            for (const elem of filters.every) {
+                if (params[elem] && getProperty(item, elem)) {
+                    if (Array.isArray(item[elem])) {
+                        if (Array.isArray(params[elem])) {
+                            // @ts-ignore
+                            return params[elem].every(a => item[elem].indexOf(a) > -1);
```
supaiku0:
```suggestion
                            return params[elem].every(a => item[elem].includes(a));
```

Issue #3088, opened 2019-10-20
default ordering of locks not sufficient


dated:
The default ordering used on the locks endpoint is not meaningful enough and the returned results could theoretically, be out of order (semantically speaking).

https://github.com/ArkEcosystem/core/blob/e928df76fb9a135ca8555a9d318c51ac11dd3762/packages/core-database/src/repositories/wallets-business-repository.ts#L213-L217

The expiration value would need to be properly normalised to guarantee that the returned order of locks matches the semantic order.

ArkEcosystemBot:
**Thanks for opening this issue!** A maintainer will review this in the next few days and explicitly select labels so you know what's going on.

_If no reviewer appears after a week, a reminder will be sent out._

dated:
Normalisation would return an estimate at best, since there is no way to precisely predict the timestamp of a future block, and viceversa, there is no way to predict the blockheight at a certain point in time.

ArkEcosystemBot:
This issue or pull request needs further investigation. Please wait for further information, thank you.

ArkEcosystemBot:
This issue has been closed. If you wish to re-open it please provide additional information.

PR #3181, opened 2019-10-30
chore(core-database): remove misleading order by expirationValue

vasild:
expirationValue could be either "height of the blockchain" or "seconds
since the genesis block", so sorting by it gives some misleading
impression that the returned results are sorted by expiration when they
are in fact not.

It is not possible to reliably convert from height to seconds, thus
ditch the ordering by expirationValue and order by lockId which will at
least deterministically return the results in the same order every time.

Closes https://github.com/ArkEcosystem/core/issues/3088

## Checklist

- [ ] Tests _(if necessary)_
- [ ] Ready to be merged

### DIFF
--- a/packages/core-database/src/repositories/wallets-business-repository.ts
+++ b/packages/core-database/src/repositories/wallets-business-repository.ts
@@ -213,7 +213,7 @@ export class WalletsBusinessRepository implements Database.IWalletsBusinessRepos
         return {
             query,
             entries,
-            defaultOrder: ["expirationValue", "asc"],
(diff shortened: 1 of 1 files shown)


dated:
Can they be ordered by sequence? That'd be a better fit than lockId.

vasild:
> Can they be ordered by sequence? That'd be a better fit than lockId.

Why? Sequence makes sense within one block only. Mix transactions from different blocks, sort them by sequence and again you get something that looks like one thing but is not.

dated:
My bad, nevermind. I misremembered what the sequence is used for.

PR #3210, opened 2019-11-05
fix(core-transactions): get locks attribute or empty object

### LINKED ISSUE #3116: TypeError: Cannot set property during HtlcClaimTransactionHandler.revertForSender
During a rollback, a `TypeError` was thrown during `HtlcClaimTransactionHandler.revertForSender`:

```
TypeError: Cannot set property 'cca606eccae695c08528e968035f19b1028fe0feab688421e63d0e6aa0507419' of undefined
at HtlcClaimTransactionHandler.revertForSender (/home/ark/.config/yarn/global/node_modules/@arkecosystem/core-transactions/dist/handlers/htlc-claim.js:121:35)
```

The offending code is https://github.com/ArkEcosystem/core/blob/bf64a68a045583b963edd1250ae228d1dac369fc/packages/core-transactions/src/handlers/htlc-claim.ts#L185-L192

As `locks` was undefined. Afterwards no new blocks were accepted until the node was manually restarted.

dated:
## Summary

Fixes #3116.

## Checklist

- [ ] Tests _(if necessary)_
- [x] Ready to be merged

### DIFF
--- a/packages/core-transactions/src/handlers/htlc-claim.ts
+++ b/packages/core-transactions/src/handlers/htlc-claim.ts
@@ -183,7 +183,7 @@ export class HtlcClaimTransactionHandler extends TransactionHandler {
         const lockedBalance: Utils.BigNumber = lockWallet.getAttribute("htlc.lockedBalance", Utils.BigNumber.ZERO);
         lockWallet.setAttribute("htlc.lockedBalance", lockedBalance.plus(lockTransaction.amount));
 
-        const locks: Interfaces.IHtlcLocks = lockWallet.getAttribute("htlc.locks");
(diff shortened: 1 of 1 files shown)


supaiku0:
This isn't a proper fix, since having no lock during revert means the lock wasn't applied in the first place.  In the linked issue the bug occurs in two different places, during a rollback and during wallet building as part of the bootstrap. The former should be fixed with #3155 since it reapplied the transactions in the wrong order at the end of a rollback which would lead to the issue described. The same seems to be the case for `buildWallets` during bootstrap.

dated:
Agreed. It was a dirty patch to get my node started. I didn't think this through obviously :) I was messing around with HTLCs at the time, I am now trying to reproduce the issue. Closing this.

PR #3229, opened 2019-11-08
refactor(crypto): adjust generic name schema

### LINKED ISSUE #3223: [2.6][T4] overly strict genericName regex
Imho business and bridgechain names should be allowed to include spaces, which is currently not the case.

https://github.com/ArkEcosystem/core/blob/5dbc8a8fcac6ad41d066d7cfa4ca2c1319d99cac/packages/crypto/src/validation/schemas.ts#L54-L57

I see no reasons to disallow spaces and many to allow them: searchability, display (in a business directory for instance, where the names would have to be parsed for aesthetic reasons -> trouble when replacing either `_` or `-` with spaces as this could result in legitimate symbols being removed), ...

The `genericName` should be adjusted to something like the following to allow spaces in the middle (untested as of now):

```
^[a-zA-Z0-9]+( [a-zA-Z0-9\._-]+)*$
```

dated:
## Summary

Closes #3223.

## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged


supaiku0:
@dated 
This breaks the following business registration: https://dexplorer.ark.io/api/transactions/b7cd7621f0dcbc030951e83e44f169a167c568acb91b32c2705499e246badf84?transform=false

The error I get:
```
data:"business-882686643483"
dataPath:".transactions[0].asset.businessRegistration.name"
keyword:"pattern"
message:"should match pattern "^[a-zA-Z0-9]+( [a-zA-Z0-9._-]+)*$""
params:Object {pattern: "^[a-zA-Z0-9]+( [a-zA-Z0-9._-]+)*$"}
parentSchema:Object {type: "string", pattern: "^[a-zA-Z0-9]+( [a-zA-Z0-9._-]+)*$"}
schema:"^[a-zA-Z0-9]+( [a-zA-Z0-9._-]+)*$"
schemaPath:"genericName/allOf/0/pattern"
```
However, the name should be valid.

dated:
You are right, my bad. A better regex would be `^[a-zA-Z0-9]+([ ._-][a-zA-Z0-9.]+)*$`

PR #3252, opened 2019-11-12
fix(core-transactions): do not attempt to convert vendorfield

### LINKED ISSUE #3251: [2.6][T4] vendorfield conversion from buffer to string in htlc transactions
#3213 introduced a bug, i.e. the `vendorField` is converted from Buffer to string in the `applyToSender` and `revertForSender` methods. This conversion is only needed during bootstrap, and otherwise cause a wrong `vendorField` (i.e. empty string) to be inserted into the wallet attributes instead.

To reproduce:

- send htlc lock with vendorfield (e.g. `foobar`)
- visit http://NODE_IP:4003/api/locks/LOCK_ID → vendorField is `""`
- restart node
- visit http://NODE_IP:4003/api/locks/LOCK_ID → vendorField is restored (e.g. `"foobar"`)

dated:
## Summary

Fixes #3251.

## Checklist

- [ ] Tests _(if necessary)_
- [x] Ready to be merged

### DIFF
--- a/packages/core-transactions/src/handlers/htlc-claim.ts
+++ b/packages/core-transactions/src/handlers/htlc-claim.ts
@@ -175,7 +175,9 @@ export class HtlcClaimTransactionHandler extends TransactionHandler {
             amount: lockTransaction.amount,
             recipientId: lockTransaction.recipientId,
             timestamp: lockTransaction.timestamp,
-            vendorField: lockTransaction.vendorField,
+            vendorField: lockTransaction.vendorField
+                ? Buffer.from(lockTransaction.vendorField, "hex").toString("utf8")
+                : undefined,
             ...lockTransaction.asset.lock,
         };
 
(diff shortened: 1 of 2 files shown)


packages/core-transactions/src/handlers/htlc-claim.ts
```diff
@@ -175,9 +175,7 @@ export class HtlcClaimTransactionHandler extends TransactionHandler {
             amount: lockTransaction.amount,
             recipientId: lockTransaction.recipientId,
             timestamp: lockTransaction.timestamp,
-            vendorField: lockTransaction.vendorField
-                ? Buffer.from(lockTransaction.vendorField, "hex").toString("utf8")
```
supaiku0: This needs to stay, only the HTLC lock is wrong. The reverts get Buffers from the database.

PR #3255, opened 2019-11-13
feat(core-magistrate-crypto): add ports to bridgechain registration/update

dated:
## Summary

Adds a `ports` property to the bridgechain transaction types. This property holds an object with plugin names as keys, and port numbers as values like the `ports` property on peers. For the time being the only allowed plugin to be included is `@arkecosystem/core-api`.

## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged


dated:
Sure, I'll have a look as soon as possible.

dated:
#3258 might require a rollback - is that up for discussion at the moment after all?

PR #3256, opened 2019-11-14
fix(core-transaction-pool): wallet-manager fallback to database wallet manager findByIndex() when no "local" match

### LINKED ISSUE #3242: [2.6][T3] htlc refunds of bootstrapped locks cannot be applied to sender but increases nonce
HTLC refunds currently throw an error in the `applyToSender` method (probably since #3183).

The assertions in the method fail since the wrong wallet manager is used to retrieve the lockWallet and lock.

air1one:
## Summary

Resolves #3242 

Transaction pool wallet manager : try database wallet manager findByIndex() when local findByIndex() doesn't return a match.

We might not have all existing wallets indexes on transaction pool wallet manager : so in case we get no match we should try with database wallet manager.

## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged

### DIFF
--- a/__tests__/integration/core-transaction-pool/wallet-manager.test.ts
+++ b/__tests__/integration/core-transaction-pool/wallet-manager.test.ts
@@ -224,7 +224,7 @@ describe("Apply transactions and block rewards to wallets on new block", () => {
     });
 });
 
-describe.only("findByIndex", () => {
+describe("findByIndex", () => {
     describe("when transaction pool wallet manager does not find a wallet by index", () => {
         it("should get the wallet from database wallet manager if it exists", async () => {
             const publicKey = "02664fe58caa4a960ed74169a5968a5f69587ba50b75087d268f5788af3a5bf56d";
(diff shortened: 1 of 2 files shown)


dated:
With these changes the following lines become obsolete, is that right?

https://github.com/ArkEcosystem/core/blob/8f028f8cafa096e7c3cd8d811f46f91719deae96/packages/core-transactions/src/handlers/htlc-refund.ts#L53-L56

Instead, we can call `findByIndex` on the wallet manager passed to the `throwIfCannotBeApplied` method directly.

dated:
And another question for my personal understanding: should the Wallet Managers be allowed to be out of sync at all?

supaiku0:
@dated Initially the PoolWalletManager has no wallets and clones every database wallet that is missing. The reason why it is an issue for HTLC is because `findByIndex` is called directly, which doesn't have the fallback to the database wallets as `findByAddress/findByPublicKey` do. This PR adds the fallback.

> With these changes the following lines become obsolete, is that right?

Yes, these should be obsolete now. @air1one

dated:
Understood! Thanks for the explanation!

air1one:
> With these changes the following lines become obsolete, is that right?

No I think it's better to keep it this way, we actually want to check just against the database wallets, with associated lock transaction forged. (we don't want to validate if the lock transaction is still in the pool and was applied just to tx pool wallet, even though in theory this shouldn't happen)

PR #3258, opened 2019-11-15
fix(core-magistrate): use genesisHash for bridgechainId

### LINKED ISSUE #3257: [2.6][T4] can't apply bridgechain resignation or update without bootstrap
Bridgechain resignations or updates cannot be applied to newly registered bridgechains without bootstrapping inbetween. Similar to #3242, in the sense that the wallet manager being out of sync is probably the reason behind it.

This may also apply for businesses (untested).

To reproduce:

1. send bridgechain registration and wait for it to be forged
2. send bridgechain resignation or update

```
{
  "data": {
    "accept": [],
    "broadcast": [],
    "excess": [],
    "invalid": [
      "9c748f9d940a2b00f9d523c84b12b5757ee45b4936673e9e851788fe168a9799"
    ]
  },
  "errors": {
    "9c748f9d940a2b00f9d523c84b12b5757ee45b4936673e9e851788fe168a9799": [
      {
        "type": "ERR_APPLY",
        "message": "Failed to apply transaction, because bridgechain is not registered."
      }
    ]
  }
}
```

air1one:
## Summary

Resolves #3257 

Use bridgechain genesisHash as id instead of using some index based on current bridgechains registered.

The current bridgechain id logic is causing the #3257 issue , relying on wallet current registered bridgechains is a bit "weak" in my opinion (and as we have database vs tx pool wallet manager which can hold different states, it is dangerous).
Using genesisHash makes sense as it is enforced as unique and defines the bridgechain.

Also, in my opinion **business id should be updated too** as it uses the same mecanism and can be bug-prone. (will open an issue for it)

## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged

**Important** : current devnet nodes cannot just update with this and be fine, as their database contains "old" ids and will mess up a lot, 2 options : 
- roll back to before the 1st typeGroup 2 transaction
- use an upgrade script (but doesn't make sense in my opinion as this is not on mainnet, better skip upgrade script when we can)

### DIFF
--- a/packages/core-magistrate-crypto/src/transactions/bridgechain-update.ts
+++ b/packages/core-magistrate-crypto/src/transactions/bridgechain-update.ts
@@ -27,7 +27,9 @@ export class BridgechainUpdateTransaction extends Transactions.Transaction {
                             type: "object",
                             required: ["bridgechainId", "seedNodes"],
                             properties: {
-                                bridgechainId: { type: "integer", minimum: 1 },
+                                bridgechainId: {
+                                    $ref: "transactionId",
+                                },
                                 seedNodes: seedNodesSchema,
                             },
                         },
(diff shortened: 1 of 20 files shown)


packages/core-magistrate-crypto/src/transactions/bridgechain-update.ts
```diff
@@ -54,7 +56,7 @@ export class BridgechainUpdateTransaction extends Transactions.Transaction {
         }
 
         const buffer: ByteBuffer = new ByteBuffer(4 + 1 + seedNodesBuffersLength + seedNodes.length, true);
```
dated: This should be `32` instead of `4` now, right? Afaik the four bytes were reserved for the bridgechain id, which was an unsigned 32bit integer.

supaiku0: Yep. See https://github.com/ArkEcosystem/core/pull/3284

PR #3272, opened 2019-11-18
chore: handle Debian/Ubuntu derivate NodeJS install

### LINKED ISSUE #3261: install.sh fails at nodejs installation step
**Expected Behavior**

install.sh script should succeed without manual any manual changes.

**Current Behavior**

On node.js & npm installation, the following error happens and the script stops:
```sh
E: The repository 'https://deb.nodesource.com/node_12.x tina Release' does not have a Release file.
N: Updating from such a repository can't be done securely, and is therefore disabled by default.
N: See apt-secure(8) manpage for repository creation and user configuration details.
```

**Possible Solution**

Install a different version of nodejs in the script, or automagically retrieve the latest 12.x release url of nodejs.

**Steps to Reproduce (for bugs)**

1. Run ark core's `bash install.sh` on a fresh machine.
2. Wait until node installation
3. Observe error

**Your Environment**

- Version used: 2.6 from develop branch
- Operating System and version: Ubuntu 18.x

adrian69:
# Summary
Fixes #3261
 
## Checklist
- [x] Ready to be merged

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -112,7 +112,7 @@ if [[ ! -z $DEB ]]; then
     sudo wget --quiet -O - https://deb.nodesource.com/gpgkey/nodesource.gpg.key | sudo apt-key add -
         if [[ "$DEB_ID" == "Debian" ]] || [[ "$DEB_ID" == "Ubuntu" ]]; then
             (echo "deb https://deb.nodesource.com/node_12.x $(lsb_release -s -c) main" | sudo tee /etc/apt/sources.list.d/nodesource.list)
-        elif [[ "$DEB_ID" != "Debian" ]] || [[ "$DEB_ID" != "Ubuntu" ]]; then
+        else
             (echo "deb https://deb.nodesource.com/node_12.x $(lsb_release -s -c -u) main" | sudo tee /etc/apt/sources.list.d/nodesource.list)
     	fi
     sudo apt-get update
--- a/install.sh
+++ b/install.sh
@@ -114,7 +114,7 @@ if [[ ! -z $DEB ]]; then
             (echo "deb https://deb.nodesource.com/node_12.x $(lsb_release -s -c) main" | sudo tee /etc/apt/sources.list.d/nodesource.list)
         else
             (echo "deb https://deb.nodesource.com/node_12.x $(lsb_release -s -c -u) main" | sudo tee /etc/apt/sources.list.d/nodesource.list)
-    	fi
+        fi
     sudo apt-get update
     sudo apt-get install nodejs -y
 
--- a/install.sh
+++ b/install.sh
@@ -47,6 +47,9 @@ RPM=$(which yum || :)
 # Detect SystemV / SystemD
 SYS=$([[ -L "/sbin/init" ]] && echo 'SystemD' || echo 'SystemV')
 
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -107,9 +110,14 @@ sudo rm -rf ~/{.npm,.forever,.node*,.cache,.nvm}
 
 if [[ ! -z $DEB ]]; then
     sudo wget --quiet -O - https://deb.nodesource.com/gpgkey/nodesource.gpg.key | sudo apt-key add -
-    (echo "deb https://deb.nodesource.com/node_12.x $(lsb_release -s -c) main" | sudo tee /etc/apt/sources.list.d/nodesource.list)
+        if [[ "$DEB_ID" == "Debian" ]] || [[ "$DEB_ID" == "Ubuntu" ]]; then
+            (echo "deb https://deb.nodesource.com/node_12.x $(lsb_release -s -c) main" | sudo tee /etc/apt/sources.list.d/nodesource.list)
+        elif [[ "$DEB_ID" != "Debian" ]] || [[ "$DEB_ID" != "Ubuntu" ]]; then
```
deanpress: Should be `&&`, not `||` here I think?

adrian69: > Should be `&&`, not `||` on line 115 I think?

Did you test? Should be fine with OR as well 🤔

rainydio: Why do you have `elif` at all instead of `else`?

adrian69: > That condition is always true. Why do you have `elif` at all instead of `else`?

Correct, as long as it's DEB package all conditions are covered, so `else` is just fine. Changed.

rainydio: Generally speaking any condition of that kind (`x != val1 || x != val2`) is true. Usually `&&` is intended instead `||` (that's what @deanpress meant).

deanpress:
@adrian69 This doesn't work on my Linux Mint:
```
Hit:9 http://packages.linuxmint.com tina Release                                     
Fetched 252 kB in 1s (431 kB/s)                    
Reading package lists... Done
Reading package lists... Done
Building dependency tree       
Reading state information... Done
curl is already the newest version (7.58.0-2ubuntu3.8).
git is already the newest version (1:2.17.1-1ubuntu0.4).
update-notifier is already the newest version (3.192.1.7).
apt-transport-https is already the newest version (1.6.12).
0 upgraded, 0 newly installed, 0 to remove and 0 not upgraded.
    ==> Installed system dependencies!
    ==> Installing node.js & npm...
OK
Usage: lsb_release [options]

lsb_release: error: no such option: -u
deb https://deb.nodesource.com/node_12.x  main
E: Malformed entry 1 in list file /etc/apt/sources.list.d/nodesource.list (Component)
E: The list of sources could not be read.
```

deanpress:
The above is issue is semi fixed by replacing line 113 with:

```
        if [[ "$DEB_ID" == "Debian" ]] || [[ "$DEB_ID" == "Ubuntu" ]] || [[ "$DEB_ID" == "LinuxMint" ]]; then
```

The issue is that (with Linux Mint 19.2), we're back at issue #3261

deanpress:
Looks like this is an issue specifically for Linux Mint tina: https://github.com/nodesource/distributions/pull/950/files

PR #3273, opened 2019-11-18
chore: collapse duplicate logic

ciband:
## Summary
Reduced two if blocks to a single if block since their conditions
were exactly the same and they were sequencial statements.

## Checklist

- [ ] Tests _(if necessary)_
- [ ] Ready to be merged


supaiku0:
This breaks the entire signature deserialization. We need to perform this check after each read, since otherwise, whenever there is no second signature, it will attempt to read non-existing or wrong bytes.

PR #3278, opened 2019-11-19
refactor(crypto): set minimum fee on transaction types

dated:
## Summary

Overwrites the new minimum fee of zero introduced in #3275 to be positive again on certain transaction types:

- transfer
- second signature
- delegate registration
- delegate resignation
- vote
- multi signature
- ipfs
- htlc lock
- multi payment

Alternatively the unit tests for these transaction types can be adjusted to check for negative fees instead of zero fees.

## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged


dated:
The `bypassGenesis: true` on transfers and delegate registrations can probably be removed, since the properties are deepmerged anyway.

PR #3282, opened 2019-11-19
chore: Add "LinuxMint" to nodesource $DEB_ID check

deanpress:
## Summary

Linux Mint had issues with the nodesource package source being added in `install.sh`. This fixes it.

## Checklist

- [ ] Tests _(if necessary)_
- [x] Ready to be merged


deanpress:
Closed after discussing with @adrian69. We're looking for a better solution that also fixes Mint 19 support.

PR #3284, opened 2019-11-19
fix(core-magistrate): use correct size for BridgechainUpdate asset buffer

supaiku0:
## Checklist

- [ ] Tests _(if necessary)_
- [ ] Ready to be merged


dated:
Fwiw I had fixed that in #3255 already

PR #3311, opened 2019-11-29
feat(core-p2p): increase maximum token symbol length

Nigui:
## Summary

In my bridgechain I have a token symbol with length 4 and current p2p communication schema breaks peers exchanges.

Reasons why I have a 4 characters symbol is : 
- I don't have unicode representing my token (like Ѧ does)
- It's on a test network (devnet, testnet, unitnet,...) and I've just prefixed my token trigram with first letter of network (similarily to DѦ, TѦ and UѦ).

I don't think I'm and I'll be the only bridgechain facing this issue.

My solution is to increment symbol length by 1 only to fit with my current needs. 

Maybe a better solution would be to remove that check ?

## Checklist

- [X] Documentation _(if necessary)_
- [X] Tests _(if necessary)_
- [X] Ready to be merged


fix:
good point!, i think we need to have a db migration for that, but nothing unachievable

faustbrian:
@Nigui we can increase this to something reasonable like a maximum of 8 characters to allow some longer strings but removing this check completely would mean that you could send any arbitrary data of any length in that specific field.

Don't understand how we need a database migration for a validation schema so just ignore that comment from @fix.

fix:
oh yes, i thought there were some database table with the data. so yes no database schema migration then :)

PR #3312, opened 2019-11-29
refactor(core-api): wallet transform

supaiku0:
## Summary
Include all wallet attributes for `/wallets` responses and change the business endpoints to only return wallets when `transform=false` is passed.

## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged


air1one:
Just for info I'm reviewing and fixing the tests so that we can merge soon :ok_hand:

PR #3314, opened 2019-11-30
[WIP] refactor(core-api): stricter validation of orderBy iteratees

### LINKED ISSUE #3315: [2.6][T4] orderBy not working with some camel cased iteratees
The `orderBy` query param has no effect on the returned results by the API with some iteratees that need to be camel cased, such as `forgedTotal`, caused by calling `toLowerCase` on the iteratee on one side (e.g. `forgedTotal` -> `forgedtotal`)
https://github.com/ArkEcosystem/core/blob/1f405267711e6fa1903486300eec6ef4160a4f7a/packages/core-database/src/repositories/utils/search-entries.ts#L47 but failing to add respective handling of the lowercased value on the other (no case for e.g. `forgedtotal`)

https://github.com/ArkEcosystem/core/blob/1f405267711e6fa1903486300eec6ef4160a4f7a/packages/core-database/src/repositories/utils/search-entries.ts#L56-L72

The current approach needs one case in the switch statement for each possible camel cased property. Another downside is that invalid iteratees might go unnoticed and yield unexpected results.

This could be solved by adding a global whitelist of accepted iteratees to be used in conjunction with the orderBy query param. This could even be improved on further by specifying a whitelist on a resource type level.

dated:
## Summary

Closes #3315 by providing a global whitelist of accepted iteratees with a verbose error message if the validation fails:

```
{
  "statusCode": 422,
  "error": "Unprocessable Entity",
  "message": "child \"orderBy\" fails because [\"orderBy\" must have the following format <iteratee>:<direction>, where <iteratee> is one of: address, amount, approval, balance, blockId, confirmations, expirationType, expirationValue, fee, forgedFees, forgedRewards, forgedTotal, generatorPublicKey, height, id, isDelegate, isResigned, latency, lockedBalance, nonce, producedBlocks, publicKey, rank, recipient, sender, senderPublicKey, signature, signSignature, timestamp, transactions, type, typeGroup, username, vendorField, version, voteBalance, votes and <direction> one of: asc, desc]"
}
```

## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged

### DIFF
--- a/packages/core-database/src/repositories/utils/search-parameter-converter.ts
+++ b/packages/core-database/src/repositories/utils/search-parameter-converter.ts
@@ -42,14 +42,15 @@ export class SearchParameterConverter implements Database.IISearchParameterConve
     }
 
     private parseOrderBy(searchParameters: Database.ISearchParameters, orderBy?: any) {
-        if (orderBy && typeof orderBy === "string") {
-            const fieldDirection = orderBy.split(":").map(o => o.toLowerCase());
-            if (fieldDirection.length === 2 && (fieldDirection[1] === "asc" || fieldDirection[1] === "desc")) {
-                searchParameters.orderBy.push({
-                    field: snakeCase(fieldDirection[0]),
-                    direction: fieldDirection[1] as "asc" | "desc",
-                });
-            }
+        if (
+            Array.isArray(orderBy) &&
+            orderBy.length === 2 &&
+            (orderBy[1] === "asc" || orderBy[1] === "desc")
+        ) {
+            searchParameters.orderBy.push({
+                field: orderBy[0],
+                direction: orderBy[1] as "asc" | "desc",
+            });
         }
     }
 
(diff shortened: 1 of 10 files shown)


packages/core-database/src/repositories/utils/search-parameter-converter.ts
```diff
@@ -42,14 +41,15 @@ export class SearchParameterConverter implements Database.IISearchParameterConve
     }
 
     private parseOrderBy(searchParameters: Database.ISearchParameters, orderBy?: any) {
-        if (orderBy && typeof orderBy === "string") {
-            const fieldDirection = orderBy.split(":").map(o => o.toLowerCase());
-            if (fieldDirection.length === 2 && (fieldDirection[1] === "asc" || fieldDirection[1] === "desc")) {
-                searchParameters.orderBy.push({
-                    field: snakeCase(fieldDirection[0]),
```
dated: This actually might have to stay :thinking: but currently it has no effect since the value is lowercased above

dated:
This had some unexpected results, will start over :lala:

Issue #3324, opened 2019-12-04
Getting Unprocessable Entity when calling transactions/unconfirmed


roccomuso:
**Expected Behavior**

Get the unconfirmed tx or something that tells me that tx is confirmed instead.
 
**Current Behavior**

Getting error:

```
{"statusCode":422,"error":"Unprocessable Entity","message":"child \"id\" fails because [\"id\" must only contain hexadecimal characters]"}
```

**Steps to Reproduce**

> curl -X GET "https://ark-server.io/api/transactions/unconfirmed/get?id=5f2936a50f283cd6f05dea73c25c10e1172a4d172e6f8855752330341413e1d5" -H "API-Version: 2"

ArkEcosystemBot:
**Thanks for opening this issue!** A maintainer will review this in the next few days and explicitly select labels so you know what's going on.

_If no reviewer appears after a week, a reminder will be sent out._

kristjank:
You can check that trx is confirmed by querying the api/transaction/{id} endpoint. The returned json has confirmation field. If field > 1 then is is confirmed. Does this help @roccomuso 

Also your call should be like this 
`https://ark.exodus-prod.io/api/transactions/unconfirmed/5f2936a50f283cd6f05dea73c25c10e1172a4d172e6f8855752330341413e1d5`

roccomuso:
Got it.
Thanks that help

ArkEcosystemBot:
This issue has been closed. If you wish to re-open it please provide additional information.

vasild:
Indeed according to the docs the usage is wrong: https://api.ark.dev/public-rest-api/endpoints/transactions#endpoint-4

However the returned error message is extremely confusing. The reason for this is that in the mistaken url: `.../unconfirmed/get?id=5f2936a50f...` the part that is interpreted as `id` is everything that follows the last `/` and before `?` - that is just `get` and it is not only hex chars.

I have opened a draft PR to include the input in the error message in the hopes to make it more clear: https://github.com/ArkEcosystem/core/pull/3326, I have been confused myself in other occasions by those validation messages that lack the "wrong" input.

PR #3326, opened 2019-12-05
feat(core-api): include the input on schema error

vasild:
Before this patch we would get error messages that did not include the
input which made it hard to guess what went wrong. For example, the
following error:

  child "id" fails because ["id" must only contain hexadecimal characters]

Returned from the API call .../unconfirmed/get?id=5f2936a50f283 is
puzzling because the id is indeed only in hexadecimal characters. An
example of such confusion: https://github.com/ArkEcosystem/core/issues/3324

After this patch:

.../unconfirmed/get?id=5f2936a50f283 ->
  "id" must only contain hexadecimal characters: key=id, value=get

.../unconfirmed/foo?bar=123 ->
  "id" must only contain hexadecimal characters: key=id, value=foo

### DIFF
--- a/packages/core-api/src/handlers/transactions/schema.ts
+++ b/packages/core-api/src/handlers/transactions/schema.ts
@@ -74,13 +74,45 @@ export const unconfirmed: object = {
     },
 };
 
-export const showUnconfirmed: object = {
+/**
+ * Create a function that would validate a given JOI object and if validation
+ * fails, then throw an error that includes the input value.
+ * See https://hapi.dev/api/?v=18.4.0#-routeoptionsvalidateparams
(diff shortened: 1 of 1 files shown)


vasild:
Any feedback on this PR? Good? Trash it? There is an easier way to achieve that? Should we do the same for all endpoints?

supaiku0:
This should be implemented as a Hapi plugin/failAction and handle everything automagically for all endpoints. The way it is now is too impractical and shouldn't require manual setup.

Anyway, I don't think it makes much sense to spend time on this for 2.6, given that V3 is quite different and a lot of it would probably end up getting rewritten.

PR #3333, opened 2019-12-09
test(e2e): simplify e2e init & network launch

air1one:
## Summary

Improve / simplify e2e initialization & network launch.

Before there was a quite complex setup for running the e2e network (using docker swarm). With this update it is just needed to have docker and docker-compose.

Also now you don't need to generate folders and copy core for each node, there is no initialization step at all.

Basically once core is built, you can go to `__tests__/e2e/lib/config` and run `docker-compose up -d`, that's it the network is launched with 5 nodes and you can see the output of each node with `docker logs config_core0_1` (check `docker ps` to have the name of each service running core).

## Checklist

- [ ] Tests _(if necessary)_
- [x] Ready to be merged


air1one:
Reverted a commit because it's a fix that has nothing to do with this e2e update, pushed it separately here #3336

PR #3336, opened 2019-12-10
fix(core-state): use parent findByPublicKey method for temp-wallet-manager

air1one:
## Summary

Use parent findByPublicKey method for temp-wallet-manager, instead of overriding it which is not of any use (please correct me if there is any use for it).

The override was causing issues with nonce when we had in the pool multiple transactions from the same sender. In this case calling `reindex` would push a clone of the wallet to database wallet manager, while returning the actual wallet. Then the actual wallet would have his nonce incremented (applying the transaction) but not the database wallet, causing a nonce error for the following transactions from the same sender. (this happens when getting transactions for forging)
This issue would only show for a "new" sender, say we have a 1st transaction from existing wallet A to a new wallet B, then B would only be indexed by address in database wallet manager : which is fine but was not handled correctly with previous temp-wallet-manager implementation.

## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged


air1one:
I'll try to add a test to reflect the issue that I was having. (that is described in PR description)

supaiku0:
Good catch!

PR #3379, opened 2020-01-08
feat(core-api): retrieve business by username, address or public key

### LINKED ISSUE #3377: regression: retrieve business by address, public key or username
The features introduced in #3310 seem to no longer be available, i.e. retrieving a business through the `businesses/{id}` endpoint is only possible by using the public key as identifier.

Original issue: #3308

dated:
## Summary

Fixes #3377.

## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged

### DIFF
--- a/__tests__/integration/core-api/handlers/businesses.test.ts
+++ b/__tests__/integration/core-api/handlers/businesses.test.ts
@@ -34,6 +34,12 @@ describe("API 2.0 - Businesses", () => {
         },
     };
 
+    const validIdentifiers = {
+        username,
+        address,
+        publicKey,
+    };
+
     beforeAll(() => {
         walletManager = app.resolvePlugin<Database.IDatabaseService>("database").walletManager;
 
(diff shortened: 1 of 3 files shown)


__tests__/integration/core-api/handlers/businesses.test.ts
```diff
@@ -60,10 +67,12 @@ describe("API 2.0 - Businesses", () => {
 
     describe("GET /businesses/:id", () => {
         it("should GET a business by the given valid identifier", async () => {
-            const response = await utils.request("GET", `businesses/${publicKey}`, { transform: false });
-            expect(response).toBeSuccessfulResponse();
-
-            expect(response.data.data.attributes.business).toEqual(businessAttribute);
+            for (const identifier of Object.values(validIdentifiers)) {
```
air1one: Can you instead of this loop, use `it.each`, you could do something like
```
it.each(Object.entries(validIdentifiers))("should GET a business by %s : %s", async (idName, idValue) => {
```
it then shows up better in the test recap (if one of address, public key or username fails you will see directly which one fails)

__tests__/integration/core-api/handlers/businesses.test.ts
```diff
@@ -73,12 +82,14 @@ describe("API 2.0 - Businesses", () => {
 
     describe("GET /businesses/:id/bridgechains", () => {
         it("should GET business bridgechains", async () => {
-            const response = await utils.request("GET", `businesses/${publicKey}/bridgechains`);
-            expect(response).toBeSuccessfulResponse();
-            expect(response.data.data).toBeArray();
-            expect(response.data.data).toHaveLength(1);
+            for (const identifier of Object.values(validIdentifiers)) {
```
air1one: see the other comment, if you can replace by `it.each`

packages/core-api/src/handlers/businesses/methods.ts
```diff
@@ -16,8 +16,20 @@ const index = async request => {
 };
 
 const show = async request => {
+    let publicKey: string;
+
+    if (request.params.id.length <= 34) {
```
air1one: I'm picky :smile: but can you do instead 
```
if (request.params.id.length === 34) {
    publicKey = request.params.id;
```
and then `else` you retrieve the public key from database service.
Would look a bit clearer to me.

dated: Sure, I'll turn it around, but using `66` :D

PR #3380, opened 2020-01-09
fix(core-api): restore search business by address / username

### LINKED ISSUE #3377: regression: retrieve business by address, public key or username

air1one:
## Summary

Resolves #3377 

Restore business search by address and username that we lost in recent commit.

## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged


air1one:
should have first checked that open PR that already fixes it :see_no_evil:

PR #3437, opened 2020-02-05
refactor(core-api): re-enable username and 2nd public key as top api attributes

air1one:
## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged


dated:
Certain properties (username, vote, 2nd signature, multi signature, ...) are returned twice for each wallets, at the top level and as nested attributes. Are there any plans to straighten that out in the future? Should the existing applications be adjusted to retrieve the nested properties, assuming the top level properties will be removed again?

air1one:
We decided to keep those top-level properties with 2.6 but this will be totally inside nested `attributes` with v3.
So yes better retrieve the nested properties than the top-level ones that will be removed with v3 @dated

dated:
Perfect, thank you 👌

PR #3499, opened 2020-02-13
refactor(core-transaction-pool): segregate pools by wallets

rainydio:
## Summary

Each sender gets his own pool state and cannot interfere with others.

Requires #3536 

## Checklist

- [ ] Documentation
- [ ] Ready to be merged


faustbrian:
Will go over this on monday.

faustbrian:
@rainydio merge conflicts

packages/core-transaction-pool/src/query.ts
```diff
@@ -0,0 +1,156 @@
...
+    private readonly iter: Iterator<Interfaces.ITransaction>;
+    private readonly predicate: Contracts.TransactionPool.Predicate | undefined;
+
+    public constructor(iter: Iterator<Interfaces.ITransaction>, predicate?: Contracts.TransactionPool.Predicate) {
+        this.iter = iter;
+        this.predicate = predicate;
+    }
+
+    public next(value?: any): IteratorResult<Interfaces.ITransaction> {
+        for (let r = this.iter.next(); !r.done; r = this.iter.next()) {
```
faustbrian: Avoid single character variable assignments unless it is a `for` loop with numerical increments and also complex loop conditions like this as they are harder to read and make sense of than a `while` or `for (const x of ...)`.

Could this be replaced with a `while` that loops until it has a match or `next()` no longer performs anything?

packages/core-transaction-pool/src/query.ts
```diff
@@ -0,0 +1,156 @@
...
+            if (!this.predicate || this.predicate(r.value)) {
+                return r;
+            }
+        }
+
+        return { done: true, value: undefined };
```
faustbrian: We should avoid indicators like `done/success/failure` based on boolean values. It should be possible to determine the success or failure of a method solely based on its return value without any additional indicators.

packages/core-transaction-pool/src/query.ts
```diff
@@ -0,0 +1,156 @@
...
+                t.data.version === transaction.data.version,
+        );
+    }
+
+    public has(): boolean {
+        return this[Symbol.iterator]().next().done === false;
+    }
+
+    public first(): Interfaces.ITransaction {
+        const r = this[Symbol.iterator]().next();
```
faustbrian: Avoid single character variable assignments unless they are used in a `for` loop for numerical increments. A brief name that makes it clear what `r` is saves another developer the time to figure out what is returned by `next()`.

rainydio:
> Always add types to **variables**, functions and return values.

Modern languages usually choose inferred types on local variables. Go and Rust were designed this way. C# added `var` keyword more than 10 years ago. Even conservative Java added `var` keyword couple of years ago too.

> I know that TypeScript and an IDE can just show you all types... If someone is using vim, emacs or some barebones editor like sublime text they will need to dig through the code

This argument of IDE vs plain editor is ages old. I think in this particular case the shift occurred not due to IDEs getting better.

Explicit types hurt readability. They clutter code with information that is only needed if you try to edit it without IDE. For example documentation example snippets (any language) which by definition strike for highest readability use inferred local variable types.

Explicit types hurt code review. Around 5 to 10 years ago online code review tools got popular. The way people read code review diffs is very similar to documentation example snippets. The meaning is in variable and method names, and how they interact. It's difficult to review big chunks of code cluttered with information that's only needed to edit code without IDE.

PS: Way more effort is needed to lay down convincing arguments than to add type information.

---

Regarding one character variable names, you are probably correct. I have rule of thumb that I'm likely overusing:

```typescript
const coreTransactionIds = transactions
  .filter(t => t.typeGroup === Enums.TransactionTypeGroup.Core)
  .map(t => t.id)
```

When arrow function contains single expression it's okay to use single character variable name. But this only work well with arrays, when it's obvious what is iterated. I'm gonna fix that.

---

Regarding errors, you are probably correct. I used the names that are similar to type field that was already there (e.g. `ERR_POOL_FULL` translated to `PoolFullError`). This doesn't have to be that way or type field can be changed too. Next PR will focus on errors and processor and we can discuss this problem there.

packages/core-state/src/wallets/wallet-repository-cow.ts
```diff
@@ -1,9 +1,10 @@
 import { Container, Contracts } from "@arkecosystem/core-kernel";
-import { Wallets } from "@arkecosystem/core-state";
 import { Identities } from "@arkecosystem/crypto";
 
+import { WalletRepository } from "./wallet-repository";
+
 @Container.injectable()
-export class PoolWalletRepository extends Wallets.WalletRepository {
+export class WalletRepositoryCow extends WalletRepository {
```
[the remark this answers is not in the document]
rainydio: I changed that to WalletRepositoryCopyOnWrite

packages/core-transaction-pool/src/query.ts
rainydio: Replaced it with while loop

packages/core-transaction-pool/src/query.ts
rainydio: This is ES iterator interface, not mine. This is what `for (const value of iterable)` expects.

packages/core-transaction-pool/src/query.ts
rainydio: Renamed that into `result`

packages/core-p2p/src/socket-server/versions/internal.ts
```diff
@@ -54,11 +54,9 @@ export const getUnconfirmedTransactions = async ({
     app: Contracts.Kernel.Application;
 }): Promise<Contracts.P2P.UnconfirmedTransactions> => {
     const collator = app.get<Contracts.TransactionPool.Collator>(Container.Identifiers.TransactionPoolCollator);
-    const transactionPool: Contracts.TransactionPool.Connection = app.get<Contracts.TransactionPool.Connection>(
-        Container.Identifiers.TransactionPoolService,
-    );
+    const transactionPool = app.get<Contracts.TransactionPool.Service>(Container.Identifiers.TransactionPoolService);
     const transactions = await collator.getBlockCandidateTransactions();
-    const poolSize = await transactionPool.getPoolSize();
+    const poolSize = await transactionPool.size;
```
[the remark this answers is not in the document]
rainydio: Changed that to method

packages/core-kernel/src/contracts/transaction-pool/service.ts
```diff
@@ -0,0 +1,15 @@
+import { Interfaces } from "@arkecosystem/crypto";
+
+export interface Service {
+    size: number;
+
+    boot(): Promise<void>;
+    clear(): void;
```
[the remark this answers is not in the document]
rainydio: Standard ES collection classes (Set and Map) use clear method, instead of flush. Rename it into flush?

packages/core-kernel/src/contracts/transaction-pool/query.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export interface Query {
+    all(): QueryIterable;
+    allFromLowestPriority(): QueryIterable;
+    allFromHighestPriority(): QueryIterable;
+    allFromSender(senderPublicKey: string): QueryIterable;
+}
+
+export interface QueryIterable extends Iterable<Interfaces.ITransaction> {
+    whenPredicate(predicate: Predicate): QueryIterable;
```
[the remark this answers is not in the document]
rainydio: Renamed

packages/core-kernel/src/contracts/transaction-pool/query.ts
```diff
@@ -0,0 +1,23 @@
...
+    whenId(id: string): QueryIterable;
```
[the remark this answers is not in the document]
rainydio: Renamed

packages/core-kernel/src/contracts/transaction-pool/query.ts
```diff
@@ -0,0 +1,23 @@
...
+    whenType(type: number): QueryIterable;
```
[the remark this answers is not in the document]
rainydio: Renamed

packages/core-kernel/src/contracts/transaction-pool/query.ts
```diff
@@ -0,0 +1,23 @@
...
+    whenTypeGroup(typeGroup: number): QueryIterable;
```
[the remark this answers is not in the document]
rainydio: Renamed

packages/core-kernel/src/contracts/transaction-pool/query.ts
```diff
@@ -0,0 +1,23 @@
...
+    whenVersion(version: number): QueryIterable;
```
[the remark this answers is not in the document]
rainydio: Renamed

packages/core-kernel/src/contracts/transaction-pool/query.ts
```diff
@@ -0,0 +1,23 @@
...
+    whenInternalType(internalType: Transactions.InternalTransactionType): QueryIterable;
```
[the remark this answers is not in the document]
rainydio: Renamed

packages/core-kernel/src/contracts/transaction-pool/query.ts
```diff
@@ -0,0 +1,23 @@
...
+    whenKind(transaction: Interfaces.ITransaction): QueryIterable;
```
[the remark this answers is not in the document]
rainydio: Renamed

packages/core-transaction-pool/src/query.ts
rainydio: Replaced it with generator and for..of loop

packages/core-transaction-pool/src/query.ts
rainydio: It is gone now

packages/core-kernel/src/contracts/transaction-pool/service.ts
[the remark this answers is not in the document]
faustbrian: Yep, all interfaces of our own packages use `flush` as name.

packages/core-transactions/src/handlers/two/delegate-resignation.ts
```diff
@@ -74,30 +82,12 @@ export class DelegateResignationTransactionHandler extends TransactionHandler {
...
-        }
-
-        return true;
-    }
-
     public async applyToSender(
         transaction: Interfaces.ITransaction,
         customWalletRepository?: Contracts.State.WalletRepository,
     ): Promise<void> {
+        this.app.log.notice(`${this.walletRepository.allByUsername().length} delegates before resignation`);
```
faustbrian: Inject the logger instead of using `this.app.log`. The `log` getter will be removed soon, it was only added initially for migration purposes.

packages/core-transaction-pool/src/service-provider.ts
```diff
@@ -18,42 +21,39 @@ export class ServiceProvider extends Providers.ServiceProvider {
...
     public async register(): Promise<void> {
-        this.app
-            .bind(Container.Identifiers.WalletRepository)
-            .to(PoolWalletRepository)
-            .inSingletonScope()
-            .when(Container.Selectors.anyAncestorOrTargetTaggedFirst("state", "pool"));
-
+        this.app.bind(Container.Identifiers.TransactionPoolCollator).to(Collator);
+        this.app.bind(DynamicFeeMatcher).toSelf();
+        this.app.bind(ExpirationService).toSelf();
```
faustbrian: We should avoid self-binding classes and always bind to a symbol/name because this makes it difficult for a plugin developers to replace this binding.

packages/core-transaction-pool/src/service-provider.ts
```diff
@@ -18,42 +21,39 @@ export class ServiceProvider extends Providers.ServiceProvider {
...
      */
     public async register(): Promise<void> {
-        this.app
-            .bind(Container.Identifiers.WalletRepository)
-            .to(PoolWalletRepository)
-            .inSingletonScope()
-            .when(Container.Selectors.anyAncestorOrTargetTaggedFirst("state", "pool"));
-
+        this.app.bind(Container.Identifiers.TransactionPoolCollator).to(Collator);
+        this.app.bind(DynamicFeeMatcher).toSelf();
```
faustbrian: We should avoid self-binding classes and always bind to a symbol/name because this makes it difficult for a plugin developers to replace this binding.

packages/core-transaction-pool/src/errors.ts
```diff
@@ -0,0 +1,95 @@
+import { Contracts } from "@arkecosystem/core-kernel";
+import { Interfaces, Utils } from "@arkecosystem/crypto";
+
+export class DuplicateError extends Error implements Contracts.TransactionPool.DuplicateError {
```
faustbrian: Is there a specific reason for needing interfaces on error classes? They all simply implement the `constructor` method for all they need to know.

packages/core-transaction-pool/src/processor.ts
```diff
@@ -1,436 +1,57 @@
...
-        const stats: string = ["accept", "broadcast", "excess", "invalid"]
-            .map(prop => `${prop}: ${this[prop] instanceof Array ? this[prop].length : this[prop].size}`)
-            .join(" ");
-
-        if (Object.keys(this.errors).length > 0) {
-            this.logger.debug(JSON.stringify(this.errors));
         }
 
-        this.logger.info(`Received ${Utils.pluralize("transaction", this.transactions.length, true)} (${stats}).`);
+        await this.networkMonitor.broadcastTransactions(broadcastable);
```
faustbrian: This should be moved out so that the sole responsibility of the processor is to decide the state of a transaction (accept, broadcast, excess, invalid, error). The broadcasting is a separate task, the processor should only provide the data that is needed for it.

faustbrian:
> Modern languages usually choose inferred types on local variables. Go and Rust were designed this way. C# added `var` keyword more than 10 years ago. Even conservative Java added `var` keyword couple of years ago too.

> This argument of IDE vs plain editor is ages old. I think in this particular case the shift occurred not due to IDEs getting better.

The argument is ages old and still sticks around for a reason. When you jump around dozens of companies you'll meet as many people that don't use IDEs as do use IDEs because they prefer a more minimal setup without all the clutter and noise that modern IDEs add to your development environment.

> Explicit types hurt readability. They clutter code with information that is only needed if you try to edit it without IDE. For example documentation example snippets (any language) which by definition strike for highest readability use inferred local variable types.

> Explicit types hurt code review. Around 5 to 10 years ago online code review tools got popular. The way people read code review diffs is very similar to documentation example snippets. The meaning is in variable and method names, and how they interact. It's difficult to review big chunks of code cluttered with information that's only needed to edit code without IDE.

> I'm concerned that explicit local variable types are bad for code. Way more effort is needed to lay down convincing arguments than to add type information.

Proper naming is obviously key but I would argue the exact opposite in terms of readability and code review problems. If I am required to use an IDE with plugins or dig through files unrelated directly to the changes of a PR to make out what type a variable has or what is returned by a function it is more distracting to me than having the code itself tell me about it.

Could argue that is personal preference but I would say that in general it makes it easier for new people to jump into the code as everything is right at your hand without having to dig deeper to see what class X from 5 dependencies down is returning because the variable doesn't tell me.

> Regarding errors, you are probably correct. I used the names that are similar to type field that was already there (e.g. `ERR_POOL_FULL` translated to `PoolFullError`). This doesn't have to be that way or type field can be changed too. Next PR will focus on errors and processor and we can discuss this problem there.

If you will create another PR that reworks them right after this is merged it might be better to do them completely in a separate a PR instead of merging something that gets touched up again right after being merged.

packages/core-transactions/src/handlers/two/delegate-resignation.ts
rainydio: Let me remove that, I left it there by accident while debugging

packages/core-transaction-pool/src/errors.ts
rainydio: No reason. I'm removing them.

packages/core-transaction-pool/src/processor.ts
rainydio: I disagree. Why i can't add to pool and broadcast instead of requiring caller to do that?

rainydio:
> If you will create another PR that reworks them right after this is merged it might be better to do them completely in a separate a PR instead of merging something that gets touched up again right after being merged.

Merged before (similarly to #3500 and #3501). It's a bit difficult to cut only errors and processor, but let me try.

faustbrian:
@rainydio conflicts

packages/core-transaction-pool/src/utils.ts
```diff
@@ -0,0 +1,80 @@
...
+            this.results.delete(iterator);
+        } else {
+            this.results.set(iterator, nextResult);
+        }
+
+        return result;
+    }
+}
+
+export const describeTransactionType = (transaction: Interfaces.ITransaction): string => {
```
faustbrian: Could we make this function more generic so it works for transactions that are developed by third-parties? Right now it would simply return `unknown` if the transaction isn't an official one that we maintain.

packages/core-transaction-pool/src/processor.ts
faustbrian: The reasoning is both responsibility and testability. The whole function right now is coupled to the network monitor from the P2P package just for broadcasting which means during tests you have to mock the implementation or fake a response.

99% of this function is attached to the pool, about deciding if transactions are valid and adding them to the pool and then there is this 1 unrelated action in the shape of the P2P call. The transaction pool itself should be unaware of the existence of the P2P layer, it only cares and knows about what transactions are and if they are valid. What the caller wants to do with the transactions after it has handled them is not of any concern for it.

packages/core-transaction-pool/src/utils.ts
rainydio: Yes, by adding describe method to transaction handler. But in a separate PR :)

faustbrian: Please make that PR first so we don't need to merge something that gets removed in a follow up PR.

packages/core-transaction-pool/src/processor.ts
rainydio: This is actually the reason I think processor should not be part of pool. But more general service on top of pool and p2p (or other broadcast service).

If instead instructions are returned back (previous processor), then someone still has to execute them. There are currently two places that do that, and they did that differently before #3514. P2P was incorrectly(?) refusing the whole batch if single transaction failed to apply, but API broadcasted whatever applied. 

It's _very_ easy to make mistake and accidently do the wrong thing. Processor that was implemented in #3514 isn't simple and probably has errors, but everything is contained within it.

faustbrian: The rejection of everything via P2P even if just 1 transaction is invalid is on purpose as a precaution to prevent someone from sending junk directly to a node via P2P and it causing problems. P2P in general has some different behaviours than the API because the API only cares about taking any data and broadcasting while P2P double checks that everything is 100% valid.

faustbrian:
@rainydio conflicts. After the merge the PR should be good to merge.

faustbrian:
@rainydio let me know when this is ready for review so we can merge this because we soon also need to merge 2.6 into 3.0

rainydio:
Regarding `describeTransaction` method, I created #3529 and will implement it later. So this PR isn't delayed futher. Otherwise it's read for review @faustbrian.

faustbrian:
Looks good, will merge this once @air1one finished the 3.0 merge which he started yesterday.

rainydio:
Because it's delayed I also updated log lines and created separate PR #3536 that implements transaction `toString` method.

PR #3504, opened 2020-02-14
refactor(core-magistrate): allow multiple ports in bridgechain schema

air1one:
## Summary

Allow to register multiple ports in bridgechain registration / update, instead of only core-api.

## Checklist

- [ ] Tests _(if necessary)_
- [ ] Ready to be merged


air1one:
Still need to make the "pattern properties" of the ports schema more specific, only including chars that could actually be in a package name.

PR #3516, opened 2020-02-20
refactor(core-p2p): transaction broadcaster

rainydio:
## Summary

Network monitor `broadcastTransactions` is extracted into separate `TransactionBroadcaster` class that contains a single function. It's injected into processor as *optional* dependency, so processor could function without p2p plugin.

## Checklist

- [ ] Documentation
- [x] Tests
- [x] Ready to be merged

Continuing argument raised in #3499. From composition perspective both processor and dynamic fee matcher are aware of transaction broadcasting. It makes more sense to call `broadcastTransactions` from processor than from api controller or p2p peer handler. Processor broadcasts whatever was successfully applied (if fee is high enough). Broadcaster dependency is small, abstract, and optional.

---

Distinct feature of unit test is a single import of test subject and mocks for everything else (that's why they are called unit). Currently tests located within `__tests__/unit` are not unit, those are actually functional tests.

~I added couple of incomplete unit tests and moved them into packages right next to file being tested. This technique is only applicable to unit tests and greatly simplifies the process of writing them.~

Test files are now linted.


packages/core-p2p/src/transaction-broadcaster.test.ts
```diff
@@ -0,0 +1,60 @@
+import { Container } from "@arkecosystem/core-kernel";
```
faustbrian: This file should go to `__tests__/unit/core-p2p/transaction-broadcaster.test.ts`

rainydio: Moved

PR #3533, opened 2020-02-25
test(core-transaction-pool): collator test

rainydio:
## Summary

Part of #3482. It's missing catch coverage, because I need to implement logging first. It's difficult to mock otherwise. Also I noticed that it makes more sense to check for transaction size first before trying to validate it.

## Checklist

- [x] Tests
- [x] Ready to be merged


rainydio:
Already merged in #3499

PR #3541, opened 2020-02-26
perf(core): use jemalloc as the memory allocator

alessiodf:
## Summary

This PR replaces the native memory allocator with jemalloc in Core to reduce memory fragmentation and results in a smaller memory footprint as confirmed by multiple Core nodes using the native memory allocator (glibc) vs jemalloc, where in all cases the jemalloc nodes consistently used less RAM. Furthermore, results from a sync from 0 test (using the same hardware with 2GB physical RAM to eliminate variability due to different specifications) demonstrate that jemalloc is more performant and uses far less memory by the end:

native memory allocator (glibc): Completed in 14h 54m 14s using 1.3GB RAM 30 minutes after completion.
jemalloc: Completed in 13h 37m 56s using 314.5MB RAM 30 minutes after completion.

Therefore a sync from 0 using jemalloc completed 1h 16m 18s faster than glibc on the same hardware and, upon finishing, was using ~1017MB less RAM.

On a different (much higher powered) machine with 32GB RAM, I rolled back 300,000 blocks on devnet and re-synced using glibc and jemalloc:

native memory allocator (glibc): Completed in 7 minutes and 20 seconds using 3.5GB RAM (!) 10 minutes after completion.
jemalloc: Completed in 6 minutes and 50 seconds using 410.9MB RAM 10 minutes after completion.

In that instance, jemalloc completed 30 seconds faster than glibc and used a staggering ~3173MB less RAM.

This PR does not change the default system memory allocator to jemalloc as this may not be desirable and we should not make such system-wide changes. It just makes the Core process use jemalloc and leaves everything else untouched.

**Special update instructions:**
Users should manually run `sudo apt-get install libjemalloc-dev` (Debian derivatives) or `sudo yum install libjemalloc-devel` (RedHat derivatives) before updating since they don't re-run the Core installer (which will automatically install it for new installations).

~Update with `ark update --no-restart` then `pm2 delete all` and restart the `relay`/`forger`/`core` process(es) via `ark {relay|forger|core}:start`. Merely executing `ark {relay|forger|core}:restart` or letting the updater restart the processes is insufficient as it will not inject the `LD_PRELOAD` variable, presumably due to pm2 caching.~ 
Now refactored so the normal `ark update` procedure is sufficient to restart Core as long as libjemalloc has been installed as above.

To confirm Core is now using jemalloc, run ``cat /proc/`pgrep -f "@arkecosystem/core/bin/run"`/smaps | grep jemalloc``. The output should show references to `libjemalloc.so` to confirm the process is running with jemalloc.

## Checklist

- [ ] Tests _(if necessary)_
- [x] Ready to be merged

### DIFF
--- a/packages/core/bin/run
+++ b/packages/core/bin/run
@@ -4,7 +4,7 @@ if (process.env.LD_PRELOAD !== 'libjemalloc.so') {
     process.env.LD_PRELOAD = 'libjemalloc.so'
     let exitStatus = 0
     try {
-        require('child_process').execSync(process.argv.join(' '), { stdio: [0, 1, 2] })
+        require('child_process').execFileSync(process.argv0, process.argv.slice(1), { stdio: [0, 1, 2] })
     } catch (error) {
         exitStatus = error.status
     }
--- a/packages/core/bin/run
+++ b/packages/core/bin/run
@@ -1,5 +1,7 @@
 #!/usr/bin/env node
 
+process.env.LD_PRELOAD = 'libjemalloc.so'
+
 require('@oclif/command').run()
 .then(require('@oclif/command/flush'))
 .catch(require('@oclif/errors/handle'))
--- a/packages/core/bin/run
+++ b/packages/core/bin/run
@@ -1,14 +1,22 @@
 #!/usr/bin/env node
 
-if (process.env.LD_PRELOAD !== 'libjemalloc.so') {
-    process.env.LD_PRELOAD = 'libjemalloc.so'
-    let exitStatus = 0
-    try {
-        require('child_process').execFileSync(process.argv0, process.argv.slice(1), { stdio: [0, 1, 2] })
-    } catch (error) {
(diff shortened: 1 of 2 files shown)


packages/core/bin/run
```diff
@@ -1,14 +1,22 @@
 #!/usr/bin/env node
 
-if (process.env.LD_PRELOAD !== 'libjemalloc.so') {
-    process.env.LD_PRELOAD = 'libjemalloc.so'
-    let exitStatus = 0
-    try {
-        require('child_process').execSync(process.argv.join(' '), { stdio: [0, 1, 2] })
-    } catch (error) {
-        exitStatus = error.status
+const child_process = require('child_process')
```
faustbrian: This whole block of new code should be wrapped into `try/catch` to not prevent core from running if any of it fails.

alessiodf: 917d5d4

PR #3543, opened 2020-02-27
test(core-state): increase coverage to 100%

bertiespell:
## Summary

Adding tests for the `core-state` package.

Some refactor work to support the tests.

I chose to reuse a `Sandbox` environment for each test, which sets up and injects every class needed. However, because of the way classes are tagged (and different classes need access to different tagged versions) this may create complications (see discussion below).

## Checklist

- [x] Documentation _(if necessary)_
- [x] Tests _(if necessary)_
- [ ] Ready to be merged

I've left some `TODO`s in the code to highlight areas where I was unsure what the intended behaviour was.

Most of these issues occur in `WalletRepository` and `TempWalletRepository` - since the latter inherits from the former.  Inheritance can be useful for small classes to extend some simple base functionality, but if the two classes are designed to behave differently or become complex - it can become quite difficult to maintain. One problem is that a method in the extended class can call into a method in the base class, which itself calls a method back in the extended class. This can make it harder to reason about the code, it has the potential to cause subtle bugs and makes the code more difficult to maintain. I note here too that there is a 3rd `PoolWalletRepository` which extends from this class and lives in a different package. It might be worth thinking about whether a shared common interface might be a better approach here?

I found that I lost a lot of time setting up tests because of issues with tagging. When two versions of a class (using the same identifier) are bound via tags - if this identifier is then consumed by a third class which does *not* specify a tag - which tagged class should it fall back to? Is this an intended consequence, or are we using this "fallback" as a feature? I'd suggest that code should be explicit in its intentions, rather than implicit. In the `ServiceProvider`, the `WalletRepository` and `TempWalletRepository` classes are bound to the same identifier via tags. However, `BlockState` consumes an *untagged* version of this identifier - what should this default to - or will it only ever exist in an environment where one tagged version could be injected? (If this is true, it would mean we cannot share the same initialised `Sandbox` for all tests). Despite initialising the tests by binging the two classes as they are done in the `ServiceProvider` - inversify fails with: `Ambiguous match found for serviceIdentifier` (suggesting that this would happen at runtime). In order to get the tests running, I’ve added a tag to the `WalletRepository` consumed by `BlockState` for now, but this isn’t the correct solution (since presumably we want `BlockState` to be able to work with *both* tagged versions), do let me know if anyone has suggestions on how to fix this, or whether I'm misunderstanding something...

Finally, two tests are intentionally left failing in `StateBuilder.test.ts` as the code implementation here needs review.


packages/core-state/src/wallets/wallet-repository.ts
```diff
@@ -139,15 +139,25 @@ export class WalletRepository implements Contracts.State.WalletRepository {
...
     }
 
+    // TODO: naming here implies the behaviour is different to reindexing
     public index(wallets: ReadonlyArray<Contracts.State.Wallet>): void {
         for (const wallet of wallets) {
             this.reindex(wallet);
         }
     }
 
+    // TODO: reindex gives the impression this wallet is already indexed - possibly change name?
```
faustbrian: We should probably merge this into `index` and drop the method completely. We could need to change the signature of the `index` method to `public index(wallets: Contracts.State.Wallet | ReadonlyArray<Contracts.State.Wallet>): void` so we can pass in either a single wallet or array. Depending on how often `reindex` is called with a single wallet we might even drop it completely and simply pass in an array.

packages/core-state/src/wallets/wallet-repository.ts
```diff
@@ -139,15 +139,25 @@ export class WalletRepository implements Contracts.State.WalletRepository {
...
+            index.entries().forEach(([name, wallet]) => {
+                if (wallet.publicKey === forgottenWallet?.publicKey) {
+                    index.forget(name);
+                }
+            });
+        }
         this.getIndex(indexName).forget(key);
     }
 
+    // TODO: naming here implies the behaviour is different to reindexing
```
faustbrian: See `reindex` comment.

packages/core-state/src/wallets/wallet-repository.ts
```diff
@@ -139,15 +139,25 @@ export class WalletRepository implements Contracts.State.WalletRepository {
     }
 
     public forgetByIndex(indexName: string, key: string): void {
+        const forgottenWallet = this.getIndex(indexName).get(key);
+        for (const index of Object.values(this.indexes)) {
+            index.entries().forEach(([name, wallet]) => {
```
faustbrian: `forEach` should be avoided in favour of `for (const ... of ...)` due to performance implications on larger datasets.

bertiespell: Resolved here => cf1e3ddf8df5f90282c8f174c3cef6885229cc2d

packages/core-state/src/wallets/wallet-repository.ts
bertiespell: Fixed here => b0d2d682c4d6356b1a0bc9250e9907396fa6ee10

packages/core-state/src/wallets/wallet-repository.ts
bertiespell: Also fixed here => b0d2d682c4d6356b1a0bc9250e9907396fa6ee10.

packages/core-state/src/block-state.ts
```diff
@@ -9,22 +9,30 @@ export class BlockState {
     private readonly app!: Contracts.Kernel.Application;
 
     @Container.inject(Container.Identifiers.WalletRepository)
+    @Container.tagged("state", "blockchain") // TODO: Without this line - and despite
```
bertiespell: @rainydio Is this class supposed to use this tagged version? Or is it designed to work with different injected tags? If so - is there a way to make it explicit which tagged version a class is supposed to be used with?

Without this line, the test suite fails with Ambiguous match found for serviceIdentifier Repository<Wallet>. The Sandbox setup is initialised in more or less the same way as the ServiceProvider - I'm wondering if this would not fail also?

packages/core-state/src/dpos/dpos.ts
```diff
@@ -3,10 +3,11 @@ import { Utils } from "@arkecosystem/crypto";
 
 @Container.injectable()
 export class DposState implements Contracts.State.DposState {
-    @Container.inject(Container.Identifiers.Application)
-    private readonly app!: Contracts.Kernel.Application;
+    @Container.inject(Container.Identifiers.LogService)
+    private logger!: Contracts.Kernel.Logger;
 
     @Container.inject(Container.Identifiers.WalletRepository)
+    @Container.tagged("state", "blockchain") // TODO: see todo in block-state
```
bertiespell: @rainydio Is this class designed to be used with a specific Container.Identifiers.WalletRepository (as per the change here), or is it intended to work with multiple tagged versions? See my comment above about the same issue.

packages/core-state/src/block-state.ts
rainydio: It's supposed to be tag agnostic. It doesn't care which particular wallet repository it's using. You probably want to use getTagged if you want blockchain version. But definitely do not add that tag there.

PR #3553, opened 2020-02-29
test(core-transaction-pool): increase coverage to 100%

### LINKED ISSUE #3482: Increase test coverage of core-transaction-pool to 100%

rainydio:
## Summary

In addition to tests there are changes that are fixing various issues that were discovered.

#### Transaction has expired error

Previously it reported how many blocks ago transaction expired. But when transaction is entering the pool its expiration is checked against next height. So when current height is 100 and transaction expiration is set to 101 it is considered expired and cannot enter. It's unclear how to report blocks ago in such case. I changed error message to report expiration height instead of how many blocks ago it expired.

#### Sender state race condition

Although hypothetical it was possible to cause races due to async nature of transaction handler. Now sener's transactions are acquiring lock while applying or manipulating stored transactions array.

#### Error during revert and in-flight transactions

If error was raised during revert (should not happen) transactions that were waiting for lock to release cannot be applied anymore. State is considered corrupted, there isn't much that can be done with it. Any transactions that were waiting for lock to release will raise an error.

#### Mempool

Sender state was split into two classes `SenderState` that has only `apply` and `revert` functions and also performs transaction related checks and `SenderMempool` that keeps list of applied transactions and performs sender related checks. `Memory` was renamed into `Mempool`.

Fixes #3482

## Checklist

- [x] Collator
- [x] Dynamic fee matcher
- [x] Errors
- [x] Expiration service
- [x] Mempool
- [x] Processor
- [x] Query
- [x] Sender mempool
- [x] Sender state
- [x] Service
- [x] Storage
- [x] Utils
- [ ] Functional tests


faustbrian:
There is still quite a lot uncovered https://codecov.io/gh/ArkEcosystem/core/tree/test%2Fcore-transaction-pool/packages/core-transaction-pool/src.

PR #3593, opened 2020-03-09
test(core-kernel): increase coverage to 100%

### LINKED ISSUE #3474: Increase test coverage of core-kernel to 100%

rainydio:
##  Summary

`core-kernel` package tests, fixes #3474. Couple of problems though:

### MemoryQueue

There is unhandled promise rejection and there is no way to handle it or to report error to caller.

Then there is also problem with `lastResults` which is passed as argument to `processFromIndex` and at the same time kept as private field. It's mutable array and it's unclear why it's being passed around. Also there is no way to get those `lastResults` back. I could not find a single usage to determine the intent.

Another problem is with `lastQueue` field. It's set in couple of places but isn't ever read. As result calling `resume` several times will rerun same task again.

### ClassManager

It has `drivers` field and functions to extend that. But that field isn't used during instantiation in `createDriver` function. I removed those. I could instantiate new instances on my own, but usual implementation through `createDriverNameDriver` is asynchronous. As for `getDrivers` method there is no way to return classes that might be created through `createDriverNameDriver`.

I guess it was refactored to support async creation and class instantiation wasn't removed completely:

```typescript
import { Kernel } from "../contracts";
// import { DriverCannotBeResolved } from "../exceptions/container";
import { Identifiers, inject, injectable } from "../ioc";
import { pascalCase } from "../utils";
```

## Checklist

- [x] WatchConfiguration bootstraper
- [x] IOC selectors
- [x] PluginConfiguration
- [x] PluginManifest
- [x] MemoryCacheStore
- [x] LocalConfigLoader
- [x] JoiValidator
- [x] Pipeline ServiceProvider
- [x] ClassManager
- [x] InstanceManager
- [x] Utils
- [x] NullCacheStore, NullEventDispatcher, NullFilesystem, NullLogger, NullPipeline, NullQueue, NullValidator
- [ ] MemoryQueue


faustbrian:
`MemoryQueue` will be sorted in another PR as discussed and `ClassManager` I'll have a look at when I am back.

PR #3623, opened 2020-03-27
test(core): increase coverage to 100%

bertiespell:
## Summary

Increase test coverage in the core package to 100%

## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged


packages/core/src/cli.ts
```diff
@@ -95,20 +95,21 @@ export class CommandLineInterface {
...
         const discoverer = this.app.resolve(Commands.DiscoverCommands);
+        const commands: Contracts.CommandList = discoverer.within(resolve(dirname, "./commands"));
 
-        const commands: Contracts.CommandList = discoverer.within(resolve(__dirname, "./commands"));
         const commandsFromPlugins = discoverer.from(
             this.app.get<Contracts.Config>(Container.Identifiers.Config).get("plugins"),
         );
 
         for (const [key, value] of Object.entries(commandsFromPlugins)) {
+            // TODO: check we can ignore this line for now
```
faustbrian: Can be ignored for now, I'll resolve it once I finalise the implementation of the CLI.

packages/core/src/commands/config-cli.ts
```diff
@@ -63,9 +63,10 @@ export class Command extends Commands.Command {
      * @memberof Command
      */
     public async execute(): Promise<void> {
-        if (this.hasFlag("token")) {
-            this.config.set("token", this.getFlag("token"));
-        }
+        // TODO: check
```
faustbrian: That must've been a copy & paste mistake on my end when I created the command. Since this is a command to alter the configuration nothing should have a default value which means `token` could actually be empty.

packages/core/src/commands/network-generate.ts
```diff
@@ -119,7 +119,11 @@ export class Command extends Commands.Command {
                 } as prompts.PromptObject<string>),
         );
 
-        if (Object.keys(flagsDefinition).find(flagName => !response[flagName])) {
+        // TODO: check this fix is acceptable
+        // the distribute flag is a boolean in the pre-existing tests
+        // and it is defined as a number in this.generateCryptoGenesisBlock()
```
faustbrian: The value should be treated as a boolean, must've also been a mistake when copying over some of the code from 2.6

packages/core/src/commands/network-generate.ts
```diff
@@ -536,12 +538,6 @@ export class Command extends Commands.Command {
             byteBuffer.writeByte(generatorByte);
         }
 
-        if (genesisBlock.blockSignature) {
```
faustbrian: What was the issue with this block of code that it had to be removed?

packages/core/src/commands/plugin-install.ts
```diff
@@ -49,12 +49,10 @@ export class Command extends Commands.Command {
         const pkg: string = this.getArgument("package");
 
         try {
-            await this.install(pkg);
+            return await this.install(pkg);
         } catch (error) {
             throw new Error(error.message);
         }
-
-        throw new Error(`The given package [${pkg}] is neither a git nor a npm package.`);
```
faustbrian: Could we preserve this error and move it into `this.install()` **after** the `for` loop? Placing it after the `for` loop means that the given pkg couldn't be found on any available source.

faustbrian:
@bertiespell conflicts

packages/core/src/commands/network-generate.ts
bertiespell: So is the fix here acceptable?

The problem is that, under the previous implementation, a value intentionally set to false here would have triggered the error (similarly a number intentionally set to 0 would trigger the error).

packages/core/src/commands/network-generate.ts
bertiespell: I think the code is unreachable - `blockSignature` is set to undefined on line 485 - then passed immediately to `this.getBlockId(block)`, which in turn immediately calls this function here.

packages/core/src/commands/network-generate.ts
faustbrian: I think it's fine for now. I'll probably revisit it when I do some other refactors and cleanups around that command.

PR #3643, opened 2020-04-06
feat(core-p2p): hapi/nes implementation

air1one:
## Summary

WIP on hapi/nes PoC implementation of the websocket layer (instead of socketcluster).

## Checklist

- [ ] Tests _(if necessary)_
- [ ] Ready to be merged


packages/core-p2p/src/socket-server/controllers/peer.ts
```diff
@@ -0,0 +1,202 @@
...
+
+        const blockchain = this.app.get<Contracts.Blockchain.Blockchain>(Container.Identifiers.BlockchainService);
+        return {
+            common: commonBlocks[0],
+            lastBlockHeight: blockchain.getLastBlock().data.height,
+        };
+    }
+
+    public async getStatus(request: Hapi.Request, h: Hapi.ResponseToolkit): Promise<Contracts.P2P.PeerPingResponse> {
+        this.validatePayload(request.payload, peerSchemas.getStatus);
```
faustbrian, line 67: Could this be moved into a plugin that is responsible for validation? Seems to be duplicated in every controller method.

packages/core-p2p/src/socket-server/controllers/peer.ts
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+    public async getBlocks(
+        request: Hapi.Request,
+        h: Hapi.ResponseToolkit,
+    ): Promise<Interfaces.IBlockData[] | Contracts.Shared.DownloadBlock[]> {
+        this.validatePayload(request.payload, peerSchemas.getBlocks);
+        
+        // Add the peer if not already (on every peer endpoint)
+        this.peerProcessor.validateAndAcceptPeer({ ip: request.info.remoteAddress } as Contracts.P2P.Peer);
```
faustbrian, line 180: Could this be moved into a plugin that is responsible for peer acceptance? Seems to be duplicated in every controller method.

air1one: done.

packages/core-p2p/src/socket-server/controllers/peer.ts
air1one: done.

faustbrian:
This line https://github.com/ArkEcosystem/core/blob/p2p/hapi-nes/jest.config.js#L21 needs to be removed now that the initial rework was done so we can make sure we get back to 100% coverage for it.

PR #3650, opened 2020-04-15
feat(crypto): dynamic block times

### LINKED ISSUE #3627: Dynamic Block Times
# Problem

The blocktime is exposed as a milestone, meaning it could potentially change at any height. But doing so will break many things in unpredicable ways as Core is not written with dynamic block times in mind.

In order to work correctly, all places in Core that work with block times have to get the block time relative to a height and when calculating slots to look at all milestones that affected the block time similar to how the round calculator takes a dynamic number of active delegates into account.

## Slots

[https://github.com/ArkEcosystem/core/blob/develop/packages/crypto/src/crypto/slots.ts#L28](https://github.com/ArkEcosystem/core/blob/develop/packages/crypto/src/crypto/slots.ts#L28)

`getSlotNumber`, `getSlotTime` and `isForgingAllowed` always assume `height=1` and use a fixed block time. Instead they need to accept a height (if omitted fall back to latest height ) and then take all milestones from first to assumed height that affect the block time, calculate each span separately and add them together to finally get the correct value.

## HtlcLock timestamp expiration

[https://github.com/ArkEcosystem/core/blob/develop/packages/core-transactions/src/handlers/two/htlc-lock.ts#L84](https://github.com/ArkEcosystem/core/blob/develop/packages/core-transactions/src/handlers/two/htlc-lock.ts#L84)

A margin of approx. one round (assuming a fixed block time) is added on top of the specified expiration for safety reasons. Can be removed with Tendermint, so wouldn't bother with this usage of block time.

## Transaction Pool

[https://github.com/ArkEcosystem/core/blob/develop/packages/core-transaction-pool/src/processor.ts#L338](https://github.com/ArkEcosystem/core/blob/develop/packages/core-transaction-pool/src/processor.ts#L338)[https://github.com/ArkEcosystem/core/blob/develop/packages/core-transaction-pool/src/memory.ts#L123](https://github.com/ArkEcosystem/core/blob/develop/packages/core-transaction-pool/src/memory.ts#L123)

The blocktime is always taken from the current height. So the actual expiration of transactions
varies (could be longer or shorter, depending on block time on next milestone). Either this is ignored, because that should only apply to V1 transactions (easiest solution) or when a transaction is memorized, the current height is memorized as well to always refer to the same height so the expiration is more deterministic.

## getCurrentRound

[https://github.com/ArkEcosystem/core/blob/develop/packages/core-p2p/src/socket-server/versions/internal.ts#L94](https://github.com/ArkEcosystem/core/blob/develop/packages/core-p2p/src/socket-server/versions/internal.ts#L94)

One of the more complex parts using block times and most error prone. The way current/next forger are calculated has to be adjusted.

## DelegateTracker

[https://github.com/ArkEcosystem/core/blob/develop/packages/core-forger/src/delegate-tracker.ts#L105](https://github.com/ArkEcosystem/core/blob/develop/packages/core-forger/src/delegate-tracker.ts#L105)

It assumes the same time for each slot, it should be sufficient to peek the next milestone at each slot and use that time for calculating `secondsToForge` more accurately.

## Blockchain

[https://github.com/ArkEcosystem/core/blob/develop/packages/core-blockchain/src/blockchain.ts#L434](https://github.com/ArkEcosystem/core/blob/develop/packages/core-blockchain/src/blockchain.ts#L434)

Also assumes a fixed block time to calculate the slot time. If the `Slots` class is fixed, this
could be replaced with something like `Slots.getSlotTime(lastHeight) <= currentBlock.data.timestamp`

## Test Framework

There are several places that use a fixed block time. But this shouldn't require any changes as
testing dynamic block times can be done via dedicated tests.

bertiespell:
## Summary

In order to support dynamic block times, the calculations for `slotNumber` and `slotTime` now require a lookup of the last block (which was forged before a new milestone takes affect) - in order to retrieve its timestamp. The time between two milestones is referred to here as a “Span” (the number of Slots over any given span is calculated as the difference between two timestamps divided by the relevant blockTime).

For example, the following milestone data has 4 spans: `0 - 2`, `3 - 5`, `6 - 7`, `8+`.

```
const milestones = [
    { height: 1, blocktime: 8 },
    { height: 3, blocktime: 9 },
    { height: 6, blocktime: 10 },
    { height: 8, blocktime: 8 },
];
```

In order to avoid creating a dependency on `Core` from the `Crypto` package, this lookup functionality is passed through to the `Slots` methods. Since this requires a database look-up, some methods that were formerly synchronous are now async. It’s worth taking care when reviewing this PR to consider when/where methods are newly awaited.

The other thing to note has to do with *which* blocks we lookup (for span calculations). Since we only ever need to lookup blocks before a milestone change, the current implementation will throw when attempting to look up other heights, as the lookup is only ever called inside the `Slots` class using `nextMilestoneHeight - 1` (on lines 107, 109 and 172, 173).  Since we will rarely be retrieving new data (the only time is when we hit a new height with a milestone change), we should consider implementing a cache here. I’ve left a `todo` in the code where I think this could be done.

## Checklist

- [x] Documentation _(if necessary)_
- [x] Tests _(if necessary)_
- [x] Ready to be merged

Closes #3627


faustbrian:
@bertiespell conflicts and failing tests

__tests__/unit/core-blockchain/processor/block-processor.test.ts
```diff
@@ -298,13 +298,15 @@ describe("BlockProcessor", () => {
...
+            const activeDelegatesWithoutGenerator = [];
+            activeDelegatesWithoutGenerator.length = 51;
+            activeDelegatesWithoutGenerator.fill(notBlockGenerator, 0);
+
+            databaseService.getActiveDelegates = jest.fn().mockReturnValueOnce(activeDelegatesWithoutGenerator);
 
             const blockProcessor = sandbox.app.resolve<BlockProcessor>(BlockProcessor);
 
-            await blockProcessor.process(block);
+            expect(await blockProcessor.process(block)).toResolve();
```
faustbrian, line 309:
```suggestion
            await expect(blockProcessor.process(block)).toResolve();
```

bertiespell: Fixed here => e15b839d3a8e064ff37e64cf74e9fafba20d919e.

PR #3657, opened 2020-04-21
test(core-database): block and transaction history services

rainydio:
## Summary

This is follow-up to #3644.

- Search-related types are moved into Contracts.Search
- Expressions are now simple objects. It's easier to test and compare them.
- Block or transaction => model => block or transaction data conversion is moved into separate class.
- Unit tests

## Checklist

- [x] Tests
- [x] Ready to be merged


rainydio:
Re-executed CI.

PR #3673, opened 2020-05-01
test(core-database): increase coverage to 100%

rainydio:
## Summary

I'm not proud of tests that I wrote. They aren't actually testing but merely covering database service. I fixed some obvious errors and left many comments around suspicious lines. It's extremely difficult to test it. I think it requires complete rewrite.

## Checklist

- [x] Tests
- [x] Ready to be merged


faustbrian:
The database service refactor had been delayed for obvious reasons but the goal for it was to split it into multiple services that each are responsible for specific things like for example handling interactions with the database that relate to blocks or rounds instead of having this god-entity that handles everything.

**This is still on the roadmap but not critical for the initial launch of the new major release.**

PR #3684, opened 2020-05-06
fix(core-kernel): get app version from package.json

air1one:
## Summary

Get app version from package.json.

Also fixed issue in core-database where detectMissedBlocks was called twice.

## Checklist

- [ ] Tests _(if necessary)_
- [x] Ready to be merged


packages/core-kernel/src/bootstrap/app/register-base-bindings.ts
```diff
@@ -27,19 +28,20 @@ export class RegisterBaseBindings implements Bootstrapper {
      */
     public async bootstrap(): Promise<void> {
         const flags: Record<string, string> | undefined = this.app.config("app.flags");
+        const version: string = require(resolve(__dirname, "../../../package.json")).version;
```
faustbrian, line 31:
```suggestion
        const { version } = require(resolve(__dirname, "../../../package.json"));
```

PR #3693, opened 2020-05-11
feat(core-transaction-pool): implement workers

### LINKED ISSUE #3626: Transaction Pool Worker
An initial explanation of the issue and idea based @supaiku0. This most likely will need refinement as implementation is ongoing.

-----

A few months @supaiku0 worked on a proof of concept:
[https://github.com/ArkEcosystem/core/tree/wip/core-transaction-pool/worker](https://github.com/ArkEcosystem/core/tree/wip/core-transaction-pool/worker)

It is nowhere near production-ready, but in principle working. Currently, the `/transactions` POST endpoint can be easily abused to cause high load on nodes, because it validates all transactions on the main thread, which causes CPU spikes and this can even be caused by broadcasting invalid transactions targeted at specific nodes since the signature check is pretty heavy. This is the reason why it is advised for node operators to secure their `core-api` access or completely disable it if used in front of a forger.

The p2p endpoint already received a workaround, by giving the main thread room to breath: [https://github.com/ArkEcosystem/core/pull/2848/files](https://github.com/ArkEcosystem/core/pull/2848/files)

So the problem only manifests when using the `core-api` endpoint. However, ideally, this workaround is replaced with a more generic solution which also affects `core-api`. This is where the pool worker comes in.

The flow right now is:
POST /transactions -> create new `Processor` instance -> validate -> addTransactionsToPool
-> return response (accepted, ignored, excess, error)

With the pool worker it would change to something like this:
POST /transactions -> enqueue transactions which creates a job -> return response (`ticketId`, e.g a sequentially increasing number)

`ticketId` represents a job, that is either in the queue (about to be sent to worker), being processed (somewhere in worker) or done (returned from worker).

End users/clients can query the status by hitting an endpoint of a peer they broadcasted to using the ticket id which would return a response resembling what they get currently if they use the endpoint.

**^ This is a significant change API wise and breaks all kind of client software so that's why the worker has been postponed to 3.0**

---

Queued jobs are then pushed to the worker which does all the heavy lifting and once done reports back to the main thread, which will add valid transactions to the transaction pool.

---

A nice benefit of this approach is that it also makes the frequency in which transactions are rebroadcasted by a node to other peers more deterministic. Right now they rebroadcast whenever they are done validating the current batch of transactions (i.e. 1 request -> 1 broadcast), while a worker could report his finished jobs like only every 100ms, then a behaving node would at most rebroadcast 10 times per second. which greatly reduces the snowball effect that can currently be observed when the network is flooded with many transactions. Also, the rate-limit on the endpoint can then be properly calibrated.

rainydio:
## Summary

Fixes #3626

Pool starts several worker processes that deserialize transactions. I added couple of IPC helper classes to `core-kernel` utils. Pool is using them to communicate with child processes. Child processes also load crypto packages specified in configuration, so custom transactions can be deserialized too,

## Checklist

- [x] Tests
- [x] Ready to be merged


packages/core-transaction-pool/src/worker.ts
```diff
@@ -0,0 +1,39 @@
...
+
+    public async getTransactionFromData(
+        transactionData: Interfaces.ITransactionData,
+    ): Promise<Interfaces.ITransaction> {
+        const currentHeight = Managers.configManager.getHeight()!;
+        if (currentHeight !== this.lastHeight) {
+            this.lastHeight = currentHeight;
+            this.ipcSubprocess.sendAction("setHeight", currentHeight);
+            this.ipcSubprocess.sendAction("setConfig", Managers.configManager["config"]);
+            this.ipcSubprocess.sendAction("setMilestone", Managers.configManager.getMilestone());
```
air1one, line 33: That should not be necessary, we get milestone from config and height which are sent above. (please double-check from crypto package in case I missed something :smile:)

packages/core-transaction-pool/src/worker.ts
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    public async getTransactionFromData(
+        transactionData: Interfaces.ITransactionData,
+    ): Promise<Interfaces.ITransaction> {
+        const currentHeight = Managers.configManager.getHeight()!;
+        if (currentHeight !== this.lastHeight) {
+            this.lastHeight = currentHeight;
+            this.ipcSubprocess.sendAction("setHeight", currentHeight);
+            this.ipcSubprocess.sendAction("setConfig", Managers.configManager["config"]);
```
air1one, line 32: Add a getter on crypto config instead of getting a private attribute here ?

faustbrian: There already is the `all` method on the `ConfigManager` which returns `config`.

air1one: right then use `all` :+1:

rainydio: Fixed

packages/core-transaction-pool/src/worker.ts
rainydio: Removed `setMilestone` call

faustbrian:
@rainydio conflicts

PR #3694, opened 2020-05-11
feat(core-manager): implement `info.coreStatus` action

sebastijankuzner:
## Summary

This PR add new action for getting current application process status and syncing status if available.

---

### Action

**Name:** info.coreStatus

**Params:**
|Name|Type|Description|
|-|-|-|
|token|String|(Optional). Name of the token. Used for getting process by token name|

**Response:**
|Name|Type|Description|
|-|-|-|
|processStatus|String|Proces status|
|syncing|Boolean|(Optional) Is node in syncing mode? Returned only if running node returned successful response.|

**Possible process statuses:**
- undefined
- online
- stopped
- stopping
- waiting restart
- launching
- errored
- one-launch-status

## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged

### DIFF
--- a/packages/core-manager/package.json
+++ b/packages/core-manager/package.json
@@ -32,6 +32,9 @@
         "hapi-auth-bearer-token": "^6.1.6",
         "latest-version": "^5.1.0"
     },
+    "devDependencies": {
+        "nock": "^12.0.3"
+    },
     "engines": {
         "node": ">=10.x"
     },
--- a/packages/core-manager/package.json
+++ b/packages/core-manager/package.json
@@ -22,18 +22,15 @@
     },
     "dependencies": {
         "argon2": "^0.26.2",
-        "axios": "^0.19.2",
         "@arkecosystem/core-kernel": "^3.0.0-next.0",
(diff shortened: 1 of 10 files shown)


packages/core-manager/package.json
```diff
@@ -22,6 +22,7 @@
     },
     "dependencies": {
         "argon2": "^0.26.2",
+        "axios": "^0.19.2",
```
faustbrian, line 25: We use https://github.com/sindresorhus/got instead of axios in all our projects.

sebastijankuzner: Ok. I was looking for something that you use. Do you have any existing wrapper around? If not, then I will use existing HttpClient class.

faustbrian: Not sure why you would need a wrapper? Just can simply call `got(url)` and use https://github.com/nock/nock to mock the responses.

sebastijankuzner: For wishes like this :) Changes or replacements of 3rd party libraries are then handled only in once place.

PR #3699, opened 2020-05-13
feat(core-manager): implement `info.databaseSize` action

sebastijankuzner:
## Summary

This PR add new action for getting database space using pg_database_size SQL method. 

### Action

**Name:** info.databaseSize

**Params:** none

**Response:**
|Name|Type|Description|
|-|-|-|
|size|Number|Size of the database in KB.|

## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged

### DIFF
--- a/packages/core-manager/src/defaults.ts
+++ b/packages/core-manager/src/defaults.ts
@@ -16,17 +16,6 @@ export const defaults = {
             },
         },
     },
-    connection: {
-        type: "postgres",
-        host: process.env.CORE_DB_HOST || "localhost",
-        port: process.env.CORE_DB_PORT || 5432,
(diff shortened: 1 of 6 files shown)


packages/core-manager/src/defaults.ts
```diff
@@ -16,6 +16,17 @@ export const defaults = {
             },
         },
     },
+    connection: {
```
faustbrian, line 19: This should be automatically handled. There is no point in the user defining duplicated data because the same database as the relay process has to be used to get the data.

dated:
Wouldn't it be more consistent with other actions (#3698) if the raw size gets returned instead of the human readable format or at least add a flag to control the formatting?

sebastijankuzner:
@dated size is now expressed in KB and returned as number type to be consistent with other actions.

PR #3710, opened 2020-05-18
feat(core-kernel): implement process actions

sebastijankuzner:
## Summary

This PS will include new core-kernel service called remote-actions. Remote actions are node endpoints that allow running node actions from other process via terminal. Current implementation uses [@pm2/io](https://www.npmjs.com/package/@pm2/io) package to support remote actions. Action can be extended to support parameters in the future. [Link](https://pm2.keymetrics.io/docs/usage/process-actions/) to official documentation. 

Command can be run with from terminal like: 

pm2 trigger [application-name] [action-name] [parameter]

## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged


sebastijankuzner:
@faustbrian I aggree. Makes much more sense if we use process actions. PM2 uses both terms for the same things. Take a look into package [documentation](https://www.npmjs.com/package/@pm2/io).

PR #3725, opened 2020-05-21
feat(core-api): estimateTotalCount API parameter

### LINKED ISSUE #3711: Add config parameter for core-database to allow using real counts
In core-database package add a config parameter to use real counts instead of estimates (default it to not using real counts in `defaults.ts`)

rainydio:
## Summary

Fixes #3711

Added two options to api configuration:
 - `estimateTotalCount` that controls default behavior of executed query
  `CORE_API_NO_ESTIMATED_TOTAL_COUNT` env variable
 - `chooseEstimateTotalCount` that enables client to choose desired behavior
  `CORE_API_CHOOSE_ESTIMATE_TOTAL_COUNT` env variable

It's implemented as Joi schema. So if client requires particular request to be executed without estimations, he can set `estimateTotalCount=no` and if it's not enabled on the server he'll get bad request error response indicating that this value isn't allowed.

So this is client controlled, but server restricted with default values chosen by server.

## Checklist

- [x] Ready to be merged

### DIFF
--- a/__tests__/functional/core-database/repositories/block-repository.test.ts
+++ b/__tests__/functional/core-database/repositories/block-repository.test.ts
@@ -280,7 +280,7 @@ describe("BlockRepository.listByExpression", () => {
                 { property: "id", direction: "desc" },
             ],
             { offset: 0, limit: 2 },
-            { estimateCount: false },
+            { estimateTotalCount: false },
         );
         expect(listResult).toStrictEqual({
             rows: [toBlockModel(block3), toBlockModel(block2)],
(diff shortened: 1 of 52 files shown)


faustbrian:
As mentioned in https://github.com/ArkEcosystem/core/issues/3711#issuecomment-630623030 there is no need to introduce anything new. Simply reuse the existing https://github.com/ArkEcosystem/core/blob/develop/packages/core-database/src/defaults.ts#L13 parameter from the initial implementation in https://github.com/ArkEcosystem/core/commit/1e185eab27770aa76ef409ad3805074847f63554 to determine if an estimate should be used.

rainydio:
@air1one @faustbrian removed it

faustbrian:
Tests are failing.

PR #3732, opened 2020-05-25
feat(core-manager): implement `snapshots.restore` action 

sebastijankuzner:
## Summary

This PR add new action which restores snapshot with given parameters. Action returns error if snapshot cannot be found. 

### Action

**Name:** snapshots.restore

**Params:** 
|Name|Type|Description|
|-|-|-|
|name|String|Snapshot name.|
|truncate|Boolean|(Optional; Default: false) Truncate database before restore?|
|verify|Booelan|(Optional; Deafault: false) Verify data during restore. |

**Response:**: Empty response if snapshot process is started successfully. It doesn't guarantee that snapshot process completed successfully. 

## Checklist

- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/packages/core-manager/src/actions/snapshots-restore.ts
+++ b/packages/core-manager/src/actions/snapshots-restore.ts
@@ -1,4 +1,5 @@
-import { Application, Container } from "@arkecosystem/core-kernel";
+import { Application, Container, Contracts } from "@arkecosystem/core-kernel";
+import { join } from "path";
 
 import { Actions } from "../contracts";
 import { Identifiers } from "../ioc";
@@ -11,7 +12,7 @@ export class Action implements Actions.Action {
     public schema = {
         type: "object",
         properties: {
-            blocks: {
+            name: {
                 type: "string",
             },
             truncate: {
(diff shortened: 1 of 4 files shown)


packages/core-manager/src/actions/snapshots-restore.ts
```diff
@@ -0,0 +1,52 @@
...
+            throw new Error("Snapshot not found");
+        }
+
+        await this.snapshotManager.restore({
+            network: this.app.network(),
+            blocks: params.name,
+            ...params,
+        });
+
+        return {};
```
faustbrian, line 50: This should respond with a `202` status code to indicate that the request was accepted but some processing is ongoing.

sebastijankuzner: 202 is for HTTP response. Actions are part of the JSON-RPC specification and doesn't know nothing about HTTP protocol. We can still add responseCode parameter into action response that will override default 200 response code, but that requires changes in [RCP](https://github.com/faustbrian/hapi-json-rpc) plugin or addition plugin that will check the action responseCode parameter.

PR #3733, opened 2020-05-25
fix(core-api): include multipayment in forged amount

### LINKED ISSUE #3495: multipayment totals not included in the amount of a block
I don't know if you consider this is a bug or normal behaviour so if it's normal behaviour then feel free to close this issue.

The amount of any multipayment is not included in the `total` or `amount` field of a block. For example https://explorer.ark.io:8443/api/blocks/16d34d2471c13adae6b329d770670cccc53244fae3983bad72510a8a2812f57c shows the forged amount as 21 ARK since it contains 3 transfer transactions for 7 ARK each, but ignored the multipayment amounts within the same block.

In my opinion that is misleading and makes things like the explorer look wrong when viewing the processed amount of a block containing multipayments. I respectfully think it should include the sum of any multipayments in the block too.

rainydio:
## Summary

Fixes #3495

Block `totalAmount` property is signed and is used to checksum database. It cannot be changed to include multipayment transactions without milestone. So this is implemented on API level, but things there aren't that simple either.

I cannot simply pull all transactions with every block that is loaded from database. Otherwise this will significantly stress database on `/blocks` and `/blocks/search` endpoints. Second there is `raw=true` parameter which presumably still has to return raw block data that can be used to check signature (so without multipayment included in total).

1. `BlockHistoryService` now has capability of joining transactions (`...byCriteriaJoinTransactions` methods).
1. New `BlockWithTransactionsResource` was added. Only its `transform` method is used which walks through transactions and calculates forged total that includes multipayment.
1. All of transformed `BlockResource` usages are now using `BlockWithTransactionsResource`.
1. Those usages requiring transformed `BlockWithTransactionsResource` use new `..byCriteriaJoinTransactions` methods to also join multipayment transactions only so forged total includes multipayment transactions.

## Checklist

- [ ] Tests
- [ ] Ready to be merged


faustbrian:
@rainydio conflicts

PR #3738, opened 2020-05-25
chore: CentOS install fixes

adrian69:
## Summary
- Added jemalloc library.
- RedHat derivatives installs update:
  - Added missing flag to `yum clean`.
  - Switched NTP to Chrony (due to CentOS 8 `ntpd` deprecation).
  - Additional record to `pg_hba.conf` (thanks to @rainydio).
## Checklist
- [x] Ready to be merged

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -149,10 +149,10 @@ success "Installed PM2!"
 heading "Installing program dependencies..."
 
 if [[ ! -z $DEB ]]; then
-    sudo apt-get install build-essential libcairo2-dev pkg-config libtool autoconf automake python libpq-dev jq -y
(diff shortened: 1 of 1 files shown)


alessiodf:
Is `epel-release` required for CentOS?

I ask because it prevents installation on Fedora as the package is not found. Removing it allows the installation to succeed and then Core works properly on Fedora. If it wasn't needed for CentOS then removing it would make the script work on both CentOS and Fedora.

adrian69:
Although it was originally added because of CentOS6 support, currently `jemalloc` install will fail without it even on CentOS8, so unfortunately it's needed.

alessiodf:
Maybe it could be possible for the script to not abort if `epel-release` is missing then?

alessiodf:
Actually, some other changes would be necessary for Fedora too, probably not worth it if you just want to support CentOS...

Issue #3759, opened 2020-06-01
Core v3: Some transactions don't return timestamp


boldninja:
**Expected Behavior**

All transactions using API should return a timestamp.

**Current Behavior**

Some transactions when using `api/wallets/<address>/transactions` don't return timestamp in response.

**Steps to Reproduce (for bugs)**

Make a call to an address that has a few tx and some don't have timestamp `api/wallets/<address>/transactions`

Example below:
```

      {
         "id":"fdd6cbaaa5c23594fd70e0c7c5637ed4abbbda4f352726bddbd77a859f53a57f",
         "blockId":"4863696540053150556",
         "version":2,
         "type":0,
         "typeGroup":1,
         "amount":"200000000",
         "fee":"10000000",
         "sender":"AFyf2qVpX2JbpKcy29XbusedCpFDeYFX8Q",
         "senderPublicKey":"02f7acb179ddfddb2e220aa600921574646ac59fd3f1ae6255ada40b9a7fab75fd",
         "recipient":"AFyf2qVpX2JbpKcy29XbusedCpFDeYFX8Q",
         "signature":"210f5e03a5ca0f4fd839c8333c0ba9e917d41a0e3eccf79a8beac46b41dc7ec2740765dd57f420681e46e235d0f7e09a974764fc532767fe861746c4c9f2a7b2",
         "confirmations":0,
         "nonce":"3"
      },
```

rainydio:
Fixed

rainydio:
Almost forgot about other endpoints. Reopening for now.

rainydio:
Fixed

Issue #3741, opened 2020-05-26
/api/transactions/search to handle search by recipient from multipayment recipients


air1one:
`POST /api/transactions/search` currently doesn't support to search multipayment tx by recipients since the recipients sit in the asset of the multipayment transaction.

rainydio:
Okay. But as an alternative I can also enable OR condition for search endpoints, so that multipayment can be included like this:

```json
[
  { "recipientId": "<recipient address>" },
  { "asset": { "payment": { "recipientId": "<recipient address>"  } } }
]
```

Or I can add additional aggregated criteria property:

```json
{ "anyRecipientId": "<recipient address>" }
```

Otherwise it isn't clear why transactions that are included into result may have unexpected recipientId that doesn't match specified criteria (because asset.payment.recipientId does).

air1one:
Yeah OR could be good, what do you think @faustbrian ?

faustbrian:
Not really a fan of that because we don't require special recipient types to be specified for anything else and that just makes it more tedious to retrieve transactions.

When you request transactions via API without a `type` or `typeGroup` filter your expectation would be that you get all transactions that could have a `recipientId` without thinking about what type the transaction is.

Also, this has to work with the `addresses` parameter which can be used to look up transactions for multiple addresses at the same time.

rainydio:
Not gonna fight for it

rainydio:
- `recipientId` now searches through multipayment too
- added `address` criteria to filter by either `recipientId` or `senderId`

PR #3761, opened 2020-06-01
fix(core-api): fix wallet/transactions confirmations and timestamp fields

### LINKED ISSUE #3758: Core v3: API response shows 0 confirmations
`/api/wallets/<wallet address>/transactions` API endpoint shows 0 confirmations for transactions that should be confirmed, which in result makes Desktop Wallet gray out transactions (eg. showing them as "pending") so you cannot use those funds.

**Expected Behavior**

It should show proper number of confirmations.

**Current Behavior**

It shows 0 confirmations for transactions that should be confirmed.

**Steps to Reproduce (for bugs)**

make an API call to` /api/wallets/<wallet address>/transactions` and see the response of transactions it shows 0 for them.

rainydio:
## Summary

Includes #3733 (should be merged first), fixes #3758 and #3759, and [comment] in #3741. All related to search and api.

Similarly to #3733 new set of methods `...joinBlock` are introduced to `TransactionHistoryService`. Joined block data is then used to calculate `confirmations` and `timestamp` in api response.

Implemented `address` criteria that searches by both `senderId` and `recipientId`. Note that it is `address`, not `addresses` which similar to other fields (`recipientId`, not `recipientIds`), but incompatible with v2.6.

## Checklist

- [ ] Tests
- [x] Ready to be merged

[comment]: https://github.com/ArkEcosystem/core/issues/3741#issuecomment-634043543


packages/core-database/src/transaction-filter.ts
```diff
@@ -32,8 +32,18 @@ export class TransactionFilter implements Contracts.Database.TransactionFilter {
...
+                    return handleOrCriteria(criteria.address!, async (c) => {
+                        return this.handleAddressCriteria(c);
+                    });
                 case "senderId":
-                    return handleOrCriteria(criteria.senderId!, (c) => this.handleSenderIdCriteria(c));
+                    return handleOrCriteria(criteria.senderId!, async (c) => {
+                        return this.handleSenderIdCriteria(c);
+                    });
+                case "recipientId":
+                    return handleOrCriteria(criteria.recipientId!, (c) => {
```
air1one, line 44:
```suggestion
                    return handleOrCriteria(criteria.recipientId!, async (c) => {
```

PR #3765, opened 2020-06-02
fix(core-transaction-pool): introduce CORE_RESET_POOL env variable

rainydio:
## Summary

Introduced new environment variable `CORE_RESET_POOL` which only resets pool. Unlike `CORE_RESET_DATABASE` that resets both main and pool databases.

## Checklist

- [x] Ready to be merged


faustbrian:
@rainydio failing tests

PR #3776, opened 2020-06-05
feat(core-manager): dispatch wallet events

sebastijankuzner:
## Summary

This PR introduces new WatcherWallet, which is inherited Wallet with ability to dispatch events on every attribute or property change. When using core-manager original wallet factory is replaced with custom factory which  return wallet proxies. 

## Checklist

- [x] Tests
- [x] Ready to be merged


packages/core-state/src/wallets/indexers/wallet-indexes.ts
```diff
@@ -43,13 +43,15 @@ export const registerIndexers = (app: Contracts.Kernel.Application): void => {
...
 
 export const registerFactories = (app: Contracts.Kernel.Application): void => {
-    app.bind(Container.Identifiers.WalletFactory).toFactory<Contracts.State.Wallet>(
-        (context: Container.interfaces.Context) => (address: string) =>
-            new Wallet(
-                address,
-                new Services.Attributes.AttributeMap(
-                    context.container.get<Services.Attributes.AttributeSet>(Container.Identifiers.WalletAttributes),
+    if (!app.isBound(Container.Identifiers.WalletFactory)) {
+        app.bind(Container.Identifiers.WalletFactory).toFactory<Contracts.State.Wallet>(
```
faustbrian, line 47: This should probably use `rebind` to avoid any issues because `core-manager` will usually be loaded after the `WalletFactory` has already been bound.

sebastijankuzner: I was thinking about it, but the problem is, because core-manager needs to be loaded right after core-logger-pino, otherwise the half of the logs (from classes created in register mehtod) are missing. We need to replace log service with LogServiceWrapper before other classes are created.

PR #3777, opened 2020-06-05
fix(core-api): set default orderBy value

rainydio:
## Summary

Order transactions and blocks in reverse chronological order by default.

## Checklist

- [ ] Ready to be merged


rainydio:
Ordering is done on `core-api` level which is incorrect, but can be implemented quickly. Ordering on `core-database` makes more sense, but isn't implemented yet.

PR #3804, opened 2020-06-16
refactor(core-api): use shorter default timeouts

alessiodf:
## Summary

This PR stops a user from opening and keeping open multiple long lived TCP connections to the public API server which is enabled by default. It prevents server resource starvation by implementing a shorter maximum timeout (default: 5 seconds) so connections will be closed if no request is received after that time.

### DIFF
--- a/packages/core-api/src/server.ts
+++ b/packages/core-api/src/server.ts
@@ -58,7 +58,9 @@ export class Server {
         this.server = new HapiServer(this.getServerOptions(optionsServer));
 
         const timeout: number = this.configuration.getRequired<number>("plugins.socketTimeout");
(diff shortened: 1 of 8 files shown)


packages/core-api/src/server.ts
```diff
@@ -56,6 +56,10 @@ export class Server {
     public async initialize(name: string, optionsServer: Types.JsonObject): Promise<void> {
         this.name = name;
         this.server = new HapiServer(this.getServerOptions(optionsServer));
+
+        const timeout: number = this.configuration.getRequired<number>("plugins.socketTimeout");
+        this.server.listener.timeout = this.server.listener.keepAliveTimeout = this.server.listener.headersTimeout = timeout;
```
faustbrian, line 61: Split this into 3 assignments with 3 lines, this is quite difficult to read.

PR #3812, opened 2020-06-18
chore(core-database): upgrade typeorm and pg packages

rainydio:
## Summary

Connection freezes on fresh Ubuntu 20.04, upgrading `typeorm` and `pg` packages fixes the issue.

## Checklist

- [x] Ready to be merged

### DIFF
--- a/packages/core-database/src/repositories/round-repository.ts
+++ b/packages/core-database/src/repositories/round-repository.ts
@@ -8,7 +8,7 @@ export class RoundRepository extends Repository<Round> {
     public async findById(id: string): Promise<Round[]> {
         return this.find({
             where: {
-                id,
+                round: id,
             },
         });
     }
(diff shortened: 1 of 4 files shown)


packages/core-database/src/repositories/round-repository.ts
```diff
@@ -8,7 +8,7 @@ export class RoundRepository extends Repository<Round> {
     public async findById(id: string): Promise<Round[]> {
         return this.find({
             where: {
-                id,
+                round: id,
```
rainydio, line 11: `Round` entity doesn't have `id` property. How the heck it worked before?

packages/core-database/src/repositories/transaction-repository.ts
```diff
@@ -116,7 +117,7 @@ export class TransactionRepository extends AbstractRepository<Transaction> {
             .addSelect('blocks.height as "blockHeight"')
             .addSelect('blocks.generatorPublicKey as "blockGeneratorPublicKey"')
             .addSelect('blocks.reward as "reward"')
-            .addFrom(Blocks.Block, "blocks")
```
rainydio, line 119: `Block.Block` is from `@arkecosystem/crypto` package. It's not `Block` entity. Previous version of `typeorm` somehow didn't cared.

Issue #3592, opened 2020-03-09
feat: expose to the api whether peer uses estimates or not


dated:
Currently the only way to find out if a peer uses estimates under the hood is by sending a request to it and check the responses meta object for the `totalCountIsEstimate` property.

```
"meta": {
  "totalCountIsEstimate": true,
}
```

This information should also be included in the `peers` api response, so that peers can be easily filtered by that property instead of having to query each peer individually if one with need to assemble a list of such peers.

This could then also be implemented into into the peer discovery package, `@arkecosytem/typescript-peers`.

ArkEcosystemBot:
**Thanks for opening this issue!** A maintainer will review this in the next few days and explicitly select labels so you know what's going on.

_If no reviewer appears after a week, a reminder will be sent out._

ArkEcosystemBot:
This issue has been closed. If you wish to re-open it please provide additional information.

ArkEcosystemBot:
This issue has now been re-opened. If applicable please provide additional information as requested by one of the reviewers.

PR #3825, opened 2020-06-23
feat(core-p2p, core-api): include estimateTotalCount into peers response

sebastijankuzner:
## Summary

This PR solves issue #3592. 

P2P API includes estimateTotalCount boolean property based on the node configuration achieved via CORE_API_NO_ESTIMATED_TOTAL_COUNT flag. 

Core-api peers endpoint now returns plugins with the estimateTotalCount property inside @arkecosystem/core-api plugin options.

Part of the response: 

```
        {
            "ip": "192.168.64.8",
            "port": "4000",
            "ports": {},
            "version": "3.0.0-next.0",
            "height": 85288,
            "latency": 31,
            "plugins": {
                "@arkecosystem/core-api": {
                    "enabled": true,
                    "port": 4003,
                    "estimateTotalCount": false
                },
                "@arkecosystem/core-webhooks": {
                    "enabled": true,
                    "port": 4004
                }
            }
        }
```

## Checklist

- [ ] Tests _(TBD)_
- [X] Ready to be merged

### DIFF
--- a/packages/core-p2p/src/socket-server/utils/get-peer-config.ts
+++ b/packages/core-p2p/src/socket-server/utils/get-peer-config.ts
@@ -25,6 +25,10 @@ const transformPlugins = (plugins: PluginConfig[]): Contracts.P2P.PeerPlugins =>
             enabled: true, // default to true because "enabled" flag is in different place based on which plugin
             port,
         };
+
+        if (name.includes("core-api")) {
+            result[name].estimateTotalCount = pluginConfig.options.options.estimateTotalCount;
+        }
     }
 
     return result;
--- a/packages/core-api/src/resources/peer.ts
+++ b/packages/core-api/src/resources/peer.ts
@@ -30,6 +30,7 @@ export class PeerResource implements Resource {
             version: resource.version,
             height: resource.state ? resource.state.height : resource.height,
             latency: resource.latency,
+            plugins: resource.plugins,
         };
     }
 }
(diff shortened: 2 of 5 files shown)


packages/core-p2p/src/socket-server/utils/get-peer-config.ts
```diff
@@ -25,6 +25,10 @@ const transformPlugins = (plugins: PluginConfig[]): Contracts.P2P.PeerPlugins =>
             enabled: true, // default to true because "enabled" flag is in different place based on which plugin
             port,
         };
+
+        if (name.includes("core-api")) {
+            result[name].estimateTotalCount = pluginConfig.options?.options?.estimateTotalCount || true;
```
faustbrian, line 30: Why would we need the `|| true` since `estimateTotalCount` is always present from the defaults?

sebastijankuzner: If you take a look in packages/core-api/src/controllers/controller.ts it states this is an optional property. 

`
     estimateTotalCount: this.apiConfiguration.getOptional<boolean>("options.estimateTotalCount", true),
`

faustbrian: Yes, but that doesn't mean that it is correct. If you look at the `defaults.ts` file you will see that it's always present from the defaults because we merge the user and default configuration so unless someone modifies the defaults and rebuilds core from source I can't think of a situation where this would not be specified.

PR #3844, opened 2020-06-29
fix(core-transactions): stream transactions during bootstrap

rainydio:
## Summary

Bootstrap implementations that used reader were limited to `1000000000` transactions only. There was a comment mentioning that and suggesting to use streams.

* `streamByCriteria` method was added to transaction history service.
* `getTransactionReader` was removed from base transaction handler.
* Every bootstrap implementation that used reader was updated.
* Entity handler received bootstrap implementation that it was missing.
* Multi-signature handlers have different implementations and now correctly load transactions by version.
* Added new attribute `multiSignature.legacy` set by legacy multi-signature handler.

## Checklist

- [x] Tests
- [x] Ready to be merged


rainydio:
It appears as coverage errors are false-positives

PR #3884, opened 2020-07-08
test(core-p2p): increase coverage to 100%

sebastijankuzner:
## Summary

This PR increases core-p2p package to 100%.

## Checklist

- [x] Tests
- [x] Ready to be merged


packages/core-p2p/src/hapi-nes/socket.ts
```diff
@@ -195,9 +179,9 @@ export class Socket {
             message.id = request.id;
         }
 
-        if (err.path) {
-            message.path = err.path;
-        }
+        // if (err.path) {
```
air1one, line 182: why is that commented ? does it need to be removed or uncommented ?

PR #3898, opened 2020-07-16
feat(core-p2p): implement peer blacklisting

sebastijankuzner:
## Summary

This PR implements missing peer blacklisting. Additionally isBlacklisted function was added to core-kernel.

## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged

### DIFF
--- a/packages/core-kernel/src/utils/is-blacklisted.ts
+++ b/packages/core-kernel/src/utils/is-blacklisted.ts
@@ -1,0 +1,18 @@
+import nm from "nanomatch";
+
+// todo: review the implementation
+export const isBlacklisted = (whitelist: string[], remoteAddress: string): boolean => {
+    if (!Array.isArray(whitelist) || !whitelist.length) {
+        return false;
+    }
+
+    for (const ip of whitelist) {
+        try {
+            if (nm.isMatch(remoteAddress, ip)) {
+                return true;
+            }
+        } catch {}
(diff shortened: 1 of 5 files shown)


packages/core-kernel/src/utils/is-blacklisted.ts
```diff
@@ -0,0 +1,18 @@
+import nm from "nanomatch";
+
+// todo: review the implementation
+export const isBlacklisted = (whitelist: string[], remoteAddress: string): boolean => {
```
air1one, line 4: Is it exactly the same code as `isWhitelisted` ? If it is can you instead make a generic one like `matchesIPAddressList` ?

sebastijankuzner: No it is not exactly the same. IsBlackilsted is returning false by default if no list is provided, but isWhitelisted return true. I am aware of the duplicated for loop in both files, this can potentially be extracted.

air1one: ah ok sorry no don't worry we can keep it this way but then can you change `whitelist` parameter to `blacklist` ? :sweat_smile:

sebastijankuzner: Of course. I have mislooked it.

PR #3902, opened 2020-07-21
chore(core-database): blocksInCurrentRound and forgingDelegates are arrays

sebastijankuzner:
## Summary

This PR set blocksInCurrentRound and forgingDelegates as array and remove possible undefined value. This fixes runtime errors when setting length value to undefined array.

## Checklist
- [X] Ready to be merged


packages/core-database/src/database-service.ts
```diff
@@ -167,7 +164,7 @@ export class DatabaseService {
                     await this.saveRound(this.dposState.getRoundDelegates());
 
                     // ! set it to empty array and why it can be undefined at all?
```
air1one, line 166:
```suggestion
```

Issue #3891, opened 2020-07-13
Some transactions in `/transactions/${id}` don't return timestamp


air1one:
This is a follow-up issue on https://github.com/ArkEcosystem/core/issues/3759 which was fixed by https://github.com/ArkEcosystem/core/pull/3761 , specifically on `/transactions` endpoints.

But it seems like the `/transactions/${id}` endpoint also has the issue (some transactions don't return timestamp on it).

PR #3925, opened 2020-08-03
fix(core-api): fix transactions/${id} confirmations and timestamp fields

rainydio:
## Summary

Quick fix of #3891 until I finish with #3911

## Checklist

- [ ] Ready to be merged


rainydio:
There is some race in `__tests__/functional/transaction-forging/htlc-lock.test.ts` and I disabled parallel execution for `core-p2p` unit tests.

Issue #3929, opened 2020-08-04
Core v3 fails to start without an `.env` file


adrian69:
It seems v3 doesn't honor exported env vars and fails to start without an `.env` file.

**Expected Behavior**
It should use exported env vars (`process.env`):
```
bash-5.0$ env | grep CORE
CORE_API_PORT=4003
CORE_DB_DATABASE=core_devnet
CORE_DB_USERNAME=node
CORE_P2P_HOST=0.0.0.0
CORE_EXCHANGE_JSON_RPC_HOST=0.0.0.0
CORE_DB_PASSWORD=password
CORE_LOG_LEVEL=debug
CORE_WEBHOOKS_PORT=4004
CORE_LOG_LEVEL_FILE=info
CORE_P2P_PORT=4002
CORE_DB_HOST=postgres-devnet
CORE_ELASTICSEARCH_HOST=elastic-devnet
CORE_API_HOST=0.0.0.0
CORE_ELASTICSEARCH_PORT=9200
CORE_WEBHOOKS_HOST=0.0.0.0
CORE_EXCHANGE_JSON_RPC_PORT=8080
```
**Current Behavior**

It simply fails to start if there is no `.env` file with the following error:

```
Error: [ERROR] Unable to load the environment file. ENOENT: no such file or directory, open '/home/node/.config/ark-core/devnet/.env'
    at Fatal.render (/home/node/.config/yarn/global/node_modules/@arkecosystem/core-cli/dist/components/fatal.js:22:15)
    at ComponentFactory.fatal (/home/node/.config/yarn/global/node_modules/@arkecosystem/core-cli/dist/component-factory.js:118:54)
    at Command.run (/home/node/.config/yarn/global/node_modules/@arkecosystem/core-cli/dist/commands/command.js:124:29)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
 ✖  unhandledRejection (a promise was rejected but not handled)  Error: [ERROR] Unable to load the environment file. ENOENT: no such file or directory, open '/home/node/.config/ark-core/devnet/.env'
    at Fatal.render (/home/node/.config/yarn/global/node_modules/@arkecosystem/core-cli/dist/components/fatal.js:22:15)
    at ComponentFactory.fatal (/home/node/.config/yarn/global/node_modules/@arkecosystem/core-cli/dist/component-factory.js:118:54)
[...]

PR #3934, opened 2020-08-05
fix(core-kernel): boot without .env file

sebastijankuzner:
## Summary

This PR solves  #3929

Changes:
1. If .env file is not present program log alert and starts
1. If variable is set in process and in .env, variable from process (system) is taken

## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged


faustbrian:
`If .env file is not present program log alert and starts` was discussed in Slack and I said there we won't allow booting without that file because it contains the defaults for every network.

sebastijankuzner:
In this case I am closing pull request.

PR #3943, opened 2020-08-09
refactor(core-utils): use JSON.parse instead of fast-json-parse

faustbrian:
Removes the `fast-json-parse` module because it's no longer necessary.

### DIFF
--- a/packages/core-utils/src/httpie.ts
+++ b/packages/core-utils/src/httpie.ts
@@ -21,7 +21,7 @@ export class HttpieError extends Error {
                 Object.defineProperty(this, "response", {
                     enumerable: false,
                     value: {
-                        body: JSON.parse(error.response.body || {}),
+                        body: JSON.parse(error.response.body),
                         headers: error.response.headers,
                         status: error.response.statusCode,
                     },
@@ -94,7 +94,7 @@ class Httpie {
             const { body, headers, statusCode } = await got[method](url, opts);
 
             return {
-                body: JSON.parse(body || {}),
+                body: JSON.parse(body),
                 headers,
                 status: statusCode,
             };
--- a/packages/core-utils/src/httpie.ts
+++ b/packages/core-utils/src/httpie.ts
@@ -1,6 +1,5 @@
 // tslint:disable: max-classes-per-file
 
-import parseJSON from "fast-json-parse";
 import got from "got";
 
 export class HttpieError extends Error {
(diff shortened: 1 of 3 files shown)


packages/core-utils/src/httpie.ts
```diff
@@ -91,7 +94,7 @@ class Httpie {
             const { body, headers, statusCode } = await got[method](url, opts);
 
             return {
-                body: parseJSON(body).value,
+                body: JSON.parse(body || {}),
```
alessiodf, line 97:
```suggestion
                body: JSON.parse(body || "{}"),
```

`JSON.parse` expects a string, not an object, otherwise you'll get a `SyntaxError` as it will try to parse `[object Object]` from `{}.toString()`.

packages/core-utils/src/httpie.ts
```diff
@@ -18,14 +17,18 @@ export class HttpieError extends Error {
...
-                    body: parseJSON(error.response.body).value,
-                    headers: error.response.headers,
-                    status: error.response.statusCode,
-                },
-            });
+            try {
+                Object.defineProperty(this, "response", {
+                    enumerable: false,
+                    value: {
+                        body: JSON.parse(error.response.body || {}),
```
alessiodf, line 24:
```suggestion
                        body: JSON.parse(error.response.body || "{}"),
```

`JSON.parse` expects a string, not an object, otherwise you'll get a `SyntaxError` as it will try to parse `[object Object]` from `{}.toString()`.

packages/core-utils/src/httpie.ts
```diff
@@ -94,7 +94,7 @@ class Httpie {
             const { body, headers, statusCode } = await got[method](url, opts);
 
             return {
-                body: JSON.parse(body || {}),
+                body: JSON.parse(body),
```
alessiodf, line 97:
```suggestion
                body: JSON.parse(body || "{}"),
```

In its current state it will throw if `body` is empty or `undefined` since parsing will fail, which is always the case when the response is to a `HEAD` request. In real terms this means port pinging in the peer communicator will set all port values to `-1` in the `/api/peers` endpoint.

packages/core-utils/src/httpie.ts
```diff
@@ -21,7 +21,7 @@ export class HttpieError extends Error {
                 Object.defineProperty(this, "response", {
                     enumerable: false,
                     value: {
-                        body: JSON.parse(error.response.body || {}),
+                        body: JSON.parse(error.response.body),
```
alessiodf, line 24:
```suggestion
                        body: JSON.parse(error.response.body || "{}"),
```
Same.

PR #3960, opened 2020-08-13
fix(core-p2p): out of memory test error

rainydio:
## Summary

Tracking down `core-p2p` out of memory error in tests.


rainydio:
It appears that problem mysteriously disappeared.

PR #3969, opened 2020-08-20
refactor(core-blockchain): initialize blockchain when resolved first time

rainydio:
## Summary

`Blockchain` class was resolved during `register`, which chain-resolved wallet repository and all wallet indices which were registered, any wallet index that is registered afterwards wasn't available.

The reason why it was resolved is because `Blockchain` class needed additional constructor logic to setup state machine. I marked `initialize` function with `@postConstruct` decorator and removed direct `resolve` and `initialize` call from `register.

Now plugin that adds wallet index (e.g. `core-magistrate-transactions`) doesn't have to be registered before `core-blockchain`.

## Checklist

- [x] Tests
- [ ] Ready to be merged

### DIFF
--- a/__tests__/unit/core-blockchain/blockchain.test.ts
+++ b/__tests__/unit/core-blockchain/blockchain.test.ts
@@ -12,6 +12,7 @@ import delay from "delay";
 describe("Blockchain", () => {
     let sandbox: Sandbox;
 
+    const configuration: any = {};
     const logService: any = {};
     const stateStore: any = {};
     const databaseService: any = {};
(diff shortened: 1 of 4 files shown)


rainydio:
I tried to reuse `core-api` integration test `setup` for `core-magistrate-api` by adding `...plugins` argument to `core-api` setup. Apparently not only `core-blockchain`, but also `core-forger` and `core-api` prevent me from doing that. Not gonna pursue that any further.

This in itself is worth merging anyway.

PR #3974, opened 2020-08-22
chore(crypto): add publickey verification

sleepdefic1t:
## Summary

Current master/develop only checks for valid hex during PublicKey validation, but does not verify the correctness of a publicKey.

This can lead to the unexpected creation of an unaccessible but technically-"valid" Address.

This PR recommends the following:
- add a `verify` method to `class PublicKey` using the already-present `secp256k1` module
- modify `Address.fromPublicKey` to drop hex-validation in favor of `PublicKey.verify`
- modify `PublicKey.fromMultiSignatureAsset` to drop hex-validation in favor of `PublicKey.verify`
- add relevant unit-tests

## Checklist

- [x] Tests
- [x] Ready to be merged

## Additional Comments

- No change of import modules needed
- maintains resolution of circular-dep fixes in c84b976

There are other areas that would benefit from the `PublicKey.verify` method; however, no additional checks are being added at this time.

It should also be stated that a developer should use this new method whenever:
- the correctness of a publicKey cannot be guaranteed
- the publicKey is NOT already being checked (_`.verify(sig, pubKey, msg)`_)

### DIFF
--- a/__tests__/unit/core-state/__utils__/build-delegate-and-vote-balances.ts
+++ b/__tests__/unit/core-state/__utils__/build-delegate-and-vote-balances.ts
@@ -4,15 +4,27 @@ import { SATOSHI } from "@packages/crypto/src/constants";
 
 export const buildDelegateAndVoteWallets = (numberDelegates: number, walletRepo: WalletRepository): Wallet[] => {
     const delegates: Wallet[] = [];
+    const delegateKeys: string[] = [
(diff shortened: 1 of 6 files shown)


sleepdefic1t:
@air1one @faustbrian 

No idea what's up with [functional/htlc-lock tests step:10:208](https://github.com/ArkEcosystem/core/pull/3974/checks?check_run_id=1022811270#step:10:208), doesn't appear to be anything I touched.

air1one (approved):

PR #3985, opened 2020-08-28
chore(core-state): autoIndex option

sebastijankuzner:
## Summary

Implements autoIndex option on indexes. When autoIndex is enabled, WalletRepository.index(wallet) method will remove all known index entries by wallet and will reindex wallet, otherwise developer is responsible to manually use index, set or delete methods directly on index. 

Disabling autoIndex option is useful in cases when index hold a lot of entries for the wallet and manual control over index will increase node performance. 

## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged


rainydio (approved):
All in all it makes perfect sense. If index isn't complicated (most cases), then you can rely on automatic re-indexing. Otherwise you'll have to call `getIndex` and do everything manually and carefully.

air1one (approved):

PR #4070, opened 2020-09-30
refactor: remove import from dist or src

air1one:
## Checklist

- [ ] Tests _(if necessary)_
- [x] Ready to be merged


packages/core-snapshots/src/contracts/database.ts
```diff
@@ -1,5 +1,5 @@
 import { Contracts } from "@arkecosystem/core-kernel";
-import { Meta, Options } from "@arkecosystem/core-snapshots/src/contracts/index";
+import { Meta, Options } from "./index";
```
faustbrian, line 2: Should avoid imports from index files since they could lead to circular references in the import tree.

faustbrian (approved):

PR #4108, opened 2020-10-19
feat(core-state): sync wallet to database

rainydio:
## Sync wallets to database table.

`WalletSyncService` was added to `core-state`. In its `boot` method it subscribes to various state, wallet, and block events to keep track of addresses that were changed and sync them to `wallets` table in PostgreSQL.

### Syncing

The `wallets` table is truncated and rebuilt each time after bootstrap. PostgreSQL has network protocol limitation on how many parameters can be sent with each query, so it's done in batches of 6000 wallets. It would take a little over 20 queries to populate it on mainnet.

Wallet events are used to keep track of wallets that were updated. After block is applied or reverted updated wallets are synced to `wallets` table.

### Atomicity

Wallets aren't synced atomically with blocks. It's possible for `transactions` table to have an entry that has `recipient_id` which does not have corresponding entry in `wallets` table. This isn't acceptable for production version of products that attempts to use this table, but should be enough for development.

### Disposing

Events are fired synchronously, but sync is asynchronous operations. Errors thrown invoke `Application.terminate` which doesn't call `process.exit` with error code. It attempts to gracefully shutdown and that doesn't work.


air1one (approved):

packages/core-state/src/defaults.ts
```diff
@@ -3,4 +3,7 @@ export const defaults = {
         maxLastBlocks: 100,
         maxLastTransactionIds: 10000,
     },
+    walletSync: {
+        enabled: false,
```
faustbrian, line 7: This needs an environment variable to be enabled.

air1one: oh didnt see that sorry, @rainydio you can fix this in another PR

rainydio: Implemented in #4127

Issue #4038, opened 2020-09-16
Persist wallets to database


air1one:
Persist wallets to database so that they can be queried directly on the db - would help improve explorer efficiency.

Use a flag so it is disabled by default, enabled only when asked.

PR #4127, opened 2020-10-23
feat(core-state): enable wallet sync though environment variable

rainydio:
Enable wallet sync (#4038 #4108) through environment variable.

### DIFF
--- a/packages/core-state/src/defaults.ts
+++ b/packages/core-state/src/defaults.ts
@@ -4,6 +4,6 @@ export const defaults = {
         maxLastTransactionIds: 10000,
     },
     walletSync: {
-        enabled: process.env.CORE_WALLET_SYNC_ENABLED === "true",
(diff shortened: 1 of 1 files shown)


packages/core-state/src/defaults.ts
```diff
@@ -4,6 +4,6 @@ export const defaults = {
         maxLastTransactionIds: 10000,
     },
     walletSync: {
-        enabled: false,
+        enabled: process.env.CORE_WALLET_SYNC_ENABLED === "true",
```
faustbrian, line 7: Follow the convention of the rest of core https://github.com/ArkEcosystem/core/blob/develop/packages/core-api/src/defaults.ts#L4

rainydio: `enabled: !!process.env.CORE_WALLET_SYNC_ENABLED`?

faustbrian (approved):

PR #4138, opened 2020-10-30
fix(core): network:generate save to default location

sebastijankuzner:
## Summary

Network:generate
- Save generated files to default location. 
- Ignore prompts when all flags are passed

Don't resolve execute method on :run commands to prevent exiting the process.

## Checklist
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/__tests__/unit/core/commands/core-run.test.ts
+++ b/__tests__/unit/core/commands/core-run.test.ts
@@ -32,7 +32,13 @@ describe("RunCommand", () => {
         const spyBootstrap = jest.spyOn(app, "bootstrap").mockImplementation(undefined);
         const spyBoot = jest.spyOn(app, "boot").mockImplementation(undefined);
 
-        await executeCommand(Command);
+        executeCommand(Command);
+
+        await new Promise((resolve) => {
+            setTimeout(() => {
+                resolve();
+            }, 50);
+        });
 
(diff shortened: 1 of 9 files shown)


ItsANameToo (commented):
https://github.com/ArkEcosystem/core/blob/2a6dd9bf4b31ddac96c2ef9143237095498a2505/packages/core/src/commands/network-generate.ts#L99-L101 @sebastijankuzner this check fails when `false` is passed (which is done for the `distribute` flag), which makes it impossible to generate a network config without prompt as I can't get into that `if`. Would it be possible to add a `--force` flag instead that disables the prompt and simply uses whatever was provided?

sebastijankuzner:
> https://github.com/ArkEcosystem/core/blob/2a6dd9bf4b31ddac96c2ef9143237095498a2505/packages/core/src/commands/network-generate.ts#L99-L101
> @sebastijankuzner this check fails when `false` is passed (which is done for the `distribute` flag), which makes it impossible to generate a network config without prompt as I can't get into that `if`. Would it be possible to add a `--force` flag instead that disables the prompt and simply uses whatever was provided?

--force is one option, another is that we check if value === undefined.

faustbrian:
Implement `--force` so people can just mute the UI whenever they want even if they omit a value.

PR #4146, opened 2020-11-03
fix(core-database): set adequate autovacuum settings

rainydio:
## Summary

Autovacuuming (and `ANALYZE`) wasn't performed frequently enough because default values of `autovacuum_vacuum_scale_factor` and `autovacuum_analyze_scale_factor` are to high.

## Checklist

- [x] Ready to be merged


faustbrian:
We could probably also include the `rounds` table. Did some more testing there and also saved me around 200ms on average when the table was vacuumed.

faustbrian (approved):

PR #4148, opened 2020-11-04
feat(core): add additional flags into network:generate command

sebastijankuzner:
## Summary

Add additional flags into network:generate command. 

- force (force generation without all flags defined)
- configPath (custom path configuration)
- overwriteConfig (overwrite existing configuration)
- peers (set list of peers)
- fees settings
- .env additional settings

## Checklist
- [x] Tests
- [x] Ready to be merged


packages/core/src/commands/network-generate.ts
```diff
@@ -46,16 +109,59 @@ export class Command extends Commands.Command {
...
+        { name: "coreDBPassword", description: "Core database password.", schema: Joi.string(), required: false, promptType: "" },
+        { name: "coreDBDatabase", description: "Core database database.", schema: Joi.string(), required: false, promptType: "" },
+
+        { name: "coreP2PPort", description: "Core P2P port.", schema: Joi.number(), required: false, promptType: "", default: 4000 },
+        { name: "coreAPIPort", description: "Core API port.", schema: Joi.number(), required: false, promptType: "", default: 4003 },
+        { name: "coreWebhooksPort", description: "Core Webhooks port.", schema: Joi.number(), required: false, promptType: "", default: 4004 },
+        { name: "coreMonitorPort", description: "Core Webhooks port.", schema: Joi.number(), required: false, promptType: "", default: 4005 },
+
+        // Peers
+        { name: "peers", description: "Peers IP addresses, separated with comma.", schema: Joi.string().allow(""), required: false, promptType: "", default: "127.0.0.1" },
```
air1one, line 157: `Joi.string().allow("")` does this mean that we only allow empty string ? :thinking: or am I missing something

sebastijankuzner: No. By default empty flags are now allowed (--peers=""). This way user can set empty flag to skip default option when he want empty list. It also works when user wants to write custom IP addresses, separated with comma.

air1one: oh ok got it thanks

air1one (approved):

PR #4175, opened 2020-11-18
refactor(core-manager): more granular filters for log:log action

sebastijankuzner:
## Summary

Modify code to include following changes:

- limit log respone to 100 lines and return array of line objects (timestamp, level, content)
- search by time
- search by content
- search by log level

## Checklist
- [x] Test
- [x] Ready to be merged


faustbrian (changes requested):
Each line that is returned should be represented as an object `{ timestamp, level, content }` so that a consumer can easily make use of the data without any further processing. I would also request to rename the action to `log:search` in a separate pull-request.

faustbrian (changes requested):
Please branch off the renaming of the command into a separate pull-request as requested in the previous comment.

faustbrian:
@sebastijankuzner After the renaming of the action the documentation needs to be updated.

PR #4177, opened 2020-11-19
fix(core-manager): respect channel version

sebastijankuzner:
## Summary

Respect channel version to resolve latest version on NPM. 

## Checklist
- [x] Test
- [x] Ready to be merged


packages/core-manager/src/actions/info-core-version.ts
```diff
@@ -18,6 +18,21 @@ export class Action implements Actions.Action {
...
 
     private async getLatestVersion(): Promise<string> {
-        return latestVersion("@arkecosystem/core");
+        return latestVersion("@arkecosystem/core", {
+            version: this.getChannel(),
+        });
+    }
+
+    private getChannel(): string {
+        const channels: string[] = ["alpha", "beta", "rc", "next"];
```
faustbrian, line 27: Only `next` and `latest` are used. All of the other channels have been deprecated.

Issue #4058, opened 2020-09-27
ipv6 not working


alessiodf:
Although Core's IP schema seems to support IPv6 addresses as well as IPv4 addresses, in practice it doesn't work because Core considers most IPv6 addresses to be part of the same subnet (~specifically `::` which is wrong~ often `::` but not always) so you theoretically cannot add much more than 5 IPv6 nodes to any network.

However, it all breaks down totally because to connect to IPv6 sockets you have to enclose the address in square brackets, but Core does not do this. So to recap, although the schema supports IPv6, Core cannot successfully connect to IPv6 addresses, and even if it could, it uses a faulty CIDR check that does not process IPv6 addresses properly.

This affects both Core 2.7 and 3.0 and although I don't expect this to be fixed in 2.7, I hope it will be addressed in 3.0.

ArkEcosystemBot:
**Thanks for opening this issue!** A maintainer will review this in the next few days and explicitly select labels so you know what's going on.

_If no reviewer appears after a week, a reminder will be sent out._

PR #4196, opened 2020-11-26
feat: P2P IPv6 support 

sebastijankuzner:
## Summary

Solves issue #4058

Changes:
- use IPv4 and IPv6 validation in schemas
- all IPv6 are set and used without brackets in pure form
- works with both IPv4 and IPv6 nodes simultaneously

Setup:
- in .env file set CORE_P2P_HOST=:: to listen on both IPv4 and IPv6 address

## Checklist
- [x] Tests
- [x] Ready to be merged


faustbrian (changes requested):
Comment left on clickup

PR #4212, opened 2020-12-04
feat(core-manager): log default paths

sebastijankuzner:
## Summary

Log default locations on core boot. 

## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged


faustbrian:
Why is this needed?

sebastijankuzner:
> Why is this needed?

To inform users and devs where they can find logs, data, etc.

faustbrian:
https://github.com/ArkEcosystem/core/blob/develop/packages/core/src/commands/env-paths.ts already exists for that and this isn't information specific to this plugin.

PR #4222, opened 2020-12-08
deps(core): include core-manager dependency

sebastijankuzner:
## Summary

Include core-manager dependency.

## Checklist

- [x] Ready to be merged


faustbrian:
This won't be included by default because it is an optional package.

PR #4232, opened 2020-12-16
fix(core-p2p): do not terminate when app not ready

### LINKED ISSUE #4124: Not getting all blocks broadcast quick enough on restart
From discussion in community slack channel :

> It took 6-7min for my upgraded node to start receiving every block… It would get one or 2 after catch up, then get none, catch up again, and same repeats
> It’s like the peers ignore me after being offline(restarting) for 15sec

> I think it's likely to do with how long Nes waits to attempt a reconnection (if at all), whereas SocketCluster did it fairly quickly so any transient loss was quickly restored.

Need to fix this so that we receive all blocks broadcast on restart.

air1one:
## Summary

Resolves #4124 which was caused by socket server disconnecting on error "App is not ready".

Defined a custom status code to be used when we want to send an explicit error back (and not terminate the socket).

## Checklist

- [ ] Tests _(if necessary)_
- [x] Ready to be merged


faustbrian (approved):

faustbrian:
🤨

PR #4257, opened 2021-01-14
chore: parse numberic values from ENV variables

sebastijankuzner:
## Summary

Parse number form ENV variables, that are expected to be numeric.

## Checklist

- [x] Ready to be merged


faustbrian (changes requested):
Use https://github.com/ArkEcosystem/core/blob/develop/packages/core-kernel/src/providers/service-provider.ts#L154-L162 instead of manually casting values which is error prone.

PR #4297, opened 2021-02-02
fix(core-p2p): do not consume when just checking rate limit

air1one:
## Summary

In outgoing rate limiter (for throttling) it is wrong to consume 1 point of the rate limiter when we are just checking rate limit (not actually sending message).
Instead consume only when we know we do not hit rate limit.

## Checklist

- [ ] Tests _(if necessary)_
- [x] Ready to be merged

### DIFF
--- a/packages/core-p2p/src/peer-communicator.ts
+++ b/packages/core-p2p/src/peer-communicator.ts
@@ -289,12 +289,16 @@ export class PeerCommunicator implements Contracts.P2P.PeerCommunicator {
 
     private async throttle(peer: Contracts.P2P.Peer, event: string): Promise<void> {
         const msBeforeReCheck = 1000;
-        while (await this.outgoingRateLimiter.hasExceededRateLimit(peer.ip, event)) {
+        while (await this.outgoingRateLimiter.hasExceededRateLimitNoConsume(peer.ip, event)) {
             this.logger.debug(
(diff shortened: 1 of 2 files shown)


packages/core-p2p/src/peer-communicator.ts
```diff
@@ -289,12 +289,16 @@ export class PeerCommunicator implements Contracts.P2P.PeerCommunicator {
...
         const msBeforeReCheck = 1000;
-        while (await this.outgoingRateLimiter.hasExceededRateLimit(peer.ip, event)) {
+        while (await this.outgoingRateLimiter.hasExceededRateLimitNoConsume(peer.ip, event)) {
             this.logger.debug(
                 `Throttling outgoing requests to ${peer.ip}/${event} to avoid triggering their rate limit`,
             );
             await delay(msBeforeReCheck);
         }
+        try {
+            await this.outgoingRateLimiter.hasExceededRateLimit(peer.ip, event);
```
rainydio, line 299: `this.outgoingRateLimiter.hasExceededRateLimit` [doesn't throw]. It returns either `true` or `false`.

[doesn't throw]: https://github.com/ArkEcosystem/core/blob/develop/packages/core-p2p/src/rate-limiter.ts#L41-L57

PR #4298, opened 2021-02-02
feat(core-manager): support ZIP archive on `log.download`

sebastijankuzner:
## Summary

Support ZIP archive for archived logs.

Added `CORE_MONITOR_ARCHIVE_FORMAT` env variable with options:
- zip
- gz

## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged


packages/core-manager/src/workers/actions/generate-log-zip.ts
```diff
@@ -0,0 +1,41 @@
...
+
+        const archive = archiver("zip", {
+            zlib: { level: 9 },
+        });
+        archive.pipe(writeStream);
+
+        archive.append(
+            readStream.pipe(
+                new LogTransformStream({
+                    writableObjectMode: true,
```
rainydio, line 22: Why this option is present here, but not in `GenerateLogGz`?

packages/core-manager/src/workers/actions/streams/iterator-to-stream.ts
```diff
@@ -0,0 +1,12 @@
...
+export class IteratorToStream extends Readable {
+    public constructor(private readonly iterator: IterableIterator<any>) {
+        super({ objectMode: true });
+    }
+
+    public _read(size: number) {
+        const item = this.iterator.next();
+        this.push(item.done ? null : item.value);
+    }
+}
```
rainydio, line 12: There is standard [`Readable.from`] method that creates readable stream from iterator.

[`Readable.from`]: https://nodejs.org/api/stream.html#stream_stream_readable_from_iterable_options

packages/core-manager/src/workers/actions/generate-log-zip.ts
sebastijankuzner: Zlib stream supports object mode, while archiver (zip) doesn't.

PR #4299, opened 2021-02-03
chore(core-p2p): queue postTransactions requests

air1one:
## Summary

Use a queue for sending postTransactions requests, so that we send them sequentially and allowing us to better control throttling according to rate limit (adding a slight delay between each request).

## Checklist

- [ ] Tests _(if necessary)_
- [ ] Ready to be merged

### DIFF
--- a/packages/core-p2p/src/peer-communicator.ts
+++ b/packages/core-p2p/src/peer-communicator.ts
@@ -31,6 +31,9 @@ export class PeerCommunicator implements Contracts.P2P.PeerCommunicator {
     @Container.inject(Container.Identifiers.LogService)
     private readonly logger!: Contracts.Kernel.Logger;
 
+    @Container.inject(Container.Identifiers.QueueFactory)
+    private readonly createQueue!: Types.QueueFactory;
+
     private outgoingRateLimiter!: RateLimiter;
 
     private postTransactionsQueueByIp: Map<string, Contracts.Kernel.Queue> = new Map();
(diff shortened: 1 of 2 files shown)


packages/core-p2p/src/peer-communicator.ts
```diff
@@ -59,9 +65,27 @@ export class PeerCommunicator implements Contracts.P2P.PeerCommunicator {
...
-    public async postTransactions(peer: Contracts.P2P.Peer, transactions: Buffer[]): Promise<any> {
+    public async postTransactions(peer: Contracts.P2P.Peer, transactions: Buffer[]): Promise<void> {
         const postTransactionsTimeout = 10000;
-        return this.emit(peer, "p2p.transactions.postTransactions", { transactions }, postTransactionsTimeout);
+        const postTransactionsRateLimit = this.configuration.getOptional<number>("rateLimitPostTransactions", 25);
+
+        if (!this.postTransactionsQueueByIp.get(peer.ip)) {
+            this.postTransactionsQueueByIp.set(
+                peer.ip,
+                await this.app.get<Types.QueueFactory>(Container.Identifiers.QueueFactory)()
```
rainydio, line 75: Injecting queue factory?

```typescript
@Container.inject(Container.Identifiers.QueueFactory)
private readonly createQueue!: Types.QueueFactory;
```

air1one: yeah you're right will refactor

PR #4353, opened 2021-03-08
refactor(core-state): split DatabaseInteractions

sebastijankuzner:
## Summary

Split DatabaseInteraction into three modules:

- RoundState, which is responsible for handling current round state and store or revert rounds to database
- DatabaseInterceptor, which checks items in store first before forwarding calls to database
- DatabaseInteraction, which is responsible for handling blockApply and blockRevert with related stores

This PR is fundamental for further changes which will include:
- handling database errors on roundState
- simplify block handling logic 
- DatabaseInterceptor will take over a role of cache in front of DatabaseService

## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged


faustbrian:
Marking this as draft because it won't be merged into 3.0 but might go into a later release.

PR #4415, opened 2021-06-10
fix(install-next): remove bad line

sleepdefic1t:
## Summary

`ark config:cli --channel=next` causes an early exit bc we're already on `@arkecosystem/core@next`.

Removing the following line resolves this issue and allows successful completion of setup:
https://github.com/ArkEcosystem/core/blob/60bf124f393c908db23cec5b3bd7cd1d497653f7/install-next.sh#L227

## Checklist

- [x] Ready to be merged


adrian69:
Can only confirm it's true :trollface:

air1one (approved):

PR #4424, opened 2021-06-18
fix(core): plugin:install/remove/update commands

sebastijankuzner:
## Summary

Resolve package name argument for commands:
- plugin:install
- plugin:update
- plugin:remove

Improve source providers logic.

## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged


rainydio:
I have a problem with the way `CORE_PLUGINS_PATH` is used. I believe [`Environement.getPaths()`] is supposed to handle path resolutions. It already handles `CORE_PATH_CONFIG` and `CORE_PATH_DATA` variables. I think it should similarly handle `CORE_PATH_PLUGINS` variable (and fallback to constructing default from `data/plugins`). Then it can be used by calling `this.app.getCorePath("plugins")` which is much cleaner than doing path resolution in commands themself.

[`Environement.getPaths()`]: https://github.com/ArkEcosystem/core/blob/fix/core/plugin-commands/packages/core-cli/src/services/environment.ts#L20-L42

sebastijankuzner:
> I have a problem with the way `CORE_PLUGINS_PATH` is used. I believe [`Environement.getPaths()`](https://github.com/ArkEcosystem/core/blob/fix/core/plugin-commands/packages/core-cli/src/services/environment.ts#L20-L42) is supposed to handle path resolutions. It already handles `CORE_PATH_CONFIG` and `CORE_PATH_DATA` variables. I think it should similarly handle `CORE_PATH_PLUGINS` variable (and fallback to constructing default from `data/plugins`). Then it can be used by calling `this.app.getCorePath("plugins")` which is much cleaner than doing path resolution in commands themself.

CORE_PATH_DATA can be used instead CORE_PLUGINS_PATH. which is going to be removed.

rainydio (approved):

PR #4433, opened 2021-06-29
test(core-pino-logger): fix log rotate check

sebastijankuzner:
## Summary

Fix log rotate check. 

Rotating file stream is comparing interval to local time. To produce deterministic results tests are delayed and starts every time with 400ms delay after the last second. 

## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged


rainydio:
Can you update description why are you waiting for next second?

rainydio (approved):

PR #4466, opened 2021-08-17
deps(core-manager): include @arkecosystem/core

sebastijankuzner:
## Summary

Include @arkecosystem/core dependency to take advantage of update action, when core-manager is installed as plugin. 

## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged

### DIFF
--- a/packages/core-manager/package.json
+++ b/packages/core-manager/package.json
@@ -21,6 +21,7 @@
         "pretest": "bash ../../scripts/pre-test.sh"
     },
     "dependencies": {
+        "@arkecosystem/core": "^3.0.0-next.33",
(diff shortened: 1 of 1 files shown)


rainydio:
Shouldn't it be inside `peerDependencies` then?

sebastijankuzner:
> Shouldn't it be inside `peerDependencies` then?

For now it's simpler that way. Node doesn't recognize @arkecosystem/core when it is installed globally and @arkecosystem/core-manger is installed as plugin, due package resolution mechanism. I will provide additional PR if final solution will work with peerDependencies and custom resolution mechanism.

rainydio (approved):

PR #4482, opened 2021-09-07
Send new height with postBlock response

### LINKED ISSUE #4448: Peer height does not update in /api/peers
I mentioned this in Discord already and I know the devs have confirmed it's a bug so I'm reposting it here as a reference:

The `height` of each peer doesn't update in `/api/peers` in Core 3.0 and seems to always report whatever height the peer was at when the connection was first established. Contrast with Core 2.7 where the `height` value updates periodically for each peer so we can tell which peers are in sync and which aren't.

rainydio:
## Summary

Send new height with `postBlock` response. It's used to update `/api/peers` endpoint. Fixes #4448

## Checklist

- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/__tests__/unit/core-forger/client.test.ts
+++ b/__tests__/unit/core-forger/client.test.ts
@@ -2,7 +2,7 @@ import "jest-extended";
 
 import { Client } from "@arkecosystem/core-forger/src/client";
 import { Application, Container } from "@arkecosystem/core-kernel";
-import { NetworkStateStatus, Nes } from "@arkecosystem/core-p2p";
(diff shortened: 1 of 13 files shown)


alessiodf:
Hi,

Will the reply schema also need to be updated so it accepts objects of type `{ status: boolean, height: number }` rather than just a boolean as is the case currently?
https://github.com/ArkEcosystem/core/blob/9cdf872344267846b35ab2da12cba8afe0e049c8/packages/core-p2p/src/schemas.ts#L187-L189

sebastijankuzner (approved):

PR #4551, opened 2021-11-09
feat: implements PluginDiscoverer

sebastijankuzner:
## Summary

Implements PluginDiscoverer that loads plugins form project packages and plugins, installed plugins and from the node_modules. 

## Checklist

- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/packages/core-kernel/src/providers/plugin-discoverer.ts
+++ b/packages/core-kernel/src/providers/plugin-discoverer.ts
@@ -16,8 +16,9 @@ export class PluginDiscoverer {
         this.plugins = [];
 
         const discoverOnPath = async (patter: string, path: string): Promise<void> => {
-            existsSync(path);
-            this.plugins = this.plugins.concat(await this.discover(patter, path));
+            if (existsSync(path)) {
+                this.plugins = this.plugins.concat(await this.discover(patter, path));
+            }
         };
 
         await discoverOnPath("*/package.json", join(__dirname, "../../../../packages")); // Project packages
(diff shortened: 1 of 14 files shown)


packages/core-kernel/src/providers/plugin-discoverer.ts
```diff
@@ -0,0 +1,68 @@
...
+        const discoverOnPath = async (patter: string, path: string): Promise<void> => {
+            if (existsSync(path)) {
+                this.plugins = this.plugins.concat(await this.discover(patter, path));
+            }
+        };
+
+        await discoverOnPath("*/package.json", join(__dirname, "../../../../packages")); // Project packages
+        await discoverOnPath("*/package.json", join(__dirname, "../../../../plugins")); // Project plugins
+        await discoverOnPath("*/package.json", this.app.dataPath("plugins")); // Installed plugins
+    }
```
rainydio, line 28: Just make a loop.

```typescript
const paths = [/* */];

for (const path of paths) {
  if (existsSync(path)) {
    this.plugins = this.plugins.concat(await this.discover("*/package.json", path));
  }
}
```

rainydio (changes requested):

packages/core-kernel/src/providers/plugin-discoverer.ts
sebastijankuzner: There is no need to loop. Last discover uses different pattern. Fixed in: Fix packages pattern

rainydio (approved):

PR #4559, opened 2021-11-26
chore: update install.sh

sebastijankuzner:
## Summary

Update install.sh script and remove install-next.sh.

## Checklist

- [x] Ready to be merged


rainydio (approved):

adrian69 (approved):
I confirm it works fine with:
Debian 10/11
Ubuntu 18/20
CentOS 8

install.sh
```diff
@@ -47,8 +47,11 @@ RPM=$(which yum || :)
 # Detect SystemV / SystemD
 SYS=$([[ -L "/sbin/init" ]] && echo 'SystemD' || echo 'SystemV')
 
+# Detect GLIBC version
+GLIBC_VERSION=$(ldd --version | grep ldd | awk '{print $NF}')
+
 # Detect Debian/Ubuntu derivative
-DEB_ID=$( (grep DISTRIB_CODENAME /etc/upstream-release/lsb-release || grep DISTRIB_CODENAME /etc/lsb-release) 2>/dev/null | cut -d'=' -f2 )
+DEB_ID=$( (grep DISTRIB_CODENAME /etc/upstream-release/lsb-release || grep DISTRIB_CODENAME /etc/lsb-release || grep VERSION_CODENAME /etc/os-release)  2>/dev/null | cut -d'=' -f2 )
```
rainydio, line 54: Why not just `lsb_release --short --codename`?

rainydio (approved):

install.sh
adrian69: Some Ubuntu derivatives would fail since this would return the derivative codename itself. That for sure was the case with Mint for example.

PR #4587, opened 2022-01-24
refactor: extract KeyPairHolder logic 

sebastijankuzner:
## Summary

Extract `KeyPairHolder` logic from core-forger delegate. Logic is moved to core-kerne **Utils** so it can be shared between packages. 

KeyPairHolder implements Bip38 and Bip39 strategy. 

## Checklist

- [x] Tests
- [x] Ready to be merged


faustbrian:
What's the reason for moving it into the **kernel** package? What packages need to share it with the **forger** package? Moving it into kernel package breaks a whole lot of responsibilities and separations so it should be properly outlined what the bigger picture looks like of which these changes are part of.

sebastijankuzner:
> What's the reason for moving it into the **kernel** package? What packages need to share it with the **forger** package? Moving it into kernel package breaks a whole lot of responsibilities and separations so it should be properly outlined what the bigger picture looks like of which these changes are part of.

This keyPairHolders are needed for the other packages that will implement signing logic. I want to implement timestamp signing in the p2p package during the handshake to determine relays that belongs to delegates and then change consensus logic around it. Code is extracted to prevent duplication of different key storing strategies. 

We don't have a shared package that will contain a logic that can be used in other packages. Ideally KeyHolderLogic could be moved under the crypto package, but it contains dependencies on the core-kernel.

faustbrian:
Move it into the **crypto** package for now until I'll move it somewhere more appropriate.

PR #4589, opened 2022-01-31
fix(core): disable prompts for network selection

sebastijankuzner:
## Summary

Disable prompts for network selection when multiple network configurations are present. 

This PR is required for core-manager and other plugins that requires running CLI command without additional prompts. 

Network selection can be achieved with --network=${network} flag to load plugins for selected network. 

## Checklist

- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/packages/core-cli/src/input/parser.ts
+++ b/packages/core-cli/src/input/parser.ts
@@ -12,13 +12,8 @@ export class InputParser {
      * @memberof InputParser
      */
     public static parseArgv(args: string[]) {
-        const parsed: yargs.Arguments = yargs(args, { count: ["v"] });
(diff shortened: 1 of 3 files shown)


alessiodf:
Will this also fix the bug where `ark update` always claims to be on the latest version when multiple networks are present even if a newer version is available?

sebastijankuzner:
> Will this also fix the bug where `ark update` always claims to be on the latest version when multiple networks are present even if a newer version is available?

Thank you for pointing out bug @alessiodf. The side effect of the PR is that it fixes the problem with the update, but it require some additional changes that will be visible in next PR.

faustbrian (approved):

PR #4597, opened 2022-02-10
fix(core-transaction-pool): pass buffer content as string to child processes

sebastijankuzner:
## Summary

Transform buffer to string before sending it to the child process. When worker receives string content it transform it back to buffer in order to deserialize it. 

This fixes exiting issue when broadcasting tranactions to other peers, where transactions are not processed. 

## Checklist

- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/packages/core-transaction-pool/src/worker-script-handler.ts
+++ b/packages/core-transaction-pool/src/worker-script-handler.ts
@@ -18,12 +18,12 @@ export class WorkerScriptHandler implements Contracts.TransactionPool.WorkerScri
     }
 
     public async getTransactionFromData(
-        transactionData: Interfaces.ITransactionData | string,
+        transactionDataDTO: Interfaces.ITransactionData | string,
     ): Promise<Contracts.TransactionPool.SerializedTransaction> {
         const tx =
-            typeof transactionData === "string"
-                ? Transactions.TransactionFactory.fromBytes(Buffer.from(transactionData, "hex"))
-                : Transactions.TransactionFactory.fromData(transactionData);
+            typeof transactionDataDTO === "string"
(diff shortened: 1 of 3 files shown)


packages/core-transaction-pool/src/worker-script-handler.ts
```diff
@@ -18,12 +18,12 @@ export class WorkerScriptHandler implements Contracts.TransactionPool.WorkerScri
     }
 
     public async getTransactionFromData(
-        transactionData: Interfaces.ITransactionData | Buffer,
+        transactionDataDTO: Interfaces.ITransactionData | string,
```
faustbrian, line 21: Preserve the old variable name. Previously it vaguely resembled a DTO but  it no longer does.

packages/core-transaction-pool/src/worker.ts
```diff
@@ -32,7 +32,13 @@ export class Worker implements Contracts.TransactionPool.Worker {
             this.ipcSubprocess.sendAction("setHeight", currentHeight);
         }
 
-        const { id, serialized, isVerified } = await this.ipcSubprocess.sendRequest("getTransactionFromData", transactionData);
+        const transactionDataDTO =
+            transactionData instanceof Buffer ? transactionData.toString("hex") : transactionData;
+
+        const { id, serialized, isVerified } = await this.ipcSubprocess.sendRequest(
+            "getTransactionFromData",
+            transactionDataDTO,
```
faustbrian, line 40:
```suggestion
            transactionData instanceof Buffer ? transactionData.toString("hex") : transactionData,
```

faustbrian (approved):

PR #4611, opened 2022-02-15
ci: adjust devnet-sync.yml

sebastijankuzner:
## Summary

Devnet-sync.yml was overwritten in #4502. Reverting back to change from #4373.

## Checklist

- [x] Ready to be merged

### DIFF
--- a/.github/workflows/devnet-sync.yml
+++ b/.github/workflows/devnet-sync.yml
@@ -2,24 +2,15 @@ name: Devnet sync
 
 on:
     schedule:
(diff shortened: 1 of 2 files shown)


.github/workflows/devnet-sync.yml
```diff
@@ -2,24 +2,15 @@ name: Devnet sync
 
 on:
```
faustbrian, line 3: We should change this condition to `push` for the sake of testing so we can see that the job works before we merge this.

faustbrian (approved):

Issue #4640, opened 2022-04-11
/api/blocks not working correctly >= 3.2.1


alessiodf:
Since Core 3.2.1, the `/api/blocks` endpoint is not working correctly. I initially reported to @dated that the default sorting method is no longer descending by height (which is still the case in 3.2.2 too), but I see there are more problems so I am posting it as an issue now.

But more importantly, some blocks are simply missing from the results. Take, for example, the current output of https://api.ark.io/api/blocks?limit=5 (prettified and limited to 5 for ease of reading below). It skips from 4115424 to 4115426 (missing 4115425), 29, 31, 33 and so on. It occurs regardless of `limit` value and even if no `limit` is specified.

The issue is not present in 3.2.0. 

```
{
  "meta": {
    "totalCountIsEstimate": false,
    "count": 5,
    "pageCount": 3958220,
    "totalCount": 19791100,
    "next": "/blocks?limit=5&transform=true&page=2",
    "previous": null,
    "self": "/blocks?limit=5&transform=true&page=1",
    "first": "/blocks?limit=5&transform=true&page=1",
    "last": "/blocks?limit=5&transform=true&page=3958220"
  },
  "data": [
    {
      "id": "7753667336870923744",
      "version": 0,
      "height": 4115424,
      "previous": "17081634046000381964",
      "forged": {
        "reward": "200000000",
        "fee": "0",
        "amount": "0",
        "total": "200000000"
      },
      "payload": {
        "hash": "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855",
        "length": 0
      },
      "generator": {
        "username": "jamiec79",
        "address": "ARKfan7iMEK3cn8QDWVzcKdCCpL4MkASHb",
        "publicKey": "025631d7f863d2688443a14654e1b308c7585cb5f78563b3e785a978fdefff2c3e"
      },
      "signature": "304402202e85fba6c8b27f89e9e61a0700b259003c5f937e771cb8d286321a6296948b0f0220068d73d8fc327acb21e05f8820c6b2c0b32f329ecfa28a3720a6e8816582ea5d",
      "confirmations": 15675676,
      "transactions": 0,
      "timestamp": {
        "epoch": 33356136,
        "unix": 1523457336,
        "human": "2018-04-11T14:35:36.000Z"
      }
    },
    {
      "id": "8737217756359020838",
      "version": 0,
      "height": 4115426,
      "previous": "3255648078916117029",
      "forged": {
[...]

PR #4641, opened 2022-04-12
fix(core-api): sorting with multiple orderBy values

alessiodf:
Reverts #4636 due to bugs referenced in #4640 and implements Solar-network/core@22a5ca27008736b12da1b6b176acffd5847251bc instead.


faustbrian (approved):

faustbrian:
Thanks!

PR #4660, opened 2022-04-28
adjust  `maxBlockPayload` to fit next maxTx size

tnga:
## Summary

This considering next `maxTxPerBlock` value to 150.
This considering the value will be updated with network when approriate
This has also to be changed in modules `typescript-crypto-networks` when time will come

## Checklist

- [ x ] Tests _(if necessary)_
- [ x ] Ready to be merged

### DIFF
--- a/__tests__/unit/core-test-framework/app/generators/__fixtures__/assets.ts
+++ b/__tests__/unit/core-test-framework/app/generators/__fixtures__/assets.ts
@@ -14,7 +14,7 @@ export const sandboxOptions: SandboxOptions = {
             delegates: 51,
             blocktime: 8,
             maxTxPerBlock: 150,
-            maxBlockPayload: 2097152,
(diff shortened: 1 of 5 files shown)


faustbrian:
What exactly is this needed for?

tnga:
It seems that from actual default ark core config, there is a planning to push the maximum transactions per block to `150`.
That said in `typescript-crypto-networks` repo, it is set to `50` `maxTxPerBlock` with  the maximum payload length by block (`maxPayload`) set to `2097152` .
There is a need to increase the value of the `maxPayload` when the `maxTxPerBlock` will be upgraded to `150`. 
This to be sure that block will have necessary size to content 150 transactions "hard" data.
This noting that `maxTxPerBlock` is already set to `150` for devnet ark core test without the adjustment of `maxPayload` accordly.

PR #4681, opened 2022-06-22
fix(core-p2p): skip highest common block search on fast validation 

sebastijankuzner:
## Summary

Add fast option to the peer verification. Fast option is only used to calculate quorum before forging. It skips the calculation of the highest common block. Result is saved in separate property to prevent collision with the normal check.  

## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged

### DIFF
--- a/packages/core-p2p/__tests__/peer-verifier.test.ts
+++ b/packages/core-p2p/__tests__/peer-verifier.test.ts
@@ -1,6 +1,6 @@
 import { Application, Container, Contracts } from "@packages/core-kernel";
 import { Peer } from "@packages/core-p2p/src/peer";
-import { PeerVerificationResult, PeerVerifier } from "@packages/core-p2p/src/peer-verifier";
+import { PeerVerificationResult, FastPeerVerificationResult, PeerVerifier } from "@packages/core-p2p/src/peer-verifier";
 import { Blocks } from "@packages/crypto";
 
 describe("PeerVerifier", () => {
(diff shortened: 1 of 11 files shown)


packages/core-p2p/__tests__/peer-verifier.test.ts
```diff
@@ -807,4 +807,349 @@ describe("PeerVerifier", () => {
...
+                    .spyOn(Blocks.BlockFactory, "fromData")
+                    .mockImplementation(blockWithIdFromDataMock);
+
+                const result = await peerVerifier.checkStateFast(claimedState, Date.now() + 2000);
+
+                expect(result).toBeInstanceOf(FastPeerVerificationResult);
+                expect(result.forked).toBeTrue();
+                expect(result.myHeight).toEqual(15);
+                expect(result.hisHeight).toEqual(18);
+                expect(result.highestCommonHeight).toBeUndefined();
```
ItsANameToo, line 877: shouldn't this have a value when they have a common block?

sebastijankuzner: Actually we don't need to determine highestCommonHeight, because we don't need to determine if peer is forked when its height is higher than ours. We need fork state only when peer height is equal or lower than ours when calculating [network state.](https://github.com/ArkEcosystem/core/blob/81838816f5a4d668fa0cbcf224ad12c05ddd0315/packages/core-p2p/src/network-state.ts#L171-L182) .

That makes Case 1 and Case 2 response actually the same, because we can skip highestCommonHeight calculation to improve performance.

ItsANameToo (approved):

PR #4714, opened 2022-09-20
feat(core-transactions): implements second and multi signature memoizers

sebastijankuzner:
## Summary

Implements memoizer for second and multi signature verification. Memoizer caches Validator result. They are used to speed up transaction processing when re-adding transactions from mempool, since signature verification is very CPU consuming. 

Memoizers uses LRU cache. 

## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged

### DIFF
--- a/packages/core-transactions/src/memoizers/multi-signature-verification.ts
+++ b/packages/core-transactions/src/memoizers/multi-signature-verification.ts
@@ -1,16 +1,16 @@
-import { Container, Providers } from "@arkecosystem/core-kernel";
+import { Container, Contracts, Providers } from "@arkecosystem/core-kernel";
 import { Interfaces, Transactions } from "@arkecosystem/crypto";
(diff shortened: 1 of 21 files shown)


packages/core-transactions/src/memoizers/multi-signature-verification.ts
```diff
@@ -0,0 +1,52 @@
...
+        }
+
+        const result = Transactions.Verifier.verifySignatures(transaction, multiSignatureAsset);
+
+        this.lruCache.set(key, { multiSignatureAsset: Utils.cloneDeep(multiSignatureAsset), result });
+
+        return result;
+    }
+
+    public clear() {
```
ItsANameToo, line 41: is this used anywhere? 🤔

ItsANameToo (approved):

